PR #205, opened 2026-05-06
Add pending read buffer for nexus async TLS

birchrust:
The previous fix for partial TLS input consumption moved the async path to `read_and_process_tls`, which is the right shape for handshake bursts and other bounded buffered input.

The problem is that steady-state application data has one more constraint: after `process_new_packets`, rustls may have plaintext waiting in its internal reader. If the adapter immediately keeps feeding more ciphertext before the caller gets a chance to drain that plaintext, rustls can eventually return:

```text
received plaintext buffer full
```

So this PR keeps the “do not drop unconsumed ciphertext” behavior, but adds a plaintext-drain point between accepted ciphertext prefixes.

I may be overfitting this to the nexus backend’s current poll loop, so please feel free to reject this PR if you would rather solve the buffering boundary in TlsCodec itself or expose a different helper from nexus-net.

### DIFF
--- a/nexus-async-net/src/maybe_tls/nexus.rs
+++ b/nexus-async-net/src/maybe_tls/nexus.rs
@@ -29,6 +29,8 @@ pub enum MaybeTls {
 pub struct TlsInner {
     pub(crate) stream: TcpStream,
     pub(crate) codec: nexus_net::tls::TlsCodec,
+    /// Ciphertext read from the transport but not yet accepted by rustls.
+    pending_read: Vec<u8>,
     /// Ciphertext waiting to be flushed to the transport.
     pending_write: Vec<u8>,
 }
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0:
I see, let me just expand on this so it's handled across all the same places in the rest of the workspace. I'll open a PR against your branch.

Abso1ut3Zer0:
Superseded by #206 — your diagnosis and fix mechanism are preserved as the first commit on the merge branch with original authorship (`3d4fef1`, authored by you). The layered work on top extends the fix to nexus-net's tokio adapter (which had the identical bug), hoists the helper as `TlsCodec::read_tls_step` so future adapter authors inherit it, migrates the pending buffers to nexus-net's cursor-based ReadBuf/WriteBuf primitives (eliminates the O(n) drain memmove on every TLS packet), closes a related AsyncWrite chunking bug we found while in the area, and adds TLS `close_notify` on shutdown. Three audit passes (code review + hot-path + deep audit) sweep through the result. Thanks for the seed — good catch on the steady-state constraint.

PR #389, opened 2026-05-29
Design notes: shm foundation

ChronoCoders:
Adds three inline reviewer notes to docs/design/nexus-shm.md covering foundation liveness, journal crash recovery, and the slot dead-writer case. The throughline: prefer kernel-backed liveness (robust futex / OFD locks) over timing-based heartbeats, and reuse that one primitive across the journal and slot designs.

### DIFF
--- a/docs/design/nexus-shm.md
+++ b/docs/design/nexus-shm.md
@@ -74,6 +74,18 @@ Regardless of crate structure, the foundation layer handles:
 - **Crash/stale peer detection:** Heartbeat or epoch mechanism to
   distinguish dead peers from slow ones. This is the core invariant
   the upper layers depend on.
+
+> **[altug]** I'd steer the foundation away from timing as the source of truth for liveness. A heartbeat threshold can't actually separate a dead peer from a busy one. Too tight and a peer mid-burst looks dead, too loose and you carry stale state for seconds.
+>
+> The robust answer is to ask the kernel instead of inferring from a clock. Two options that give crash detection for free:
+>
+> - A robust futex on the control word (FUTEX_OWNER_DIED, PTHREAD_MUTEX_ROBUST). The kernel flags it when the owning process dies, the next attacher gets EOWNERDEAD and runs recovery. Canonical Linux answer to "shm lock owner died".
+> - An flock or OFD lock per process. The kernel releases it on death, so a peer can test it to know definitively whether the other side is alive.
+>
+> Pair either one with a generation counter in the control block so restarts and PID reuse stay unambiguous. A restarted peer bumps the generation, and stale state from the dead instance is detectable structurally, not by timing.
+>
+> Heartbeat is still useful as a soft "making progress" hint for the slow-but-alive case. It just shouldn't be the thing that declares a peer dead. That call should come from the OS.
+
 - **Recovery routine:** Given a mapping that may contain half-written
   data from a crashed writer, restore to a consistent state.
 
@@ -123,6 +135,11 @@ recovery. Primary caller: FIX message journaling.
   thread, no extra copies
 - Crash recovery: scan last segment on startup, find the last
   committed record, discard partial writes
+
+> **[altug]** Worth making the commit-ordering invariant explicit, since the whole recovery story rests on it. Write body first, publish the length or a commit marker last behind a release fence. Recovery scans forward while records carry a valid committed marker and stops at the first partial one. That ordering is the real correctness property, and it's exactly the kind of thing a later refactor can silently reorder and break.
+>
+> One open call: a SIGKILL can land mid-memcpy on the payload, so a length-last marker alone won't catch a torn body. A per-record checksum catches it but costs you on the hot path. Might be worth a config flag. Checksums on for the journal where durability matters, off for callers fine with discarding the last partial record on a length mismatch.
+
 - Cross-process read: other processes mmap segments read-only
 
 **Design question: sequence awareness**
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0:
Good notes on all three. Spent some time thinking through the liveness design and wanted to share where I landed.

**On foundation liveness (comment 1):**

Agree that timing shouldn't be the source of truth for death — but I want to separate this into two tiers rather than picking one mechanism.

**Tier 1 — Atomic status field in the shared memory header.** Set to `ALIVE` on attach, set to `DEAD` in a `Drop` guard. Covers normal shutdown and panic-with-unwind, which is the vast majority of real-world exits. ~1 cycle to check. Always present, zero cost. Not a heartbeat — it's a state declaration, not a progress indicator. For in-process shared memory (e.g., the FIX journal where the same process writes and reads for resend), the atomic alone is sufficient — if the process dies, there's no surviving peer that needs to detect it. The kernel-backed tier 2 is only relevant for cross-process communication where one side can die independently.

**Tier 2 — OFD lock via `fcntl`, exposed as an opt-in method.** Acquired on attach, kernel releases on process death including SIGKILL. The library exposes `peer_alive_kernel()` but never calls it automatically. The user decides when and where — monitoring thread, health check endpoint, periodic sweep. Their deployment, their policy.

Quick bench of `F_OFD_GETLK` via the `nix` crate showed **~330 cycles p50 (~95ns), tight through p999 (~500 cycles)**, but this was a simple isolated test — not a realistic environment. The max hit ~87µs from system noise (TLB shootdowns, SMI, scheduler preemption while in kernel). This needs further benchmarking under realistic workloads before we commit to using it inline anywhere. A syscall on the event loop is a potential latency killer if we're getting context-switched under load or hitting unexpected kernel contention. Until we have more data, detection policy stays with the user so they can isolate the syscall to a non-critical thread.

Generation counter — agreed, necessary for PID reuse safety.

`nix` is the dependency for fcntl operations. Well-maintained, typed API, covers `F_OFD_SETLK` / `F_OFD_GETLK`.

**On journal commit ordering (comment 2):**

Agree on making the ordering invariant explicit and documenting it as the correctness property. Write body → release fence → publish commit marker. Recovery scans forward, stops at first uncommitted record.

On checksums: Aeron's log buffers don't checksum — they use negative frame length as a sentinel (claimed, not committed), and recovery writes a PAD frame over uncommitted claims. Artio is built on top of this for production FIX messaging, so the pattern is proven for our use case. I'd follow that model: sentinel-based commit markers, no checksums by default. If we want opt-in checksums for the durable file-backed journal (disk errors, torn page writeback), we can add that as a config flag, but Aeron's approach should be the baseline design.

**On ShmSlot liveness (comment 3):**

Agree on reusing the foundation's liveness rather than a separate mechanism. Reader logic:

1. Version even → read value, verify version unchanged after read
2. Version odd (mid-write) → check atomic status (~1 cycle)
3. Atomic says `ALIVE` → retry, writer is genuinely mid-write
4. Atomic says `DEAD` → shadow copy + staleness flag
5. User wants SIGKILL coverage → call `peer_alive_kernel()` (opt-in, their choice when/where)

Shadow copy of last good value — agreed. Note: Pod (safe for shared memory — flat repr, no heap pointers, no Drop) is a separate concept from Copy. A large Pod struct (e.g., a 4KB book snapshot) shouldn't have implicit Copy semantics. The shadow is an explicit memcpy by the reader, not a Copy trait bound.

**Other design resolutions from this discussion:**

- **One crate.** `nexus-shm` with the mmap foundation as an internal module. No separate `nexus-memmap` crate — if someone needs raw mmap wrappers later, we extract then.
- **Both backing stores.** mmap'd files on disk for `ShmJournal` (durability). POSIX shm (`/dev/shm`) for `ShmRingBuffer` and `ShmSlot` (speed, no persistence needed). Foundation supports both.
- **SPSC first, separate types for multi-participant.** No generic type parameterized over topology. If MPSC materializes as a need, it's a separate type with its own liveness complexity.
- **Pod only, index-based addressing.** No relocatable pointers. All shared memory types must be Pod. Internal structure uses integer indices, not pointers — same as nexus-collections. Pod is distinct from Copy — large Pod types should not have implicit copy semantics.

If these directions look right, I'll open issues to track each primitive and capture the design context.

ChronoCoders:
All of this looks right, and the two-tier liveness split is better than where I started. The ~1-cycle atomic state declaration plus an opt-in kernel check is the correct decomposition, and keeping the GETLK syscall off the hot path and user-controlled is exactly right given that 87us tail. On the event loop it would be a latency landmine.

One thing to document on Tier 1: the Drop guard only fires on graceful shutdown and unwinding panic. Under panic=abort, which a low-latency or embedded build may well use, Drop never runs, so the atomic stays ALIVE through the panic and you fall straight through to Tier 2. Same for SIGKILL and segfault. So I would state Tier 1's guarantee precisely as "graceful Drop plus unwinding panic only," and flag that a panic=abort build needs Tier 2 as mandatory, not optional, for death detection. Worth a doc line so it is not a surprise later.

On the journal, you are right and I over-specified the checksum. Length-last sentinel covers the crash case fully. The marker is a single aligned store written after the body, so recovery never sees a committed-but-torn record. Aeron plus Artio is the proven baseline, follow it. Checksums only earn their place against storage corruption on the durable file, torn page writeback or bit rot, which is a different failure domain. Opt-in flag there is the right scope.

Slot logic is exactly right, and good catch on Pod versus Copy, I was loose there. A 4KB snapshot should not be implicitly Copy, the shadow is an explicit reader-side memcpy. Minor note for large hot slots: the seqlock retry copies the whole struct per attempt, so retry cost scales with size. Fine for low-write-rate config and book snapshots, worth knowing if a slot is ever both large and hot.

One crate, both backing stores, SPSC-first, Pod-only with index addressing, all good. The no-relocatable-pointers rule is the one that bites people who have not done shm before, so glad it is stated up front. Open the issues, this is a solid foundation to build on.

Abso1ut3Zer0:
New issues are opened!

Abso1ut3Zer0 (approved):

Issue #390, opened 2026-05-30
nexus-shm: foundation layer (mmap, liveness, Pod)


Abso1ut3Zer0:
## Overview

Foundation layer for all nexus-shm primitives. Handles mmap lifecycle,
liveness detection, alignment, and the Pod trait boundary.

Internal module of nexus-shm (not a separate crate).

### mmap lifecycle
- Create, open, resize, unmap
- `MAP_POPULATE` (pre-fault) and `MAP_HUGETLB` (huge pages) as opt-in
- Both backing stores: mmap'd files on disk + POSIX shm (`/dev/shm`)
- `memmap2` for mmap internals, `nix` for fcntl

### Cache-line alignment
- All shared structures aligned to 64-byte boundaries
- False sharing prevention between producer and consumer metadata

### Liveness detection — two-tier

**Tier 1 — Atomic status field** in shared memory header.
`ALIVE` on attach, `DEAD` in Drop guard. ~1 cycle. Always present.
Covers graceful shutdown and unwinding panic only.

`panic=abort` caveat: Drop never runs under `panic=abort`. Atomic stays
ALIVE through the panic. Same for SIGKILL and segfault. Document
Tier 1's guarantee precisely. A `panic=abort` build needs Tier 2 as
mandatory, not optional.

For in-process shm (FIX journal, same-process read/write), atomic
alone is sufficient — process death takes out both sides.

**Tier 2 — OFD lock via fcntl** (`nix` crate). Acquired on attach,
kernel releases on process death (including SIGKILL). Library exposes
`peer_alive_kernel()` but never calls it automatically. User decides
when/where. Only relevant for cross-process communication.

Benchmarked `F_OFD_GETLK`: ~330 cycles p50, ~500 cycles p999. Max ~87µs
from system noise. Needs further benchmarking under realistic workloads —
potential latency killer if context-switched under load. Detection
policy stays with the user.

Generation counter in control block for PID reuse safety.

### Pod trait
All shared memory types must be Pod (flat repr, no heap pointers, no Drop).
Pod is distinct from Copy — large Pod types (e.g., 4KB book snapshot)
should not have implicit Copy semantics.

Index-based addressing only (integers into arrays). No relocatable pointers.
Same pattern as nexus-collections.

## Dependencies
- `nix` (fcntl `F_OFD_SETLK` / `F_OFD_GETLK`)
- `memmap2` (mmap internals)

## References
[...]

PR #403, opened 2026-05-30
nexus-shm: foundation layer (#390)

ChronoCoders:
Foundation layer per #390. Implements the rulings from the design thread.

- `Pod` — stricter than nexus-slot's: valid for every bit pattern (excludes
  bool/char), since a reader can observe a crashed writer's bytes.
- Control block — 32-byte `repr(C)` inner + `CachePadded` wrapper, no `_reserved`.
- `Segment` — mmap via raw `nix` (no memmap2). Huge pages opt-in, hard error
  when unavailable — no silent fallback.
- Liveness — Tier 1 atomic status + Drop guard; Tier 2 OFD lock with 3-state
  `Liveness { Alive, Dead, Unknown }`.
- Criterion bench for the `F_OFD_GETLK` callsite. Numbers need a pinned,
  turbo-off run on real hardware; harness header documents the procedure.

8 tests pass; clean under `cargo clippy --all-targets -- -D warnings` and fmt.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
 	"nexus-logbuf",
 	"nexus-pool",
 	"nexus-queue",
+	"nexus-shm",
 	"nexus-slab",
 	"nexus-slot",
 	"nexus-smartptr",
--- a/nexus-shm/Cargo.toml
+++ b/nexus-shm/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "nexus-shm"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+description = "Shared-memory IPC primitives for multi-process trading systems"
+repository.workspace = true
+keywords = ["shared-memory", "ipc", "mmap", "low-latency", "lock-free"]
+categories = ["concurrency", "data-structures", "os::unix-apis"]
+
+[lints]
+workspace = true
+
+[dependencies]
+crossbeam-utils.workspace = true
+nix = { version = "0.30", features = ["mman"] }
--- a/nexus-shm/Cargo.toml
+++ b/nexus-shm/Cargo.toml
@@ -14,4 +14,4 @@ workspace = true
 
 [dependencies]
 crossbeam-utils.workspace = true
-nix = { version = "0.30", features = ["mman"] }
+nix = { version = "0.30", features = ["mman", "fs"] }
--- a/nexus-shm/Cargo.toml
+++ b/nexus-shm/Cargo.toml
@@ -15,3 +15,10 @@ workspace = true
 [dependencies]
 crossbeam-utils.workspace = true
 nix = { version = "0.30", features = ["mman", "fs"] }
+
+[dev-dependencies]
+criterion = { version = "0.5", features = ["html_reports"] }
+
+[[bench]]
+name = "ofd_liveness"
+harness = false
--- a/nexus-shm/README.md
+++ b/nexus-shm/README.md
@@ -1,0 +1,130 @@
+# nexus-shm
+
+Shared-memory IPC primitives for multi-process trading systems.
+
+> **Status: foundation in progress.** The mmap lifecycle, two-tier liveness,
+> and the primitives below are under active development on top of the
+> foundation layer that exists today (`Pod`, the segment control block). Track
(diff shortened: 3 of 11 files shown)


Abso1ut3Zer0:
## Preliminary Review — nexus-shm foundation layer

Architecture is solid. Control block layout, two-tier liveness, OFD locks, Pod strictness — all match the design doc. Good foundation. Comments below are things to address before merge.

---

### Must fix

**1. SAFETY comments on every unsafe block/impl.** This is a project-wide requirement. I count 6 sites that need them:

- `region.rs` — `unsafe impl Send for Mapping {}` / `unsafe impl Sync for Mapping {}` — why is it safe to send/share a mapping across threads?
- `region.rs` — `mmap` call — alignment, validity of the fd, lifetime of the mapping
- `region.rs` — `munmap` in Drop — pointer and length are valid from construction
- `segment.rs:38` — `&mut *mapping.as_ptr().cast::<ControlBlock>()` — exclusive access guaranteed by OFD lock, alignment guaranteed by mmap page alignment
- `segment.rs:56` — `&*mapping.as_ptr().cast::<ControlBlock>()` — shared access is safe because fields are atomic or write-once
- `segment.rs:62` — `.add(HEADER)` in `data()` — offset is within the mapped region

**2. `Segment::create` should reject `data_len == 0`.** Currently `HEADER + 0` = 64, converts to `NonZeroUsize` fine, and you get a segment with no data space. Add an early `if data_len == 0 { return Err(ShmError::EmptySegment); }`.

**3. The `try_into()` error mapping in `create` is dead code.** `HEADER` is 64, so `HEADER + data_len` is always non-zero (and normal `+` panics on overflow in debug). The `.map_err(|_| ShmError::EmptySegment)` never fires. If you want overflow protection, use `checked_add` and return an appropriate error — `EmptySegment` is the wrong variant for "your data_len overflowed usize."

**4. `data()` needs a doc comment explaining the safety contract.** It returns `*mut u8` through `&self` — callers need to know: what synchronization is required for reads/writes, what the valid length is, that the pointer is only valid for the lifetime of the `Segment`. This is the primary API surface for the primitives built on top.

---

### Naming

**5. Rename `PeerStatus` to `Status` and `peer_status()` to `status()`.** The "peer" prefix is redundant on a segment you attached to. More importantly, the doc comment should make the one-way reliability clear: "`Dead` is authoritative. `Alive` may be stale if the owner died without cleanup (SIGKILL, panic=abort) — use `peer_liveness()` to confirm via kernel lock."

---

### Design note

**6. Segment should not grow file management methods.** After looking at how Aeron and iceoryx2 handle this — Aeron's media driver nukes and rebuilds the directory, iceoryx2 puts ownership transfer and cleanup on a higher layer (Node/Service) — the right model for `Segment` is: it's a mapping with liveness signals. File lifecycle (unlink, rotate, archive) belongs on the primitives built on top (`ShmJournal`, `ShmRingBuffer`) or the caller. The current PR gets this right by not having any file management. A one-liner in the `Segment` doc or README saying "file cleanup is the caller's responsibility; the segment file persists after drop for crash recovery" would make this an explicit design choice rather than an omission.

---

### Minor / nits

**7. `lock.rs:34` — `e as i32` on nix `Errno`.** Works because `Errno` is `repr(i32)`, but a comment noting that would help, or use the nix conversion.

**8. Huge pages error attribution in `region.rs:67-72`.** Any mmap failure when `opts.huge_pages` is true gets wrapped as `HugePagesUnavailable`, even if the real cause is something else. Minor but the error can mislead.

ChronoCoders:
Addressed all 8.

- SAFETY comments on the 6 unsafe sites.
- Zero-len guard; overflow now `checked_add` → `SizeOverflow` (not `EmptySegment`).
- `data()` safety doc — valid length, lifetime, caller-owned synchronization.
- `PeerStatus` → `Status` / `status()`, with the Dead-authoritative / Alive-may-be-stale note.
- File-lifecycle note on `Segment` (caller's responsibility; file persists after drop for recovery).
- `Errno::into()` for the errno cast.

On #8 I scoped huge-pages attribution to `ENOMEM` only — `EINVAL` is ambiguous (alignment, bad flags, unsupported), so it falls through to `Os` rather than risk the same misattribution narrowed. Ready for another look.

Abso1ut3Zer0 (approved):

Issue #411, opened 2026-05-31
nexus-fix-engine: session journal + replay index


Abso1ut3Zer0:
## Overview

Pluggable persistence via traits, with a default implementation backed by
nexus-shm's ShmJournal for durable message storage.

Parent: #408
Design doc: `docs/design/nexus-fix-engine.md`

### Trait definitions

```rust
pub trait MessageStore {
    type Error;
    fn store(&mut self, session_id: SessionId, direction: Direction,
             seq_num: u32, msg: &[u8]) -> Result<(), Self::Error>;
    fn retrieve(&self, session_id: SessionId, range: SeqRange)
             -> impl Iterator<Item = Result<StoredMessage<'_>, Self::Error>>;
}

pub trait SessionStore {
    type Error;
    fn load(&self, id: SessionId) -> Result<Option<SessionState>, Self::Error>;
    fn save(&mut self, id: SessionId, state: &SessionState) -> Result<(), Self::Error>;
}
```

### ShmJournal integration (default impl)

```
TCP read -> raw bytes
  -> journal.append(...)     <- mmap write, ~ns
  -> fix_codec.parse(bytes)  <- zero-copy from read buffer
  -> session.handle_message  <- state machine
  -> poll_event loop         <- dispatch

Resend request:
  -> journal.read_range(...) <- mmap read, zero-copy
  -> TCP write
```

### Performance target

Message store write (ShmJournal): < 500ns

## Dependencies

- `nexus-fix-codec` (#406)
- `nexus-shm` (ShmJournal — #391)
- Session state machine (#409)

Abso1ut3Zer0:
Revisiting the framing here. The engine is sans-IO — it shouldn't own persistence traits. The caller knows their storage (ShmJournal, file, whatever) and handles replay when they get a ResendRequest from the `Message` enum.

What's actually needed is a seq-num index that sits alongside the journal — a mapping from FIX sequence number to journal offset so you can look up "give me messages in seq range 42..50" efficiently. The journal itself is just an append-only byte buffer; it doesn't know about FIX sequence numbers.

Need to research common patterns here before committing to a design — how other FIX engines handle message replay, what the typical resend window looks like in practice, and whether the index lives in shared memory alongside the journal or separately.

Abso1ut3Zer0:
Updated design direction after deeper research into Artio's model and mmap behavior.

**Architecture: single-owner per-session mmap'd journal with three-segment rotation**

[...]

Issue #410, opened 2026-05-31
nexus-fix-engine: TCP framing


Abso1ut3Zer0:
## Overview

Sans-IO framer that splits a TCP byte stream into individual FIX messages.
Consumes `&[u8]` chunks and produces framed messages — no sockets, no
buffering policy.

Parent: #408
Design doc: `docs/design/nexus-fix-engine.md`

### Framing logic

1. Scan for `8=FIX...` header
2. Parse BodyLength (tag 9) to determine message extent
3. Extract exactly BodyLength bytes plus the trailer (tag 10)
4. Validate checksum before handing the message to the session

### API

```rust
pub struct FixFramer {
    state: FramerState,
}

impl FixFramer {
    /// Feed bytes from TCP. Returns framed message (if complete)
    /// and the number of bytes consumed.
    pub fn decode<'a>(&mut self, buf: &'a [u8])
        -> Result<(Option<&'a [u8]>, usize), FrameError>;
}
```

### Design question (from design doc)

Should the framer live in nexus-fix-codec (pure byte-level concern) or in the
engine (TCP-stream-aware)? Framing needs BodyLength parsing which is codec-level,
but the framer is only useful in a streaming context which is engine-level.

## Dependencies

- `nexus-fix-codec` (#406) for checksum validation and tag parsing

Abso1ut3Zer0:
Covered by #464 (FrameReader/FrameWriter). PR #466.

Issue #409, opened 2026-05-31
nexus-fix-engine: session state machine


Abso1ut3Zer0:
## Overview

Sans-IO FIX session state machine with poll-based event dispatch and
caller-owned encode buffer. Handles all admin messages internally;
application messages emitted as events for user code.

Parent: #408
Design doc: `docs/design/nexus-fix-engine.md`

### API shape

```
handle_message(msg, now) -> pushes events to internal buffer
handle_timeout(now)      -> pushes timer events
poll_event()             -> drains events, caller processes
encode_pending(kind, buf)-> encodes into caller's buffer
```

Session never allocates. The caller provides a workhorse buffer for encoding.
Events are `Copy` enum variants (no borrowed data).

### State diagram

```
Disconnected -> LogonSent -> Active <-> Resending -> LogoutPending -> Disconnected
```

### Admin messages handled internally

- Logon / Logout
- Heartbeat / TestRequest
- ResendRequest / SequenceReset
- Reject

### Sequence management

Inbound and outbound sequence number tracking. Gap detection triggers
ResendRequest. SequenceReset-GapFill for admin messages during resend.

### Open questions (from design doc)

- **Heartbeat timer integration:** Should the session own its timer state,
  or delegate to the caller's timer wheel? Owning simplifies the API but
  couples to a time source.
- **Session multiplexing:** One session per TCP connection is standard FIX.
  Some venues multiplex. Higher-level concern or engine support?

## Dependencies

- `nexus-fix-codec` (#406)

Abso1ut3Zer0:
Covered by PR #431 (SessionState) and #461 (Session<D> routing layer).

Issue #408, opened 2026-05-31
nexus-fix-engine: sans-IO session layer


Abso1ut3Zer0:
## Overview

Sans-IO FIX session state machine, TCP framing, and persistence. Same
architectural pattern as nexus-net's WebSocket codec — pure state machines,
no sockets, no async, no runtime dependency. Time is injected.

Design doc: `docs/design/nexus-fix-engine.md`

Depends on `nexus-fix-codec` for parsing primitives.

## Sub-issues

This is the tracking issue. Implementation is broken into:

- Session state machine (admin message handling, sequence management)
- TCP framing (splitting byte stream into FIX messages)
- Persistence traits + ShmJournal integration

## Performance targets

| Operation | Target |
|-----------|--------|
| Session logic (admin msg handling) | < 100ns |
| Message store write (ShmJournal) | < 500ns |
| Full inbound path (frame + codec + session + store) | < 1μs |

## Dependencies

- `nexus-fix-codec` (#406)
- `nexus-shm` (for persistence integration)

PR #412, opened 2026-05-31
docs: nexus-fix implementation plan (#406-#411)

ChronoCoders:
Design-first, before implementation — same path as #389.

Maps the FIX roadmap (#406–#411) onto five code PRs plus a tracking comment on #408, in dependency order: #406 → {#407, #409, #410} → #411.

Doc: `docs/design/nexus-fix-impl-plan.md`. Per-PR scope, design rulings on the open questions from the design docs, and what's deferred.

Two decisions I'd like your call on before I implement:

- **#407 multi-char enums** — `SecurityType` (CS/FUT/OPT), `MsgType` etc. can't be `#[repr(u8)]`. Plan: generator detects enum value width, emits `from_bytes`/`as_bytes` for multi-char, keeps the `u8` form for single-char. Right shape?
- **#411 persistence** — default impl needs ShmJournal (#391), not built yet. Lean (a): ship traits + in-memory impl now so #409/#410 integrate against the trait, ShmJournal impl follows #391. Or (b) hold until #391 lands. Your preference?

Design rulings folded in from the earlier DM thread: timer state in the session driven by `handle_timeout(now)` + `next_deadline()` (#409); framer in the engine with a `max_message_len` cap on attacker-controlled BodyLength (#410); `Direction`-tagged outbound store for resend (#411).

No code in this PR — review the plan and I'll start at #406.


Abso1ut3Zer0:
## Review — implementation plan

Good structure. Dependency order is right, per-PR scope is clear, design rulings are sound. Answers to your two asks plus notes on the open questions below.

---

### Decision 1: Multi-char enums (#407)

Yes, this is the right shape. Generator detects value width, emits `from_byte`/`as_byte` for single-char and `from_bytes`/`as_bytes` for multi-char. Most FIX enums are single-char so the fast path stays fast.

### Decision 2: Persistence (#411)

Option (a) — traits + in-memory impl now, ShmJournal impl follows #391. No reason to block #409/#410 integration on a dependency that isn't built yet.

---

### Design rulings — all confirmed

- Timer: `handle_timeout(now)` + `next_deadline()`, session never reads a clock. Correct.
- Framer in the engine, not the codec. Codec stays a pure leaf.
- `max_message_len` cap belongs on the framer, before the session sees bytes. Correct.
- Direction-tagged outbound store for resend. Correct.

---

### Open questions — my calls

**PR 1 (#406):**
1. Keep `is_present()` / `EMPTY`. Reads naturally at call sites, `is_empty()` has the wrong connotation.
2. Reserve-ahead via `header_len()` — avoids the memmove, preferred.
3. Bench on pinned-core / turbo-off setup, yes.

**PR 2 (#407):**
1. `quick-xml` — pull-based, fast, right choice.
2. Emit pre-formatted. No rustfmt runtime dep.
3. Cap nesting depth. Real dictionaries never exceed 2.

**PR 3 (#409):**
1. Raw `u64` nanos for `now`. No coupling to any time crate.
2. One session per connection. Multiplexing is out of scope.
3. Caller-drained-before-next-call invariant for the event buffer.

**PR 4 (#410):**
1. Caller re-feeds, framer returns consumed count. Allocation-free.
2. Resync to next `8=FIX` on garbage, not hard error.

**PR 5 (#411):**
1. RPITIT — clean, MSRV 1.94 supports it.
2. Per-message persistence — that's the whole point of ShmJournal.

---

### Note: reuse nexus-net buffer primitives

The framer should use `ReadBuf` / `WriteBuf` from nexus-net rather than rolling its own buffer management. These are cursor-FIFO types built for exactly this pattern (TCP partial reads, compact-on-drain, no per-write memmove). nexus-fix-engine depending on nexus-net for these is a clean dependency — networking buffer primitives from the networking crate.

If the FIX framer needs features those types don't have yet, we add them to nexus-net rather than building parallel infrastructure.

Longer term, the WS/HTTP protocol code in nexus-net may move to a dedicated crate (nexus-web or similar), with nexus-net remaining as the shared networking primitives layer. That's a separate concern — doesn't affect anything here.

Issue #391, opened 2026-05-30
nexus-shm: ShmJournal — append-only durable log


Abso1ut3Zer0:
Depends on: #390 (foundation layer)

## Overview

Append-only durable log over mmap'd files. Primary caller: FIX message
journaling for resend. Aeron Archive model.

## Key properties
- Segmented file layout (configurable segment size)
- Writing to mmap IS the persistence — no archival thread, no extra copies
- Cross-process read: other processes mmap segments read-only
- SPSC (single writer, single/multiple readers)

## Crash recovery — Aeron model
Write body → release fence → publish commit marker. Recovery scans
forward, stops at first uncommitted record. Aeron uses negative frame
length as sentinel (claimed, not committed), recovery writes PAD frame.
Artio uses this for production FIX.

No checksums by default — shared memory failure mode is partial writes
from crash, not bit flips. Opt-in checksum flag for durable file-backed
journal (disk errors, torn page writeback, bit rot). Different failure
domain from the crash case.

## Liveness
For in-process journal (FIX same-process read/write), Tier 1 atomic
is sufficient. Cross-process readers use Tier 2 opt-in.

## API sketch
```rust
let (writer, reader) = Journal::open(path, config)?;

// Write (hot path, ~memcpy cost)
let mut claim = writer.try_claim(type_id, len)?;
claim.as_mut_slice().copy_from_slice(payload);
claim.commit();

// Read (sequential)
while let Some(record) = reader.next_record() {
    process(record.as_slice());
}
```

## References
- Aeron Archive — sentinel-based commit, PAD frame recovery
- Artio — production FIX on Aeron, validates the model
- nexus-logbuf — claim-based API pattern (WriteClaim/ReadClaim)

## Design doc
`docs/design/nexus-shm.md`

ChronoCoders:
Closed: implemented and merged via #416.

PR #414, opened 2026-05-31
docs: ShmJournal design notes (#391)

ChronoCoders:
Design-first for #391, same path as #389.

Doc: `docs/design/nexus-shm-journal.md`. Builds on the #390 foundation.

Most of #391 was already resolved in `nexus-shm.md` (commit ordering, checksum policy, liveness tier, segments). This pins the one open question and specifies the layout/protocols.

**Open question resolved → Path C (pluggable header).** `ShmJournal<H: RecordHeader>`, frames `[frame_len][H][payload]`. FIX uses `H = (seq, ts)` and gets `read_range`; non-FIX callers use `H = ()` at zero overhead. Path A forced a separate seq→position indexer for resend; Path B baked FIX semantics into every record. C is general without the waste — `read_range` lives behind `H: SeqHeader` so `()` callers never see it.

Spec covers: record layout with `i32` sign-as-sentinel commit marker (Aeron model), write/commit protocol (body → release fence → publish marker), zero-copy read path, forward-scan recovery, segment roll, PR scope.

Three open questions at the bottom for your call:
1. `i32` frame_len (sign = PAD sentinel) vs explicit marker byte + `u32` len.
2. Whether graceful drop ever needs to PAD-over, or "leave 0, recovery stops" suffices (SPSC → uncommitted claim is always the tail).
3. `read_range` borrowing iterator vs collected positions.

No code — review the plan and I'll implement.

### DIFF
--- a/docs/design/nexus-shm-journal.md
+++ b/docs/design/nexus-shm-journal.md
@@ -15,9 +15,10 @@ awareness — and specifies the record/commit/recovery/segment layout.
 - **Commit ordering** is the correctness property: write body → release fence →
   publish commit marker. Recovery scans forward, stops at first uncommitted
   record. (Aeron sentinel + PAD frame model.)
-- **Checksums** off by default; opt-in flag for the file-backed durability
-  domain (torn page writeback, bit rot). The commit marker alone covers the
-  crash case fully.
+- **Checksums** deferred. The reviewed frame layout carries no checksum field,
+  and a config flag wired to nothing is dead code; the commit marker alone covers
+  the crash case fully. A checksum (trailing CRC + its own `commit_len`
+  accounting) is a follow-up with its own layout review.
 - **Liveness:** in-process journal (FIX same-process read+write) needs only
   Tier-1 atomic. Cross-process read-only attach uses Tier-2 OFD opt-in.
 - **Segmented layout**, default 64 MB segment size.
@@ -26,8 +27,8 @@ awareness — and specifies the record/commit/recovery/segment layout.
 
 ## Open question resolved: sequence awareness → Path C (pluggable header)
 
-`ShmJournal<H: RecordHeader>`. The journal frames `[frame_len][H][payload]` and
-reads/writes `H` but never interprets it.
+`ShmJournal<H: RecordHeader>`. The journal frames `[frame header][H][payload]`
+and reads/writes `H` but never interprets it.
 
 - FIX (#411) uses `H = FixHeader { seq: u64, timestamp: u64 }` and gets
   `read_range` by sequence.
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0 (commented):
Good doc. Path C is the right call — general without waste, keeps FIX semantics behind a trait bound. Commit protocol and recovery follow Aeron correctly. One substantive issue and answers to the three open questions.

---

### frame_len must be unpadded

The doc says frame_len includes padding: "value is the total frame size (header + payload + padding, excluding the length word itself)." This makes payload length unrecoverable — you can't separate payload from padding without storing the payload length separately, which defeats the point of a single-field frame.

Aeron's convention is frame_len = unpadded size (header + payload only). The reader advances by `4 + align_8(frame_len)`. This gives `payload_len = frame_len - size_of::<H>()` — clean, no ambiguity, no extra field. Should follow that convention here.

---

### Open questions

**Q1 — i32 sign-as-sentinel:** Agree. Aeron model is proven, saves a field, single atomic store publish. The 2 GiB cap is well above the 64 MB segment size. No reason to deviate.

**Q2 — PAD on drop:** "Leave 0, recovery stops" is sufficient. In SPSC the uncommitted claim is always the tail — there's nothing past it to protect. PAD-on-drop would only matter if tooling needs to distinguish "crashed mid-write" from "hasn't written yet," and for recovery those are the same thing.

**Q3 — read_range:** Borrowing iterator. Matches `next_record()`, zero-copy, caller can `.collect()` if needed. FIX resend is effectively blocking on the session anyway — tying up the reader during retransmit is fine.

---

### Minor points

- `try_claim` returns `Option` in the API sketch but mentions `RecordTooLarge` in prose. Should this be `Result<WriteClaim, JournalError>` so the caller can distinguish "segment rolling" from "record exceeds segment size" from "journal closed"?
- Cross-process reader segment discovery — when the writer rolls to segment N+1, how does a read-only reader know to open the new file? Polling for `{base}.{N+1}`? A shared metadata region? Worth a sentence.
- Endianness of frame_len and FixHeader fields — if these files are read by external tools or across processes, document the byte order assumption (native LE on x86, or explicit LE always).

ChronoCoders:
Folded all in (8ecaf46): frame_len now unpadded (header + payload), reader advances `4 + align_8(frame_len)`, `payload_len = frame_len - size_of::<H>()` — agreed, the padded version made payload length unrecoverable. `try_claim` returns `Result<WriteClaim, JournalError>` (RecordTooLarge / Closed / rolling distinguishable). Added cross-process segment discovery (filename probe on tail-0) and the native byte-order contract. Q1/Q2/Q3 confirmed as you called them. Starting implementation.

PR #416, opened 2026-06-01
feat(shm): ShmJournal append-only durable log (#391)

ChronoCoders:
Implements #391. Design in #414 (approved); review feedback folded in.

## What

Append-only durable log over mmap segments on the #390 foundation. Aeron Archive model.

- `Journal<H>` — `open` returns `(Writer<H>, Reader<H>)`, SPSC.
- Pluggable `Pod` header: `()` for position-only logs at zero per-record overhead, `FixHeader { seq, timestamp }` for FIX. `read_range` is gated behind `H: SeqHeader`.
- `frame_len` i32 commit marker, **unpadded** (header + payload), sign-as-sentinel: `0` uncommitted/stop, `>0` committed, `<0` PAD. Reader advances `4 + align_8(frame_len)`.
- Write: reserve (marker stays 0) → `as_mut_slice()` → `commit()` writes header, zeroes the next marker slot, then Release-stores the marker.
- Read: Acquire-load marker, skip PAD, yield zero-copy payload.
- Forward-scan recovery to the first uncommitted tail; PAD-over segment roll; cross-process segment discovery via `{base}.{N+1}` filename probe.

## Verification

16 tests (roundtrip, `()` header, multi-segment roll, uncommitted-tail recovery, range query, too-large/empty rejection), clippy clean under `-D warnings`. Criterion bench: `try_claim`+`commit` ~135 ns, `next_record` drain ~21 M/s (unpinned dev box, indicative only).

## Two deviations from the design doc — want your call

1. **No checksum.** The reviewed frame layout has no checksum field, and a config flag that does nothing violates the no-dead-code bar. Dropped it; recovery relies on the commit marker alone, which the doc already states fully covers the crash case. Checksum can come as a follow-up with its own layout (trailing CRC, `frame_len` accounting) for review.
2. **`ReadRecord::header()` returns `H` by value, not `&H`.** With an i32 marker the header sits at a 4-aligned offset, so `&H` would be unaligned UB for `FixHeader` (align 8). `H: Pod: Copy` and headers are tiny; payload stays zero-copy `&[u8]`. The alternative is an 8-byte marker cell, which I avoided to keep the Aeron i32 convention you confirmed.

Out of scope (noted in the doc): O(1) ring index, MPSC, fsync policy.

### DIFF
--- a/nexus-shm/Cargo.toml
+++ b/nexus-shm/Cargo.toml
@@ -23,3 +23,7 @@ criterion = { version = "0.5", features = ["html_reports"] }
 [[bench]]
 name = "ofd_liveness"
 harness = false
+
+[[bench]]
+name = "journal"
+harness = false
--- a/nexus-shm/benches/journal.rs
+++ b/nexus-shm/benches/journal.rs
@@ -77,7 +77,7 @@ fn bench_read(c: &mut Criterion) {
         b.iter_batched(
             || Journal::<FixHeader>::open(&base, cfg(1 << 20)).unwrap(),
             |(w, mut r)| {
-                while let Some(rec) = r.next_record() {
+                while let Some(rec) = r.next_record().unwrap() {
                     black_box(rec.payload());
                 }
                 drop((w, r));
--- a/nexus-shm/benches/journal.rs
+++ b/nexus-shm/benches/journal.rs
@@ -1,0 +1,81 @@
+use std::hint::black_box;
+
+use criterion::{BatchSize, Criterion, Throughput, criterion_group, criterion_main};
+use nexus_shm::{FixHeader, Journal, JournalConfig, MapOptions};
+
+fn cfg(segment_size: usize) -> JournalConfig {
+    JournalConfig {
+        segment_size,
+        map: MapOptions::default(),
+    }
+}
+
+fn cleanup(base: &std::path::Path) {
+    for i in 0..64u64 {
+        let mut p = base.as_os_str().to_owned();
+        p.push(format!(".{i}"));
+        let _ = std::fs::remove_file(std::path::PathBuf::from(p));
+    }
+}
+
+fn bench_write(c: &mut Criterion) {
+    let base = std::env::temp_dir().join(format!("nexus-journal-bench-w-{}", std::process::id()));
+    cleanup(&base);
+
+    let (mut w, _r) = Journal::<FixHeader>::open(&base, cfg(1 << 31)).unwrap();
+    let payload = [0u8; 32];
+    let mut seq = 0u64;
+
+    c.benchmark_group("journal").bench_function("try_claim_commit", |b| {
+        b.iter(|| {
+            seq += 1;
+            let mut claim = w
+                .try_claim(FixHeader { seq, timestamp: seq }, payload.len())
+                .unwrap();
+            claim.as_mut_slice().copy_from_slice(&payload);
+            claim.commit();
+        });
+    });
+
+    drop((w, _r));
+    cleanup(&base);
+}
+
+fn bench_read(c: &mut Criterion) {
+    let base = std::env::temp_dir().join(format!("nexus-journal-bench-r-{}", std::process::id()));
+    cleanup(&base);
+
+    const N: u64 = 1000;
+    {
+        let (mut w, _r) = Journal::<FixHeader>::open(&base, cfg(1 << 20)).unwrap();
+        let payload = [0u8; 32];
+        for seq in 1..=N {
+            let mut claim = w
+                .try_claim(FixHeader { seq, timestamp: seq }, payload.len())
+                .unwrap();
+            claim.as_mut_slice().copy_from_slice(&payload);
+            claim.commit();
+        }
+    }
+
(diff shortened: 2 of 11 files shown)


Abso1ut3Zer0 (changes requested):
**Overall:** Solid implementation — pluggable header is clean, WriteClaim pattern works well. The two deviations (no checksum, header by value) are both correct calls. One structural issue with the frame layout needs fixing before merge, plus a few smaller items.

---

### Frame layout: sentinel collision bug (must fix)

The `i32` sign-as-sentinel has an edge case. When `remaining == LEN_SIZE` (4 bytes left at segment end), `roll()` stores `-(0 as i32)` = `0` as the PAD marker. The reader interprets `0` as "no more records" and never advances to the next segment. Records in subsequent segments become unreachable.

Reproducible at minimum segment size: `segment_size=64`, `()` header, `payload_len=4` → footprint=12. After 5 records, `tail=60`, `remaining=4`. Bug.

The root cause is the 4-byte marker making footprints `4 + 8k` (4 mod 8), so remaining can be exactly 4 — marker-sized but with a body of 0, colliding with the uncommitted sentinel.

The fix I'd suggest is adopting the Aeron model more fully: an 8-byte frame header with a type discriminant.

```
[u32 commit_len] [u16 frame_type] [u16 flags]
```

- `commit_len = 0` → uncommitted (always unambiguous)
- `commit_len > 0, type = DATA (0)` → record
- `commit_len > 0, type = PAD (1)` → skip, advance by `align8(commit_len)`
- `flags` → reserved for future use (CRC, compression, etc.)

With an 8-byte header and 8-byte alignment, footprint = `8 + align8(body)` = always a multiple of 8. Remaining space is always 0 or >= 8, so a PAD header always fits. The collision becomes structurally impossible — same invariant Aeron uses (header size == alignment), just 8 instead of 32.

`commit_len` becomes `u32` (no sign trick needed), which also drops the implicit ~2GB segment size cap from the `i32` cast in the PAD calculation.

This is a bigger change, but the layout isn't shipped yet — better to get it right now.

---

### `load_next()` swallows all errors

```rust
Err(_) => false
```

File-not-found (no more segments) and permission-denied / disk-failure / corrupted-segment all silently return `false`. At minimum, distinguish `ErrorKind::NotFound` from real errors — the current code would silently lose data if a segment file has wrong permissions.

---

### Missing doc comments on public types

`Writer`, `Reader`, `WriteClaim`, `ReadRecord`, `JournalConfig`, `Journal`, `RecordHeader`, `SeqHeader` — these are all public but have no doc comments. The PR body is well-documented but the types themselves need it.

---

### Nit: error display

`"record frames to zero bytes"` → `"empty record"` or `"record payload is empty"`. Current wording is a bit unclear.

Abso1ut3Zer0:
One more note on the frame layout: the `i32` marker in the original design came from Aeron, which needs to support Java (no unsigned types — our best guess for why it's signed). Since we don't have that constraint, we have more freedom with the header layout. The `u32 + u16 + u16` split gives us an unsigned length, a type discriminant, and room for flags, all in the same 8 bytes.

ChronoCoders:
Thanks — good catch on the sentinel collision. All four addressed in the latest push:

1. **Frame layout** — adopted your 8-byte header `[u32 commit_len][u16 frame_type][u16 flags]`. Footprint is now always a multiple of 8, so leftover tail space is 0 or ≥8 and a PAD frame always fits — the `remaining == marker-size` collision is structurally impossible. Dropped the i32 2 GiB cap (u32 length, no sign trick). Added a regression test (`pad_at_frame_header_boundary`) that drives `remaining == FRAME_HEADER` exactly and asserts every record stays reachable across the roll.
2. **`load_next`** — now returns `Result`; distinguishes `NotFound` (caught up) from real I/O errors. `next_record` is `Result<Option<_>>` so faults surface instead of looking like end-of-log; `read_range` pre-loads so it just gained a `?`.
3. **Doc comments** — added on all the public types, one line each, matching the `segment.rs` idiom.
4. **Nit** — `EmptyRecord` message is now `"empty record"`.

Design doc (#414) updated to match. `fmt`/`clippy -D warnings`/tests all green.

Abso1ut3Zer0 (approved):
All four items addressed cleanly. Frame layout change is structurally sound — `FRAME_HEADER == ALIGN == 8` makes remaining always 0 or ≥ 8, so the sentinel collision is impossible by construction. The regression test (`pad_at_frame_header_boundary`) nails the exact boundary. `load_next` error handling, doc comments, and the nit are all good.

One non-blocking nit:

**Segment size cap**: `roll()` stores `remaining as u32` for the PAD `commit_len`. If `segment_size > u32::MAX` (~4 GiB), the cast silently truncates — potentially back to 0, recreating the sentinel collision. The DATA path already guards this (`body > u32::MAX as usize`), but the PAD path relies on `remaining < segment_size`. Adding a cap in `open()` (e.g. error if `segment_size > u32::MAX as usize`) would close the gap, since the frame format can't represent larger values anyway. Not blocking — nobody's using 5 GiB segments — but worth sealing.

Issue #407, opened 2026-05-31
nexus-fix-codegen: dictionary-driven code generator


Abso1ut3Zer0:
## Overview

Standalone binary + library that reads QuickFIX XML data dictionaries and
writes readable `.rs` files. Can be invoked as a CLI tool (like protoc/SBE)
or called as a library from a `build.rs`.

Design doc: `docs/design/nexus-fix-codec.md` (codegen section)

### XML dictionary parsing

Parse QuickFIX XML format (de facto standard — every FIX exchange publishes one):
- `<fields>` — tag number, name, type, enum values
- `<messages>` — message type (tag 35 value), required/optional fields, component refs
- `<components>` — reusable field groups (e.g., Instrument, Parties)
- `<groups>` — repeating group definitions with delimiter tag

### Generated output (per dictionary)

**`fields.rs`** — Tag number constants (`pub const TAG_SYMBOL: u32 = 55;`) and
typed enum types (`Side`, `OrdType`, etc.) with `from_byte`/`as_byte` methods.

**`messages.rs`** — Per-message-type flyweight decoder structs. Each has:
- `Cell<FieldSpan>` slots per field (interior mutability for `&self` accessors)
- `Cell<u32>` watermark tracking scan progress
- Progressive scan: scan forward on demand, cache every tag encountered
- Typed accessors returning `Option<&'buf [u8]>`, `Option<&'buf str>`, or typed enums
- Tag dispatch via `match` in the scan loop — compiler picks jump table or binary search

**`groups.rs`** — Repeating group iterators and entry decoders. Count tag triggers
group walking. Entries are small (5-10 fields), linear scan within an entry.
Nested groups follow the same pattern recursively.

**`encoders.rs`** — Builder-pattern encoders per message type. No validation at
the codec layer (SBE model). Consume-self chaining with `finish() -> usize`.

**`mod.rs`** — Re-exports, MsgType dispatch.

### DATA field handling

The codegen must emit special scanner logic for LENGTH/DATA tag pairs (95/96,
90/91, 212/213, 348/349, 358/359). DATA fields can contain embedded SOH and `=`
— the scanner must consume by length, not by SOH scan. The dictionary identifies
these pairs (type DATA preceded by a LENGTH field).

### Watermark "not present" semantics

FIX body field order is not guaranteed on the wire. A field is only declared
absent when `scanned_to == end_of_message` and the field was not cached. The
generated scanner must scan to end-of-message before returning `None`.

### Invocation

[...]

Issue #406, opened 2026-05-31
nexus-fix-codec: core scanning and encoding library


Abso1ut3Zer0:
## Overview

Core library crate providing the runtime primitives that all generated FIX
codecs depend on. Zero-copy, SIMD-accelerated, no allocation.

Design doc: `docs/design/nexus-fix-codec.md`

### FieldSpan / GroupSpan

`FieldSpan { offset: u32, len: u32 }` — 8-byte pointer into the message buffer.
`u32` length accommodates DATA-type fields (RawData, XmlData, EncodedText) that
can exceed 64KB. `GroupSpan { offset: u32, count: u16 }` for repeating group
locations.

### SIMD SOH scanning

`find_soh(buf, pos) -> Option<usize>` and `find_eq(buf, pos) -> Option<usize>`.
SSE2 baseline (`cmpeq_epi8` + `movemask_epi8`, 16 bytes/iter), AVX2 (32 bytes/iter).
Scalar fallback for tail. These are the innermost loop of the parser.

### SIMD checksum

FIX checksum = sum of all bytes mod 256. `PSADBW` against zero vector for
horizontal byte sums in 8-byte chunks. Not CRC32 — no dedicated hardware
instruction.

`checksum(buf, start, end) -> u8` and `validate_checksum(msg) -> Result<(), ChecksumError>`.

### Tag number parsing

`parse_tag(buf) -> (u32, usize)`. ASCII integer parsing on the hot path.
Scalar multiply-accumulate baseline; SWAR for known-length tags if beneficial.

### Encoder primitives

`encode_field(buf, pos, tag, value) -> usize` — writes `tag=value\x01`.
`frame_message(buf, body_start, body_end, begin_string) -> usize` — writes
header (8=, 9=) and computes/writes trailer (10=).

### Error types

`DecodeError`, `ChecksumError`. Minimal, no thiserror (workspace convention).

## Performance targets

| Operation | Target |
|-----------|--------|
| SOH scan + checksum (SIMD) | < 100ns |
| Field access (cached) | < 10ns |
| Field access (watermark scan) | < 50ns |
| Outbound encode | < 300ns |

## Dependencies

None beyond std. This is a leaf crate.

Abso1ut3Zer0:
Shipped in #415. Core read/write primitives with SIMD acceleration — FieldReader, FieldWriter, scan cascade, fused PSADBW checksum, GroupSpan pattern. Framing pushed to codegen layer (deviation documented in PR).

PR #418, opened 2026-06-01
docs: nexus-fix-codegen design (#407)

ChronoCoders:
Design/plan for #407 (nexus-fix-codegen), now that #406 is merged. No code yet — posting the shape first for your steer before I build the generator, per the same flow as the ShmJournal design (#414 → #416).

Full plan in `docs/design/nexus-fix-codegen.md`. Summary:

- Standalone generator (lib + `clap` CLI + `build.rs` API) reading QuickFIX XML, emitting readable Rust on top of the merged codec.
- Output: `fields.rs` (TAG consts + typed enums, single- vs multi-char `from_byte`/`from_bytes`), `messages.rs` (flyweight decoders over `FieldReader`/`find_tag` into `FieldSpan` slots), `groups.rs` (`GroupSpan` + bounded sub-scan), `encoders.rs` (`FieldWriter`/`encode_field`/`format_checksum`), `mod.rs`.
- Dictionary-driven, unknown tags skipped, DATA fields length-delimited, no encoder-side validation.
- **Modeled on Prost** (your pointer): copy its tooling — `Config` builder, `build.rs` + `OUT_DIR`/`include!`, checked-in CLI — but keep our **zero-copy flyweight** decode (Prost copies into owned structs; we point spans into the buffer).

Open questions for you in the doc: XML parser (`quick-xml` lean), rustfmt-as-post-step, group nesting depth, and whether to keep the `Cell<FieldSpan>` watermark layer given `FieldReader` is already single-pass.

### DIFF
--- a/docs/design/nexus-fix-codegen.md
+++ b/docs/design/nexus-fix-codegen.md
@@ -63,6 +63,32 @@ Generated output:
    should generated decoders just expose the iterator + typed accessors and let
    the caller pass once? Depends on #406's intended ergonomics.
 
+## Modeled after Prost (Michael's pointer)
+
+Prost is the model for the **tooling**, not the decode model.
+
+Adopt from `prost` / `prost-build`:
+- **`build.rs` integration** — a `Config`-style builder (`generate().out_dir(..)
+  .dictionary(..).run()`) mirroring `prost_build::Config::compile_protos`, so a
+  consumer crate generates in `build.rs` and pulls the result with
+  `include!(concat!(env!("OUT_DIR"), "/fix.rs"))`.
+- **Generated code is plain, readable Rust** — no runtime reflection, no dynamic
+  dictionary at runtime; the dictionary is consumed entirely at generation time.
+- **Builder knobs** — type attributes / derives, module layout, opt-in
+  per-message selection, analogous to Prost's `type_attribute` /
+  `btree_map` style configuration.
+- **A checked-in CLI** alongside the `build.rs` path, like `protoc` + `prost`'s
+  split, for ahead-of-time generation into the repo when preferred.
+
+Diverge from Prost where the data model differs:
+- Prost decodes length-delimited binary into **owned structs** (it copies). FIX
+  here is **zero-copy flyweight** — generated "messages" are decoders holding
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0 (approved):
Good shape. Answers to the open questions and one design simplification:

### Q1: XML parser — `quick-xml`

Agreed. Pull-based, fast, no DOM allocation, widely supported. No reason to use a heavier parser for build-time codegen.

### Q2: Formatting — post-step `rustfmt`

Agreed. Pre-formatting generated code is a maintenance problem. Prost does exactly this. Graceful degradation if rustfmt isn't found is fine for the `build.rs` path.

### Q3: Group nesting — fully recursive, no depth guard

No existing FIX codegen library (QuickFIX/J, QuickFIX C++, QuickFIX/n, QuickFIX/Go, SBE) imposes a depth limit. They all follow the dictionary. Real-world maximums: FIX 4.2 tops at 2, FIX 4.4 at 4 (NestedParties chains), FIX 5.0 SP2 at 5 (ComplexEvents). The dictionary constrains depth naturally. Just recurse — no guard needed.

### Q4: Watermark — drop it, checksum forces a full pass

This is the key simplification. FIX requires a mod-256 checksum over every byte in the message body. You're doing the full pass regardless — so the lazy watermark buys nothing. Fill all `FieldSpan` slots during that single mandatory pass.

This means:
- No `Cell<FieldSpan>` — plain `FieldSpan` fields, populated during the scan
- No `Cell<u32>` watermark
- No `find_tag` in generated code — every tag is dispatched into its slot during the checksum pass
- Accessors become pure reads: `fn symbol(&self) -> Option<&'buf [u8]>` checking `span.is_present()`. No mutation, no interior mutability.
- One pass, two jobs (checksum + decode), zero wasted work

`find_tag` stays in the codec as a convenience for ad-hoc use without generated code, but generated decoders shouldn't touch it.

### Minor note

The doc says "random-access fields use `codec::find_tag`" — with the full-scan approach this is no longer relevant. Every field is already found during the single pass. Worth clarifying in the next revision.

---

Overall the Prost-tooling + zero-copy-flyweight split is well articulated. Good to build once the journal PR lands.

PR #426, opened 2026-06-02
feat: nexus-fix-codegen — dictionary-driven code generator (#407)

### LINKED ISSUE #407: nexus-fix-codegen: dictionary-driven code generator

ChronoCoders:
Closes #407. Depends on #406 (`nexus-fix-codec` runtime). Implements the design approved in #418.

## What

New crate `nexus-fix-codegen`: a lib + `clap` CLI + `build.rs` API that reads a QuickFIX dictionary XML and emits readable Rust over the #406 codec primitives. Generated code is zero-copy, zero-alloc, no runtime reflection — the dictionary is consumed entirely at generation time (Prost tooling model).

```
nexus-fix-codegen/
  src/
    lib.rs    — Config builder + generate()
    main.rs   — clap CLI
    dict/     — quick-xml dictionary parse -> typed model
    emit/     — fields / messages / groups / encoders / mod emitters
```

## Generated output

- **`fields.rs`** — `TAG_*: u32` consts; typed enums per `<field>`. Single-char -> `from_byte`/`as_byte`; multi-char -> `from_bytes`/`as_bytes`. Lookups key off `RawField.value.slice(buf)`.
- **`messages.rs`** — per-`MsgType` flyweight decoder. One-pass populate: run `FieldReader` once, dispatch each `RawField.tag` into the message's `FieldSpan` slots. O(n), zero-copy, borrows `&[u8]`. No watermark, no `Cell` — accessors are pure reads checking `span.is_present()`.
- **`groups.rs`** — repeating-group iterators + per-entry decoders over `GroupSpan` + a bounded sub-scan. Fully recursive nesting.
- **`encoders.rs`** — consume-self builders over `FieldWriter` / `encode_field`, `finish() -> usize`, `finish_with_checksum()`.
- **`mod.rs`** — re-exports, `BEGIN_STRING`, `MsgType` dispatch.

## Design (per #418)

- **Dictionary-driven, not version-driven** — one codepath for any QuickFIX XML.
- **Unknown tags: silent skip** — no slot, no error.
- **Decode model: one-pass populate, no watermark.** `find_tag` stays in the codec for ad-hoc use; generated decoders don't call it.
- **DATA fields** (95/96, 90/91, 212/213, 348/349, 358/359) — read length-delimited: take the preceding length tag, consume exactly N bytes, re-seat `FieldReader` past the value so a `0x01` in a binary payload never mis-splits. DATA inside a repeating group is rejected at generation time (`EmitError::DataInGroup`) rather than emitting subtly-wrong code.
- **No `#[allow]` in generated code** — the output is clean under the workspace's `clippy::all/pedantic/nursery -D warnings` with zero suppressions.

## Resolved open questions (#418)

`quick-xml` parser · post-step `rustfmt` (graceful degrade if absent) · fully-recursive group nesting, no depth guard.

## Tests

- `nexus-fix-codegen` unit tests: dictionary parse + emitted-source assertions + DATA-in-group rejection.
- `nexus-fix-codegen-tests`: a `build.rs` consumer that generates from a sample dictionary into `OUT_DIR` and `include!`s it, then exercises decode (scalars, enums, DATA with embedded SOH, repeating + nested groups), `MsgType` dispatch, and encode round-trips. This is the end-to-end proof that generated code compiles and decodes correctly.

## Out of scope

Session/transport (#409/#410), persistence (#411). Targets the codec layer only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ members = [
 	"nexus-async-rt",
 	"nexus-inference",
 	"nexus-fix-codec",
+	"nexus-fix-codegen",
+	"nexus-fix-codegen-tests",
 ]
 
 [workspace.package]
--- a/nexus-fix-codec/Cargo.toml
+++ b/nexus-fix-codec/Cargo.toml
@@ -14,6 +14,7 @@ categories = ["parsing", "network-programming"]
 workspace = true
 
 [dependencies]
+nexus-ascii = { path = "../nexus-ascii" }
 
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
--- a/nexus-fix-codec/src/convert.rs
+++ b/nexus-fix-codec/src/convert.rs
@@ -1,0 +1,109 @@
+//! Parsing helpers for typed FIX field values.
+
+/// Parse a FIX integer (`INT`, `SEQNUM`) value. Returns `None` on empty
+/// input, non-digit bytes, or overflow.
+#[inline]
+pub fn parse_fix_int(bytes: &[u8]) -> Option<i64> {
+    let (neg, digits) = match bytes.first()? {
+        b'-' => (true, &bytes[1..]),
+        b'+' => (false, &bytes[1..]),
+        _ => (false, bytes),
+    };
+    if digits.is_empty() {
+        return None;
+    }
+    let mut acc: i64 = 0;
+    for &b in digits {
+        let d = b.wrapping_sub(b'0');
+        if d > 9 {
+            return None;
+        }
+        acc = acc.checked_mul(10)?.checked_add(d as i64)?;
+    }
+    Some(if neg { -acc } else { acc })
+}
+
+/// Parse a FIX unsigned integer (`LENGTH`, `NUMINGROUP`) value. Returns
+/// `None` on empty input, non-digit bytes, or overflow.
+#[inline]
+pub fn parse_fix_uint(bytes: &[u8]) -> Option<u32> {
+    if bytes.is_empty() {
+        return None;
+    }
+    let mut acc: u32 = 0;
+    for &b in bytes {
+        let d = b.wrapping_sub(b'0');
+        if d > 9 {
+            return None;
+        }
+        acc = acc.checked_mul(10)?.checked_add(d as u32)?;
+    }
+    Some(acc)
+}
+
+/// Format an unsigned integer into `buf` as ASCII digits, returning the
+/// written slice. `buf` of 20 bytes holds any `u64`. Allocation-free.
+#[inline]
+pub fn format_uint(buf: &mut [u8; 20], mut v: u64) -> &[u8] {
+    let mut p = buf.len();
+    loop {
+        p -= 1;
+        buf[p] = b'0' + (v % 10) as u8;
+        v /= 10;
+        if v == 0 {
+            break;
+        }
+    }
+    &buf[p..]
+}
+
+/// Parse a FIX boolean (`BOOLEAN`) value: `Y` is true, `N` is false.
+#[inline]
+pub fn parse_fix_bool(bytes: &[u8]) -> Option<bool> {
+    match bytes {
+        b"Y" => Some(true),
+        b"N" => Some(false),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn ints() {
+        assert_eq!(parse_fix_int(b"0"), Some(0));
+        assert_eq!(parse_fix_int(b"42"), Some(42));
+        assert_eq!(parse_fix_int(b"-17"), Some(-17));
+        assert_eq!(parse_fix_int(b"+5"), Some(5));
+        assert_eq!(parse_fix_int(b""), None);
+        assert_eq!(parse_fix_int(b"-"), None);
+        assert_eq!(parse_fix_int(b"1a"), None);
+    }
+
+    #[test]
+    fn uints() {
+        assert_eq!(parse_fix_uint(b"123"), Some(123));
+        assert_eq!(parse_fix_uint(b""), None);
+        assert_eq!(parse_fix_uint(b"-1"), None);
+        assert_eq!(parse_fix_uint(b"99999999999"), None);
+    }
+
+    #[test]
+    fn uint_format() {
+        let mut buf = [0u8; 20];
+        assert_eq!(format_uint(&mut buf, 0), b"0");
+        assert_eq!(format_uint(&mut buf, 7), b"7");
+        assert_eq!(format_uint(&mut buf, 12345), b"12345");
+        assert_eq!(format_uint(&mut buf, u64::from(u32::MAX)), b"4294967295");
+    }
+
+    #[test]
+    fn bools() {
+        assert_eq!(parse_fix_bool(b"Y"), Some(true));
+        assert_eq!(parse_fix_bool(b"N"), Some(false));
+        assert_eq!(parse_fix_bool(b"y"), None);
+        assert_eq!(parse_fix_bool(b""), None);
+    }
+}
--- a/nexus-fix-codec/src/lib.rs
+++ b/nexus-fix-codec/src/lib.rs
@@ -10,6 +10,7 @@
 //!
 //! Generated FIX codecs (from `nexus-fix-codegen`) depend on these primitives.
 
+mod convert;
 mod error;
 mod span;
 
@@ -17,7 +18,9 @@ pub mod reader;
 pub mod scan;
 pub mod writer;
 
+pub use convert::{format_uint, parse_fix_bool, parse_fix_int, parse_fix_uint};
 pub use error::{ChecksumError, DecodeError};
+pub use nexus_ascii::AsciiTextStr;
 pub use reader::{FieldReader, RawField, checksum, find_tag, parse_tag, validate_checksum};
 pub use scan::DelimiterScanner;
 pub use span::{FieldSpan, GroupSpan};
--- a/nexus-fix-codegen-tests/Cargo.toml
+++ b/nexus-fix-codegen-tests/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "nexus-fix-codegen-tests"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+nexus-fix-codec = { path = "../nexus-fix-codec" }
+
+[build-dependencies]
+nexus-fix-codegen = { path = "../nexus-fix-codegen" }
--- a/nexus-fix-codegen-tests/build.rs
+++ b/nexus-fix-codegen-tests/build.rs
@@ -1,0 +1,14 @@
+use std::path::PathBuf;
+
+fn main() {
+    let out = std::env::var("OUT_DIR").expect("OUT_DIR not set");
+    let dict = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../nexus-fix-codegen/tests/fixtures/FIX_sample.xml");
+    println!("cargo:rerun-if-changed={}", dict.display());
+    nexus_fix_codegen::generate()
+        .dictionary(&dict)
+        .out_dir(&out)
+        .rustfmt(false)
+        .run()
+        .expect("codegen failed");
+}
--- a/nexus-fix-codegen-tests/build.rs
+++ b/nexus-fix-codegen-tests/build.rs
@@ -2,11 +2,15 @@ use std::path::PathBuf;
 
 fn main() {
     let out = std::env::var("OUT_DIR").expect("OUT_DIR not set");
-    let dict = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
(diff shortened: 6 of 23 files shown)


Abso1ut3Zer0:
Solid foundation — dictionary-driven, one-pass populate, flyweight decoders, SIMD reader underneath. The architecture is right. Below is what I'd like to see addressed in this PR vs. what can follow.

## Must address

### Correctness

**DATA field decode can panic on truncated input** — the generated `data_arm` (`emit/messages.rs:156-173`) does manual buffer arithmetic that bypasses `FieldReader`'s bounds safety. If the length tag says `95=100` but the buffer is truncated, `FieldSpan::slice()` will panic on access. Fix: clamp `n` to available buffer length.

**Group count `as u16` truncation** — `parse_tag` returns `u32` but group count is cast with `count as u16` (e.g. `emit/messages.rs:188`). Silent truncation if count > 65535. Add a `debug_assert` or saturating cast.

### Type system

Right now every accessor returns `Option<&[u8]>` regardless of FIX type. FIX fields have well-defined types and the generated code should reflect that.

**String fields → `&AsciiTextStr`** — FIX STRING is printable ASCII (0x20-0x7E), which is exactly what `nexus_ascii::AsciiTextStr` validates. Validate during decode (SIMD, cheap), then accessors use `from_bytes_unchecked`. Re-export `AsciiTextStr` from `nexus-fix-codec` so consumers only need one dep. Applies to: STRING, CHAR, EXCHANGE, CURRENCY, COUNTRY, and for now FLOAT/PRICE/QTY/AMT/PERCENTAGE/timestamps (we'll add richer types for those later).

**Parsed integer accessors** — add `parse_fix_int(bytes) -> Option<i64>` and `parse_fix_bool(bytes) -> Option<bool>` helpers to `nexus-fix-codec`. INT/SEQNUM → `Option<i64>`, LENGTH/NUMINGROUP → `Option<u32>`, BOOLEAN → `Option<bool>`. DATA stays `&[u8]`.

**Enum `Unknown` variant** — FIX counterparties send values not in your dictionary. Without `Unknown`, `from_byte` returns `Option` and every call site needs fallback logic.

Single-char enums: `Unknown(u8)` carries the wire byte. Multi-char enums: `Unknown(&AsciiTextStr)` carries the validated text. `from_byte`/`from_bytes` become infallible — return `Self`, not `Option`. Example:

```rust
// Single-char enum (no lifetime, Copy)
pub enum Side {
    Buy,
    Sell,
    Unknown(u8),
}

impl Side {
    pub fn from_byte(b: u8) -> Self {
        match b {
            b'1' => Self::Buy,
            b'2' => Self::Sell,
            _ => Self::Unknown(b),
        }
    }
}

// Multi-char enum (borrows from buffer)
pub enum OrdType<'buf> {
    Market,
    Limit,
    Unknown(&'buf AsciiTextStr),
}

impl<'buf> OrdType<'buf> {
    pub fn from_bytes(b: &'buf AsciiTextStr) -> Self {
        match b.as_bytes() {
            b"1" => Self::Market,
            b"2" => Self::Limit,
            _ => Self::Unknown(b),
        }
    }
}
```

### Codegen quality

**`if/else if` → `match` for tag dispatch** — the generated decode uses an if/else if chain (`messages.rs:30-59` in output). The compiler can't optimize this into a jump table. A `match f.tag { TAG_X => ..., _ => {} }` enables the compiler to pick the optimal dispatch (jump table, binary search tree, etc). This matters for real dictionaries where messages have 50+ fields. Same fix needed in group entry decode.

**`[tags].contains()` → `matches!()` for group membership** — the generated `[448u32, 452, 802, ...].contains(&gf.tag)` is a linear scan. `matches!(gf.tag, 448 | 452 | 802 | ...)` lets the compiler optimize it.

**`clap` behind a feature flag** — `clap` (+ ~10 transitive crates) is an unconditional dependency but only used by the CLI binary. When used as a `build-dependency` (the primary use case), all of clap gets compiled for nothing. Put it behind a `cli` feature with `required-features = ["cli"]` on the binary target.

**Heap allocation in DATA encoder** — `value.len().to_string()` in the generated encoder allocates. Use a stack buffer for the digit conversion.

**Deduplicate `tag_array`** — identical function in `emit/messages.rs:249` and `emit/groups.rs:202`. Move to `emit/mod.rs`.

## Defer (separate PRs)

- **Decimal/price type strategy** — FLOAT/PRICE/QTY/AMT/PERCENTAGE return `&AsciiTextStr` for now. Need a design discussion about whether codegen accepts a configured decimal type or we add parsing helpers to the codec.
- **Timestamp parsing** — same, `&AsciiTextStr` for now. Potential `FixTimestamp` struct later.
- **Multi-dictionary module naming** — `.dictionary_with_module(path, name)` API and generated root wiring. Current file-stem approach works.
- **Encoder group support** — encoders skip repeating groups. Document as known limitation.
- **`REQUIRED` const** — exists but has no corresponding validation method. Either add `is_complete()` or remove the dead constant.
- **Benchmarks + assembly audit** — critical for a trading codec, but separate PR. Depends on the `match` dispatch fix landing first.

Abso1ut3Zer0 (commented):
I've read through the full PR. Architecture is solid — dictionary-driven codegen, one-pass decode, consume-self encode, fully recursive groups, DATA length-delimited handling, multi-dictionary namespacing. The Prost tooling model is the right call.

Issues below, roughly priority-ordered.

---

### 1. `convert.rs` duplicates SWAR-optimized parsers (must fix)

`convert.rs` adds scalar `parse_fix_int`, `parse_fix_uint`, `parse_fix_bool`, and `format_uint` to `nexus-fix-codec`. These already exist in `types.rs` (landed on main) with SWAR-optimized digit parsing — 30–44 cycles vs scalar per-digit loops. `format_uint` also overlaps with the new `encode_fix_uint`/`encode_fix_seqnum` which use a digit-pair LUT.

**Resolution:** Drop `convert.rs` entirely. The generated code should call the existing `parse_fix_int`, `parse_fix_uint`, `parse_fix_bool` from `types.rs`. For the DATA encoder (`format_uint` call in `encoders.rs`), use `encode_fix_uint` or `encode_fix_seqnum` instead.

This is also the main merge conflict — the `lib.rs` re-exports diverge.

### 2. `parse_tag` misused as a value parser (bug risk)

`messages.rs` emits `nexus_fix_codec::parse_tag(f.value.slice(buf))` to parse LENGTH field values and group counts. `parse_tag` is the reader's tag-number parser — it works here by coincidence (both are unsigned digit sequences), but it's the wrong abstraction. If `parse_tag` ever gains tag-specific behavior (max width, leading-zero handling, etc.), the codegen silently breaks.

Should emit `nexus_fix_codec::parse_fix_uint(f.value.slice(buf)).unwrap_or(0)` instead. Same applies in `group_body()` for the count extraction.

### 3. `nexus-ascii` as non-optional dependency on nexus-fix-codec

The PR adds `nexus-ascii` as a hard dependency and re-exports `AsciiTextStr` from nexus-fix-codec. The generated code needs it for string field accessors, which makes sense — FIX values are ASCII. But this couples the codec layer to nexus-ascii for all consumers, even those not using codegen.

Two options:
- Feature-gate it: `ascii = ["dep:nexus-ascii"]` on nexus-fix-codec, codegen enables it
- Have the generated code import `nexus_ascii::AsciiTextStr` directly instead of going through nexus-fix-codec

I lean toward the second — the generated code already takes `nexus_fix_codec` as a dependency; adding `nexus_ascii` is one more line in the consumer's Cargo.toml and keeps the codec layer dependency-free.

### 4. Dictionary field lookup is O(n)

`Dictionary::field()` does a linear scan over all fields. For a FIX 4.4 dictionary (~950 fields), this runs during resolution of every field reference in every message — potentially thousands of lookups. Not a hot path (generation-time), but a `HashMap<&str, &FieldDef>` built once would be cleaner and ~100x faster. Same for `component()`.

### 5. DATA handling is fragile

The DATA body emission (`data_body()` in `messages.rs`) manually re-parses the buffer to find the DATA field's value offset:
```rust
let dstart = r.pos();
let (_, dtl) = parse_tag(&buf[dstart..]);
let vstart = dstart + dtl + 1;
```
This re-implements tag parsing logic that `FieldReader` owns. If the wire format has any deviation (e.g., tag number with leading whitespace, which some broken counterparties send), this diverges from what `FieldReader` would do.

Also, the `.min()` clamp silently truncates oversized DATA — the caller has no signal that data was lost. Consider at least documenting this behavior or providing a `data_len()` accessor alongside `raw_data()` so callers can detect truncation.

### 6. No typed accessors (expected, but flag for #427)

All field accessors return `&[u8]`, `AsciiTextStr`, `i64`, `u32`, or `bool`. No integration with the domain types (`FixDecimal`, `FixTimestamp`, etc.) from #427. This is fine for V1 — the domain type system is designed to layer on top — but the generated code should be the primary consumer of those types in the next iteration. The accessor emitter (`emit_value_accessor`) would gain `FieldType::Price | FieldType::Qty | FieldType::Amt` arms that return `FixDecimal`.

### Nits

- `run_rustfmt` silently swallows failures. Even a `eprintln!` on non-zero exit would save debugging time when the generated output doesn't compile due to formatting issues.
- `FieldType::from_str` shadows the `FromStr` trait name — not blocking, just mildly surprising. `from_quickfix_type` would be clearer.
- No `#[must_use]` on `Config::new()`.

### Overall

The design decisions are sound — silent skip for unknown tags, one-pass populate without watermark, `DataInGroup` rejection at generation time, fully recursive nesting. The two-venue integration test is good proof that multi-dictionary works. Solid V1, needs the fixes above before merge.

Abso1ut3Zer0:
Looks good to merge — the codegen engine, dictionary model, and encoder builders are solid. Merging to unblock further progress.

A few items I'll address in a follow-up PR:

- **`convert.rs` scalar parsers** → replace with the SWAR-based domain type parsers from `types.rs` (`parse_fix_int`, `FixDecimal::parse`, etc.) which are 3-5x faster
- **`parse_tag` used as value parser** → swap to the proper value parsers; `parse_tag` is a tag accumulator with different overflow semantics
- **Typed field accessors** → generated accessors will return domain types (`FixDecimal`, `FixTimestamp`, `FixDate`, `FixTime`) via `Result<T, FixValueError>` instead of raw bytes
- **Additional codegen + perf work** — will iterate on the generated code paths in the background while other work progresses

PR #431, opened 2026-06-03
feat(fix-engine): sans-IO FIX session state machine

ChronoCoders:
Reworks the session state machine per the architecture direction from code review.

**What changed**

`Session<'a>` → `SessionState`: pure state machine with typed `on_<admin_msg>` handlers. Each handler takes pre-decoded fields (no raw bytes) and returns `Out` — a stack-allocated struct carrying at most two outbound `AdminMsg`s and one `Event`. Never allocates.

Removed from the session layer:
- Wire encoding (`encode_pending` / `encode_admin` / `FieldWriter`)
- Raw-byte scanning (`scan` / `FieldReader` / tag constants)
- Internal ring buffers (`Ring<Pending>` / `Ring<Event>`)
- `SessionConfig` CompID and `BeginString` fields (framework concerns)
- `mod timestamp` (belongs with the encoder)

`AdminMsg` carries a pre-allocated `MsgSeqNum` so the framework can encode without calling back. `SequenceReset` uses the gap-fill start seq with `PossDupFlag=Y`; all others consume `next_outbound`.

**What's still here**

All session protocol logic: logon/logout state machine, heartbeat and TestRequest timers, inbound gap detection, ResendRequest / SequenceReset / GapFill handling, seq-too-low disconnect, PossDupFlag replay.

**Tests**

24 integration tests covering the same scenarios as before: initiator/acceptor handshake, logon reset flag, app message events, TestRequest echo, heartbeat idle timer, double-heartbeat guard, inbound silence probe → disconnect, gap detection → resend, GapFill, SequenceReset-Reset, ResendRequest gap-fill, seq-too-low, PossDupFlag ignore, CompID mismatch, initiated/counterparty logout, logon/logout timeouts, reject event, seq persistence across reconnect, timeout deadline tracking, disconnected ignore.

**Next**

The framework layer (`Session<D: FixDictionary>`) — uses `D::Header<'buf>::decode()` to dispatch to these typed handlers — is a follow-on PR after this merges.


Abso1ut3Zer0:
Good foundation here — the sans-IO state machine approach is the right call, and the test coverage is solid.

After looking at this alongside the codegen work (#426), I'm going to do some work on the codegen layer first that will inform the final architecture of the session engine. Specifically:

- **Header-first decoding** — decode header fields (8, 9, 35, 34, 49, 56, 43, 52) once, then dispatch to per-message decoders that continue from where the header left off. This eliminates the internal `scan()` function that rescans all tags.
- **Dictionary ZST** — codegen generates a zero-sized type per dictionary (e.g. `Fix44Dict`) that the session framework parameterizes on. Compile-time dispatch, no runtime branching on FIX version.
- **Typed admin message structs** — Logon, Heartbeat, etc. go through the same codegen path as app messages, so the session state machine receives typed data instead of raw `FieldSpan` slicing.

The session state machine itself will become a pure `SessionState` struct with typed `on_<admin_msg>` handlers taking the minimum required data — no ring buffers, no encoding, no decoding at that level. The framework layer above it handles codec dispatch, encoding, and journal interaction.

Separately, I've filed #432 for the mmap'd journal primitive (two-segment rotation, sequence-indexed, zero-copy replay for ResendRequests). That's independent of the codegen work and can be built in parallel — it builds on the same mmap territory as the nexus-shm work.

Holding off on merging this until the codegen pieces land so the architecture can find its footing, but the state machine logic and test scenarios here will carry forward.

Abso1ut3Zer0:
I've been thinking about how `SessionState` fits into the broader FIX engine architecture — specifically how it composes with the codec/codegen layer and the journal work from #454. Here's the direction I'm considering.

## Architecture

The design follows the same pattern as nexus-web's WebSocket. In the WebSocket layer, `FrameReader` handles continuation frame assembly internally — the user never sees continuation frames, only complete `Message::Text`/`Message::Binary`. The FIX session applies the same principle: admin messages (Heartbeat, Logon, Logout, TestRequest, ResendRequest, SequenceReset, Reject) are handled internally by the session state machine. The user only sees app messages.

The split also follows the WebSocket's `into_parts()` pattern from nexus-async-web, where `(WsReader, WsWriter, conn)` are three independent values with no borrow conflicts. Reader and writer take `&mut conn` when they need IO.

```
┌──────────────────────────────────────────────┐
│               User Code                       │
│  match guard.header().msg_type() { ... }      │
│  Only app messages surface here.              │
└──────────────┬───────────────────────────────┘
               │
┌──────────────┴───────────────────────────────┐
│    FixConn<S, D>                              │
│    Owns transport S + Session<D>              │
│    Session wraps SessionState, routes:        │
│      admin → SessionState (invisible)         │
│      app   → header returned to user          │
└───────┬─────────────────────────┬────────────┘
        │ &mut conn               │ &mut conn
┌───────┴─────────┐       ┌──────┴──────────┐
│   FixReader      │       │   FixWriter      │
│   Arc<Inbound    │       │   Arc<Outbound   │
│     Journal>     │       │     Journal>     │
│                  │       │   + WriteBuf     │
│   Scans for FIX  │       │   (framing)     │
│   msg boundaries │       │                  │
└──────────────────┘       └─────────────────┘
```

**Inbound path:** IO → inbound journal (mmap'd) → reader scans for complete messages → session routes admin internally → app messages surface to user

**Outbound path:** Encode into WriteBuf (framing) → append to outbound journal (archival) → flush to conn's transport (IO)

Reader is backed by the inbound journal — no private `ReadBuf`. The journal IS the buffer. Network bytes land directly in the journal's mmap'd segments, and the reader scans that memory for complete FIX messages. The user gets a `MessageGuard` that borrows directly from the journal — zero copy from network to user code.

Writer owns a `WriteBuf` for message framing (body length, checksum, header field ordering) and flushes completed frames to both the outbound journal (archival) and the transport (network). The journal handles persistence — archival is free.

## SessionState — the pure state machine

This is what this PR builds. The core protocol engine — a pure state machine with no IO, no encoding, no journal awareness.

**State transitions:**
```
Disconnected → LogonSent → Active → LogoutSent → Disconnected
```

**Responsibilities:**
- Inbound and outbound sequence number tracking
- Admin message handling — each admin type has a typed handler that updates internal state and returns the minimal data needed for the response (e.g., `on_test_request` returns the `TestReqID` to echo back)
- Heartbeat interval tracking — records when the last message was sent/received so the layers above can drive heartbeat logic
- Comp ID validation — verifies SenderCompID/TargetCompID match expectations
- Gap detection — identifies sequence number gaps for ResendRequest logic
- Logout negotiation — handles the logout handshake (both initiated and received)

**Each handler returns response data directly — not an intermediate descriptor:**

```rust
impl SessionState {
    /// Heartbeat received — update last-received timestamp.
    /// No response needed.
    pub fn on_heartbeat(&mut self, header: &impl FixHeader) { ... }

    /// TestRequest received — returns the TestReqID to echo
    /// back in a Heartbeat response.
    pub fn on_test_request(&mut self, header: &impl FixHeader) -> TestReqId { ... }

    /// Logon received — returns response data.
    /// May include a gap range if sequence numbers are mismatched.
    pub fn on_logon(&mut self, header: &impl FixHeader) -> LogonResponse { ... }

    /// Logout received — returns whether to acknowledge.
    pub fn on_logout(&mut self, header: &impl FixHeader) -> LogoutResponse { ... }

    /// ResendRequest received — returns gap-fill or resend info.
    pub fn on_resend_request(&mut self, header: &impl FixHeader) -> ResendResponse { ... }

    /// SequenceReset received — updates expected seq num.
    pub fn on_sequence_reset(&mut self, header: &impl FixHeader) -> Result<(), SessionError> { ... }

    /// Reject received — updates state.
    pub fn on_reject(&mut self, header: &impl FixHeader) { ... }

    /// App message — sequence bookkeeping only.
    pub fn on_app(&mut self, header: &impl FixHeader) -> Result<(), SessionError> { ... }

    /// Assign next outbound sequence number.
    pub fn next_outbound_seq(&mut self) -> u64 { ... }

    /// Whether the session is Active (can send app messages).
    pub fn is_active(&self) -> bool { ... }
}

pub struct LogonResponse {
    pub heartbeat_interval: u32,
    pub gap: Option<SeqGap>,
}

pub struct SeqGap {
    pub begin: u64,
    pub end: u64,
}
```

**What it does NOT own:**
- IO — no sockets, no reads, no writes
- Encoding — doesn't know how to produce FIX bytes
- Journal — no awareness of persistence
- Timers — doesn't own a clock; tracks timestamps of last sent/received so the layer above can evaluate heartbeat intervals

## Session\<D: FixDictionary\> — dictionary-powered routing

Wraps `SessionState`. Routes inbound messages using the raw `MsgType(35)` tag. Admin message tags are protocol-level constants (`A`, `5`, `0`, `1`, `2`, `4`, `3`) — identical in every FIX version and matched before the dictionary's typed `MsgType` enum is ever consulted. The dictionary's `MsgType` is only for user-side app dispatch.

```rust
pub struct Session<D: FixDictionary> {
    state: SessionState,
    _dict: PhantomData<D>,
}

impl<D: FixDictionary> Session<D> {
    /// Route an inbound message.
    ///
    /// Admin messages are consumed by SessionState internally.
    /// App messages return the dictionary's decoded header.
    pub fn on_message<'buf>(
        &mut self,
        buf: &'buf [u8],
    ) -> Result<Option<D::Header<'buf>>, SessionError> {
        let header = D::Header::decode(buf);

        match header.raw_msg_type().map(|v| v.value()) {
            Some(b"A") => { self.state.on_logon(&header); Ok(None) }
            Some(b"5") => { self.state.on_logout(&header); Ok(None) }
            Some(b"0") => { self.state.on_heartbeat(&header); Ok(None) }
            Some(b"1") => { self.state.on_test_request(&header); Ok(None) }
            Some(b"2") => { self.state.on_resend_request(&header); Ok(None) }
            Some(b"4") => { self.state.on_sequence_reset(&header); Ok(None) }
            Some(b"3") => { self.state.on_reject(&header); Ok(None) }
            Some(_)    => { self.state.on_app(&header)?; Ok(Some(header)) }
            None       => Err(SessionError::MissingMsgType),
        }
    }
}
```

The codegen already generates a `Dict` ZST implementing `FixDictionary` per venue, plus `HeaderDecoder<'buf>` implementing `FixHeader<'buf>` with `raw_msg_type()`, `msg_seq_num()`, `sender_comp_id()`, etc. `D::Header::decode(buf)` is the zero-copy entry point.

## FixConn\<S, D\> — transport + session ownership

Owns the transport and the session, just like `TlsStream<S>` owns the socket + `TlsCodec`. The transport is generic — `S` can be a std `TcpStream`, `MaybeTls<TcpStream>`, a tokio stream, a Unix domain socket, a shared memory pipe, or anything that implements the required IO traits.

```rust
pub struct FixConn<S, D: FixDictionary> {
    stream: S,
    session: Session<D>,
}

impl<S, D: FixDictionary> FixConn<S, D> {
    /// Direct construction — bring any transport.
    pub fn new(stream: S, session: Session<D>) -> Self {
        Self { stream, session }
    }

    pub fn session(&self) -> &Session<D> { &self.session }
}
```

Builder for the common TCP/TLS case:

```rust
let (mut reader, mut writer, mut conn) = FixSessionBuilder::<VenueAlpha>::new()
    .sender("TRADER1")
    .target("EXCHANGE")
    .heartbeat_interval(30)
    .journals(&inbound_journal, &outbound_journal)
    .connect("fix+tls://exchange.example.com:9876")
    .await?;
```

Escape hatch for non-TCP transports:

```rust
// Unix domain socket, shared memory, mock — whatever S: Read + Write
let stream = UnixStream::connect("/var/run/fix-gateway.sock")?;
let conn = FixConn::new(stream, Session::<VenueAlpha>::new(config));
let mut reader = FixReader::new(Arc::clone(&inbound));
let mut writer = FixWriter::new(Arc::clone(&outbound), 8192);
```

## FixReader — journal-backed inbound framing

Holds shared ownership of the inbound journal (`Arc`). Scans journal memory for complete FIX message boundaries. Returns a `MessageGuard` borrowing directly from the journal's mmap'd memory.

When `recv()` encounters an admin message, the routing layer handles it and the writer encodes + flushes the response immediately. The user never sees admin traffic.

```rust
pub struct FixReader {
    journal: Arc<InboundJournal>,
    cursor: u64,
}

impl FixReader {
    pub fn new(journal: Arc<InboundJournal>) -> Self {
        Self { journal, cursor: 0 }
    }

    pub fn recv<'a, S, D>(
        &'a mut self,
        conn: &mut FixConn<S, D>,
        writer: &mut FixWriter,
    ) -> Result<Option<MessageGuard<'a, D>>, Error>
    where
        S: Read + Write,
        D: FixDictionary,
    {
        loop {
            if let Some(msg_range) = self.scan_message()? {
                let buf = self.journal.slice(msg_range);
                let header = D::Header::decode(buf);

                match header.raw_msg_type().map(|v| v.value()) {
                    Some(b"0") => {
                        conn.session.state.on_heartbeat(&header);
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"1") => {
                        let id = conn.session.state.on_test_request(&header);
                        writer.write_heartbeat(conn, Some(&id))?;
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"A") => {
                        let resp = conn.session.state.on_logon(&header);
                        writer.write_logon(conn, resp.heartbeat_interval)?;
                        if let Some(gap) = resp.gap {
                            writer.write_resend_request(conn, gap.begin, gap.end)?;
                        }
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"5") => {
                        let resp = conn.session.state.on_logout(&header);
                        if resp.acknowledge {
                            writer.write_logout(conn, None)?;
                        }
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"2") => {
                        let resp = conn.session.state.on_resend_request(&header);
                        writer.write_sequence_reset(conn, resp.new_seq, true)?;
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"4") => {
                        conn.session.state.on_sequence_reset(&header)?;
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(b"3") => {
                        conn.session.state.on_reject(&header);
                        self.cursor = msg_range.end;
                        continue;
                    }
                    Some(_) => {
                        conn.session.state.on_app(&header)?;
                        return Ok(Some(MessageGuard {
                            header,
                            buf,
                            journal: &self.journal,
                            end_offset: msg_range.end,
                        }));
                    }
                    None => return Err(Error::MissingMsgType),
                }
            }

            let n = self.journal.read_from(&mut conn.stream)?;
            if n == 0 { return Ok(None); }
        }
    }
}
```

### MessageGuard — RAII journal offset tracking

Borrows from journal's mmap'd memory. On drop, advances the journal's read watermark so segments behind that point can be reclaimed. Same pattern as `ReadClaim` in nexus-logbuf.

```rust
pub struct MessageGuard<'a, D: FixDictionary> {
    header: D::Header<'a>,
    buf: &'a [u8],
    journal: &'a InboundJournal,
    end_offset: u64,
}

impl<D: FixDictionary> MessageGuard<'_, D> {
    pub fn header(&self) -> &D::Header<'_> { &self.header }
    pub fn buf(&self) -> &[u8] { self.buf }
}

impl<D: FixDictionary> Drop for MessageGuard<'_, D> {
    fn drop(&mut self) {
        self.journal.advance_read_cursor(self.end_offset);
    }
}
```

## FixWriter — WriteBuf framing + journal archival

Owns a `WriteBuf` for FIX message framing (body length, checksum, header fields) and an `Arc` to the outbound journal. Each write method: frame into WriteBuf → append to outbound journal → flush to conn's transport.

Typed methods for each admin message — no intermediate enum or descriptor. The writer IS the encoder for admin messages.

```rust
pub struct FixWriter {
    journal: Arc<OutboundJournal>,
    write_buf: WriteBuf,
}

impl FixWriter {
    pub fn new(journal: Arc<OutboundJournal>, write_buf_capacity: usize) -> Self {
        Self {
            journal,
            write_buf: WriteBuf::new(write_buf_capacity, 0),
        }
    }

    pub fn write_heartbeat<S, D>(
        &mut self,
        conn: &mut FixConn<S, D>,
        test_req_id: Option<&[u8]>,
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary {
        let seq = conn.session.state.next_outbound_seq();
        encode_heartbeat(&mut self.write_buf, seq, &conn.session, test_req_id);
        self.commit_and_flush(&mut conn.stream)
    }

    pub fn write_logon<S, D>(
        &mut self, conn: &mut FixConn<S, D>, heartbeat_interval: u32,
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary { ... }

    pub fn write_logout<S, D>(
        &mut self, conn: &mut FixConn<S, D>, text: Option<&str>,
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary { ... }

    pub fn write_resend_request<S, D>(
        &mut self, conn: &mut FixConn<S, D>, begin: u64, end: u64,
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary { ... }

    pub fn write_sequence_reset<S, D>(
        &mut self, conn: &mut FixConn<S, D>, new_seq: u64, gap_fill: bool,
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary { ... }

    /// Send a user's app message.
    pub fn send<S, D>(
        &mut self,
        conn: &mut FixConn<S, D>,
        body: &[u8],
    ) -> Result<(), Error>
    where S: Write, D: FixDictionary {
        let seq = conn.session.state.next_outbound_seq();
        encode_app(&mut self.write_buf, seq, &conn.session, body);
        self.commit_and_flush(&mut conn.stream)
    }

    /// Frame is complete in WriteBuf — append to journal, flush to transport.
    fn commit_and_flush<S: Write>(&mut self, stream: &mut S) -> Result<(), Error> {
        self.journal.append(self.write_buf.data());
        stream.write_all(self.write_buf.data())?;
        self.write_buf.clear();
        Ok(())
    }
}
```

## User Code

### Standard connection

```rust
let inbound = Arc::new(InboundJournal::open("session_01_in")?);
let outbound = Arc::new(OutboundJournal::open("session_01_out")?);

// Builder handles transport setup + logon handshake.
// Transport type is inferred — MaybeTls<TcpStream> here.
let (mut reader, mut writer, mut conn) = FixSessionBuilder::<VenueAlpha>::new()
    .sender("TRADER1")
    .target("EXCHANGE")
    .heartbeat_interval(30)
    .inbound_journal(Arc::clone(&inbound))
    .outbound_journal(Arc::clone(&outbound))
    .connect("fix+tls://exchange.example.com:9876")
    .await?;

// Admin messages (Heartbeat, Logon, TestRequest, etc.) are handled
// inside recv() — responses framed into WriteBuf, appended to
// outbound journal, and flushed to transport automatically.
// Only app messages surface.
while let Some(guard) = reader.recv(&mut conn, &mut writer)? {
    match guard.header().msg_type() {
        Some(MsgType::ExecutionReport) => {
            let report = ExecutionReportDecoder::decode(guard.buf(), guard.header());
            on_fill(&report);
        }
        Some(MsgType::NewOrderSingle) => {
            let order = NewOrderSingleDecoder::decode(guard.buf(), guard.header());
            let response = build_ack(&order);
            writer.send(&mut conn, &response)?;
        }
        _ => {}
    }
    // guard drops → journal advances read cursor
}
```

### Escape hatch — custom transport

Any `S: Read + Write` works. Conn, reader, and writer constructed independently:

```rust
// Unix domain socket
let stream = UnixStream::connect("/var/run/fix-gateway.sock")?;
let mut conn = FixConn::new(stream, Session::<InternalVenue>::new(config));
let mut reader = FixReader::new(Arc::clone(&inbound));
let mut writer = FixWriter::new(Arc::clone(&outbound), 8192);

while let Some(guard) = reader.recv(&mut conn, &mut writer)? {
    // same dispatch — transport is abstracted away
}

// Shared memory, mock stream, etc. — same pattern
let mut conn = FixConn::new(ShmStream::open("/fix-01")?, Session::<V>::new(cfg));
let mut conn = FixConn::new(MockStream::from_bytes(data), Session::<V>::new(cfg));
```

### Replay

```rust
let archive = Arc::new(InboundJournal::open_readonly("session_01_in")?);
let mut reader = FixReader::new(archive);
let mut writer = FixWriter::new(Arc::new(OutboundJournal::sink()), 4096);
let mut conn = FixConn::new(io::empty(), Session::<VenueAlpha>::new(config));

while let Some(guard) = reader.recv(&mut conn, &mut writer)? {
    // Deterministic replay — identical dispatch code
}
```

## How This Relates to This PR

`SessionState` is the foundation everything else layers around. The key design boundaries:

1. **SessionState stays pure** — no IO, no encoding, no journal. A state machine that takes decoded header info, updates state, and returns the minimal response data needed (a `TestReqId`, a `LogonResponse`, a gap range). The layers above use that data to write directly — no intermediate enums or descriptor queues.

2. **Handlers return response data, not descriptors** — `on_test_request` returns the `TestReqID` to echo back. `on_logon` returns the heartbeat interval and optional gap. The caller (inside `recv()`) takes that data and calls the writer's typed methods immediately. No `PendingAdmin` queue, no `AdminKind` enum to match on.

3. **Sequence numbers** — SessionState owns the counters. The writer calls `state.next_outbound_seq()` when framing each outbound message (both admin and app).

4. **No opinion on transport or persistence** — SessionState works the same whether bytes came from a TCP socket, a Unix pipe, shared memory, or a replayed journal. The layers above handle all of that.

The pieces that don't exist yet (journals, FixReader/FixWriter, FixConn, Session\<D\>) will layer on top of what's built here. Getting SessionState's contract right — what it owns, what it returns, what it explicitly does NOT know about — is what matters most at this stage.

ChronoCoders:
The architecture outline makes sense end to end. SessionState as a pure state machine returning typed response data (LogonResponse, TestReqId, gap range) rather than AdminMsg descriptors is cleaner — the FixReader/FixWriter layer owns encoding and acts immediately on what SessionState returns.

Aligned on the Session<D> routing by raw MsgType bytes, FixConn<S,D> owning transport + session, and MessageGuard for zero-copy RAII cursor advancement.

Ready to proceed on the framework layer once this and #454 are on main. Happy to keep PRs scoped small if that helps throughput.

ChronoCoders:
Makes sense to merge as-is and let the handler return types evolve with the framework layer. Will track FixDictionary encoding context and FixReader/FixWriter as follow-on issues.

ChronoCoders:
Happy to leave it as-is. The typed per-handler returns only make sense once FixReader/FixWriter exist to consume them — changing the API now without the callers would be speculative.

Abso1ut3Zer0:
Merging this so you can continue building on the architecture direction from the earlier comment without being blocked. A few cleanup items to fold into your next iteration:

1. `ring.rs` and `timestamp.rs` are dead code — no `mod` declarations pull them in. Can remove or wire up as needed.
2. `nexus-fix-codec` is listed as a dependency in Cargo.toml but nothing imports from it yet. Can drop until it's actually used.
3. `AdminMsg` uses `u32` for seq numbers — the codec/codegen uses `u64`. Will sort itself out when the handlers move to returning typed response data per the architecture comment.
4. The `([u8; 64], u8)` tuple for TestReqID echo works but is a bit opaque — same thing, gets replaced with typed response structs.

None of these are blockers — just things to clean up in the next pass.

Issue #432, opened 2026-06-03
fix-engine: mmap'd two-segment journal for session replay


Abso1ut3Zer0:
## Summary

Fixed-capacity, mmap-backed journal for FIX session message replay. Stores raw outbound wire bytes indexed by sequence number. Used by the session framework to service ResendRequests — read the stored bytes and retransmit as-is (zero-copy replay).

## Architecture

Two layers:

**nexus-shm: SegmentedLog** — general-purpose three-segment mmap'd append log with conductor thread. Not FIX-specific. Reusable for audit trails, event sourcing, persistent archival.

**nexus-fix-engine: Journal** — thin wrapper adding sequence number → offset indexing.

## Three-segment rotation with conductor

```
[current]   — active appends (hot path, lock-free)
[prev]      — read-only, available for lookups
[standby]   — cleaned and ready to become current
```

Rotation when current fills:
1. prev → released to conductor for cleaning
2. current → becomes prev (read-only)
3. standby → becomes current (start appending)
4. conductor thread → cleans old prev (reset index, zero data, or remap), becomes new standby

Hot path never blocks on cleanup. Conductor just needs to finish before the next rotation. Only synchronization is the rotation swap (atomic index update).

### SegmentedLog (nexus-shm)

```rust
let log = SegmentedLog::open(path, segment_size)?;

// hot path: append bytes, get offset
let offset = log.append(&wire_bytes);

// read by offset (checks current + prev segments)
let msg = log.read(offset);
```

### Journal (nexus-fix-engine)

```rust
let journal = Journal::new(log);

// hot path: append with sequence number
journal.append(seq_num, &wire_bytes);

// resend path: lookup by sequence number
if let Some(msg) = journal.read(seq_num) {
    // retransmit raw bytes
}

// range replay for ResendRequest
for msg in journal.range(begin_seq, end_seq) {
    // retransmit each
}

// session checks this to decide: replay or SequenceReset-Reset?
if !journal.contains(seq_num) {
    // message has been rotated out
}
```

## Implementation notes

- Each segment is a fixed-size mmap'd region (pre-allocated)
- Conductor thread handles dirty→clean transition (reset write cursor, zero data area, signal standby ready)
- No explicit fsync — mmap handles persistence via OS page writeback. This is session replay, not crash recovery.
- Writes are sequential appends (memcpy into mapped region)
- Reads are positional lookups across current + prev segments
- Segment rotation must not invalidate in-progress range iteration

## Constraints

[...]

PR #454, opened 2026-06-05
feat(shm): three-segment mmap'd append log

ChronoCoders:
Implements `SegmentedLog` in `nexus-shm` — the bounded, lock-free mmap'd log primitive for #432.

**Design**

Three fixed mmap'd segments cycle through three roles:
- **current** — active segment, append target
- **prev** — previous segment, still readable
- **standby** — being cleaned by the conductor thread

On rotation, current becomes prev, prev is handed off to the conductor (zeroes the first commit_len to clear the segment), and standby (cleaned) becomes current. The hot-path append never blocks on cleaning.

**Frame format**

`commit_len(4) | kind(2) | reserved(2) | payload(N)` aligned to 8 bytes.

`commit_len` stores `body_len + 1` so that `0` is an unambiguous "not committed" sentinel — enabling zero-length payload round-trips. Committed frames are published with a Release store; readers use Acquire load.

**`LogOffset`**

Opaque u64 handle: bits `[33:32]` = slot index, `[31:0]` = local byte offset. Valid until the slot is rotated out (two rotations after the write).

**Drop order**

`ConductorHandle` is the first field in `SegmentedLog` so it drops first — joining the conductor thread before the `Slot` mmaps are unmapped.

**Tests** (6): roundtrip, multiple records per segment, rotation readability, slot eviction (`None` after two rotations), oversized payload rejection, empty payload round-trip.

### DIFF
--- a/nexus-shm/src/lib.rs
+++ b/nexus-shm/src/lib.rs
@@ -10,6 +10,7 @@ mod journal;
 mod lock;
 mod pod;
 mod region;
+mod seglog;
 mod segment;
 
 pub use error::ShmError;
@@ -20,4 +21,5 @@ pub use journal::{
 pub use lock::Liveness;
 pub use pod::Pod;
 pub use region::MapOptions;
+pub use seglog::{LogOffset, SegmentedLog, SegmentedLogError};
 pub use segment::{Segment, Status};
--- a/nexus-shm/src/seglog.rs
+++ b/nexus-shm/src/seglog.rs
@@ -9,7 +9,6 @@ use crate::segment::Segment;
 
 const FRAME_HDR: usize = 8;
 const ALIGN: usize = 8;
-const TYPE_DATA: u16 = 0;
 
 const fn align_up(n: usize) -> usize {
     (n + ALIGN - 1) & !(ALIGN - 1)
@@ -19,12 +18,18 @@ const fn footprint(body: usize) -> usize {
     FRAME_HDR + align_up(body)
 }
 
-unsafe fn commit_len_at<'a>(ptr: *mut u8) -> &'a AtomicU32 {
-    unsafe { AtomicU32::from_ptr(ptr.cast()) }
+// Returns a `*mut AtomicU32` pointing at the commit_len field of a frame header.
+// Callers must ensure `ptr` is 4-byte aligned and points into a live mmap'd segment.
+fn commit_len_ptr(ptr: *mut u8) -> *mut AtomicU32 {
+    ptr.cast()
 }
 
 /// Opaque position handle returned by [`SegmentedLog::append`], passed to
 /// [`SegmentedLog::read`]. Valid until the slot it references is rotated out.
+///
+/// Encoding: `[63:34]` = generation, `[33:32]` = slot index, `[31:0]` = local offset.
+/// The default value (`u64::MAX`) encodes slot index 3 (no valid slot), so
+/// [`SegmentedLog::read`] always returns `None` for a default offset.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct LogOffset(u64);
 
@@ -35,16 +40,20 @@ impl Default for LogOffset {
 }
 
 impl LogOffset {
-    fn new(slot: u8, local_off: usize) -> Self {
-        Self((slot as u64) << 32 | local_off as u64)
+    fn new(slot: u8, local_off: usize, epoch: u32) -> Self {
+        Self((epoch as u64) << 34 | (slot as u64) << 32 | local_off as u64)
     }
 
     fn slot(self) -> usize {
-        (self.0 >> 32) as usize
+        ((self.0 >> 32) & 0x3) as usize
     }
 
     fn local_off(self) -> usize {
         (self.0 & 0xFFFF_FFFF) as usize
+    }
+
+    fn epoch(self) -> u32 {
+        (self.0 >> 34) as u32
     }
 }
 
@@ -94,10 +103,11 @@ unsafe impl Send for CleanRequest {}
 
 fn conductor_main(rx: std::sync::mpsc::Receiver<CleanRequest>, ready: &Arc<AtomicBool>) {
     for req in rx {
-        // Zero commit_len at offset 0 — makes the segment appear empty to any
-        // reader that checks before the next writer publishes a record there.
-        unsafe { commit_len_at(req.data).store(0, Ordering::Release) };
-        let _ = req.segment_size; // may expand cleaning here later
+        // SAFETY: `req.data` points to the start of a live mmap'd segment sent by
+        // `rotate()`. The segment remains mapped until the owning `Slot` drops,
+        // which happens only after `ConductorHandle` drops and this thread joins.
+        unsafe { (*commit_len_ptr(req.data)).store(0, Ordering::Release) };
+        let _ = req.segment_size;
         ready.store(true, Ordering::Release);
     }
 }
@@ -141,6 +151,8 @@ pub struct SegmentedLog {
     prev: usize,
     standby: usize,
     cursor: usize,
+    epoch: u32,
+    slot_gen: [u32; 3],
 }
 
 fn slot_path(base: &Path, i: u8) -> PathBuf {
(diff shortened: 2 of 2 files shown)


Abso1ut3Zer0 (commented):
Will review

Abso1ut3Zer0 (commented):
## Review — seglog.rs

Good foundation — three-slot rotation, drop ordering, and commit_len sentinel are well thought out. A few items to address, plus a design question.

### 1. Stale LogOffset returns wrong data after full slot cycle

`read()` checks `slot != self.current && slot != self.prev` — but slot is just an index (0/1/2). After three rotations, slot 0 is `current` again with new data. A stale `LogOffset` from the first cycle decodes as `slot() == 0`, passes the check, and silently returns wrong data instead of `None`.

Bits [63:34] in the u64 handle are unused — a generation counter there eliminates this class of bug at zero cost.

### 2. Only slot 0 initialized — slots 1 and 2 may contain stale mmap data

```rust
unsafe { commit_len_at(s0.data).store(0, Ordering::Relaxed) };
```

Slot 1 starts as `prev` (readable). If the segment file existed from a prior run, stale commit_len values could cause `read()` to return garbage. All three slots should have their first `commit_len` zeroed at open.

### 3. No SAFETY comments on unsafe blocks

Per CONTRIBUTING.md: "Unsafe code requires a `// SAFETY:` comment explaining why it's sound." There are ~8 unsafe blocks in this file with none. `commit_len_at` especially needs one — it creates an `&AtomicU32` from a raw pointer with an unbounded lifetime, relying on alignment from `align_up` and the mmap staying alive.

### 4. `commit_len_at` has unbounded lifetime

```rust
unsafe fn commit_len_at<'a>(ptr: *mut u8) -> &'a AtomicU32
```

The caller picks any lifetime — the reference could outlive the mmap. All current call sites are transient so it's safe in practice, but the signature is overly permissive. Either tighten the lifetime or inline the unsafe at each call site.

### 5. Conductor resilience

The conductor can't panic in its current form (just an atomic store on valid mmap memory), and the drop order is correct. Two things to think about for the next iteration:

- If we add fallible work to the conductor (archival, compression), we'll need error propagation back to the writer — the current `let _ =` on the channel send would silently swallow failures.
- On shutdown, the channel disconnects and the conductor exits its loop. If we need flush-on-exit semantics (e.g. archive the final active segment before dropping), the conductor needs a shutdown signal distinct from channel close.

Neither blocks this PR, but worth noting in a comment near `rotate()`.

### 6. `kind` field written but never read

`TYPE_DATA` is written into every frame header but `read()` never checks it. If this is scaffolding for future frame types, add a comment noting the intent — otherwise remove until needed.

### Design question: archival and log viewing

This is a solid bounded circular buffer over three mmap'd files. For the FIX journal use case, I think we'll want the conductor to archive segments before cleaning — write the full segment to a timestamped file (optionally compressed), then zero. That would make `StandbyNotReady` more likely, which validates the three-segment design (headroom for archive I/O).

For viewing FIX logs: the segment files are binary frames, so we'd need a reader tool that walks frames by commit_len and decodes the FIX payload — something like Aeron's log reader tooling. Could be a shared frame iterator that both a CLI viewer and the conductor's archive path use. Worth discussing how we want this to work before building more on top.

### Nits

- `LogOffset::default()` uses `u64::MAX` as a sentinel — add a comment explaining why (encodes an impossible slot index so `read()` always returns `None`).

ChronoCoders (commented):
Addressed in a3b8444:

1. Generation counter added to LogOffset bits [63:34] — stale offsets after a full slot cycle now return None
2. All three slots zeroed at open
3. SAFETY comments on all unsafe blocks
4. commit_len_at<'a> replaced with commit_len_ptr returning *mut AtomicU32
5. Conductor resilience and flush-on-exit noted in a comment near rotate()
6. TYPE_DATA and kind/reserved writes removed

Archival and the frame iterator for a log viewer — agree those belong in a Discussion before we build more on top. Will open one once this merges.

Abso1ut3Zer0:
Connecting this to the FIX engine architecture discussion from #431 (see [comment](https://github.com/Abso1ut3Zer0/nexus/pull/431#issuecomment-4636681020)) — `SegmentedLog` is the natural backing store for the `FixReader` and `FixWriter` in that design. The core primitive here (three-segment rotation, background conductor cleaning, generation-based offset validity) is solid. This comment is about where it could evolve to support the FIX engine integration.

## Journal as the buffer

In the FIX engine design, the journal replaces `ReadBuf`/`WriteBuf` entirely:

```
Inbound:  socket → journal (mmap'd segment) → FixReader scans for message boundaries → zero-copy &[u8] to user
Outbound: FixWriter frames into WriteBuf → appends to journal → flushes to socket
```

The inbound journal IS the read buffer. Network bytes land directly in the mmap'd segment, and the reader scans that memory. The user gets a `MessageGuard` that borrows from the journal — on drop, it advances a read cursor so the journal knows what can be reclaimed. This is the same pattern as `ReadClaim` in nexus-logbuf.

For this to work, `SegmentedLog` needs a way to expose a slice of its current segment for reading — something like:

```rust
/// View into the active segment's committed region for scanning.
pub fn readable(&self) -> &[u8] { ... }

/// Or: read from a cursor position forward
pub fn read_from(&self, cursor: u64) -> &[u8] { ... }
```

The current `read(offset) -> Option<&[u8]>` API returns individual records by offset. The FIX reader would need sequential scanning — "give me the next N bytes from where I left off" — since FIX message boundaries are discovered by scanning (looking for `8=FIX` through checksum), not by knowing record offsets upfront.

## Conductor sharing

Currently each `SegmentedLog` spawns its own conductor thread. For a FIX engine with multiple sessions, that's one conductor per journal — potentially many threads just for cleaning. A shared conductor pattern would let all journals in a venue share one background thread:

```rust
// Possible evolution — not changing what's here, just sketching the layer above
let conductor = Conductor::start()?;

// Each journal registers with the shared conductor
let inbound_a  = conductor.open("session_a_in",  segment_size)?;
let outbound_a = conductor.open("session_a_out", segment_size)?;
let inbound_b  = conductor.open("session_b_in",  segment_size)?;
let outbound_b = conductor.open("session_b_out", segment_size)?;

// One thread cleans segments for all four journals.
// Each journal sends CleanRequests to the shared conductor's MPSC channel.
```

This would be a new type layered above `SegmentedLog` — the core rotation/cleaning logic here stays the same. The `ConductorHandle` would become injectable rather than always spawning a new thread internally.

## Multiplexing (Artio pattern)

Artio (the FIX engine on Aeron) multiplexes all sessions onto two shared archive streams — one inbound, one outbound. Each message carries a session ID in the header, and a `ReplayIndex` maps `(sessionId, seqNum) → position` for O(1) replay lookups.

We could support both modes:

**Per-session journals** (simpler, current `SegmentedLog` works as-is):
```
session_a_in.seg0, session_a_in.seg1, session_a_in.seg2
session_b_in.seg0, session_b_in.seg1, session_b_in.seg2
```

**Multiplexed journal** (Artio-style, all sessions share one journal):
```
venue_inbound.seg0, venue_inbound.seg1, venue_inbound.seg2
```

For the multiplexed case, the frame header would need a session tag:

```
Current:      [commit_len: u32] [reserved: u32] [payload]
Multiplexed:  [commit_len: u32] [session_id: u32] [payload]
```

Or we always include the session ID — it's 4 bytes in the existing reserved space, and it means any journal can be shared or dedicated without changing the format. Worth considering whether to bake this into the frame header now.

The multiplexed case also needs lock-free concurrent append (`&self` instead of `&mut self`) since multiple sessions write to the same journal. This is the nexus-logbuf MPSC pattern — CAS on the cursor to claim space, then write payload, then commit.

## Shared ownership

For the FIX engine integration, `FixReader` and `FixWriter` need to hold references to their journals. If journals are shared across sessions (multiplexed) or across threads, the handles need `Arc`-based shared ownership:

```rust
let journal = Arc::new(InboundJournal::open(...)?);

// Multiple readers can hold references
let reader_a = FixReader::new(Arc::clone(&journal));
let reader_b = FixReader::new(Arc::clone(&journal));
```

This doesn't need to change `SegmentedLog` itself — the `Arc` wrapping can happen at the layer above. But it's worth keeping in mind that the append/read API needs to be compatible with shared ownership (e.g., `&self` methods with internal synchronization for the multiplexed case).

## Summary

The core `SegmentedLog` here is the right primitive. The evolution for FIX engine integration would layer above it:

| Concern | Current (this PR) | Evolution |
|---------|-------------------|-----------|
| Conductor | Per-log thread | Shared across journals |
| Ownership | Single owner (`&mut self`) | `Arc<Journal>` handles |
| Access pattern | Random read by offset | Sequential scan + cursor |
| Multiplexing | N/A | Optional session ID in frame header |
| Concurrent writes | Single writer | Lock-free MPSC (for multiplexed) |

None of this needs to happen in this PR — just planting the seeds for discussion. The three-segment rotation, generation-based validity, and conductor cleaning pattern are the foundation that everything else builds on.

Also tracked in discussion #457 for broader context.

ChronoCoders:
Agreed on the evolution path. read_from(cursor) -> &[u8] for sequential scanning makes sense — FIX message boundaries are discovered by scan, not by known offsets. Will add that once this merges.

Session ID in the reserved 4 bytes is worth baking in now given it costs nothing and avoids a format break later. Can do that in a follow-on before the FIX engine integration lands.

Shared conductor and Arc ownership are clean layering above SegmentedLog — the core rotation and cleaning logic here stays untouched.

ChronoCoders:
Will open issues for shared conductor + Arc ownership and FixReader/FixWriter after this merges.

Issue #460, opened 2026-06-06
feat(fix-engine): FixDictionary encoding context


ChronoCoders:
Depends on #431 (merged).

The FixDictionary trait currently provides decode-side information (Header<'buf>, MsgType, is_admin, BEGIN_STRING). The framework layer needs compile-time encoding context to frame outbound admin messages without runtime branching on FIX version — sender/target CompID, BeginString, and any version-specific field encoding.

Extend FixDictionary with the minimum required encoding surface. Exact additions to be determined when FixWriter is being built (#459).

Abso1ut3Zer0:
**Reframing this issue.** The original description assumed the engine would encode admin messages internally, requiring encoding context on the trait. After reviewing how real FIX engines work (QuickFIX's `toAdmin`/`fromAdmin` hooks, venue-specific Logon auth fields, etc.), the cleaner design is: **the engine owns the protocol state machine, the user owns the wire format.** The dictionary needs admin message *decoder* types, not encoding context.

### Why

Admin messages aren't uniform across venues. Logon varies the most — some require `EncryptMethod`, others need username/password (tags 553/554), API keys, or `ApplicationVersionID` (FIX 5.0+). Internalizing encoding forces the engine to know about all of these, or requires callback hooks to customize. Surfacing admin messages to the user with typed decoders is simpler and gives full control.

### `Message` enum

The session layer returns a typed `Message` instead of `(Out, Option<Header>)`. Each admin message is its own variant carrying the dictionary's decoder. The user matches, reads whatever fields they need, and encodes the response themselves using the generated encoders.

```rust
pub enum Message<'buf, D: FixDictionary> {
    /// Inbound Logon — we're the acceptor, send our Logon back.
    LogonRequest {
        msg: D::Logon<'buf>,
    },
    /// Logon reply to our outbound Logon — session is active.
    LogonAcknowledged {
        msg: D::Logon<'buf>,
    },
    /// Inbound Logout — counterparty initiated, send ack.
    LogoutRequest {
        msg: D::Logout<'buf>,
    },
    /// Logout ack for our outbound Logout — session is done.
    LogoutAcknowledged {
        msg: D::Logout<'buf>,
    },
    /// Heartbeat (35=0). No action required.
    Heartbeat {
        msg: D::Heartbeat<'buf>,
    },
    /// TestRequest (35=1). Read `test_req_id()`, echo in a Heartbeat.
    TestRequest {
        msg: D::TestRequest<'buf>,
[...]

Issue #459, opened 2026-06-06
feat(fix-engine): FixReader / FixWriter over journal memory


ChronoCoders:
Depends on #454 (merged).

FixReader — journal-backed inbound framing. Network bytes land directly in the mmap'd segment. Scans for FIX message boundaries (8=FIX through checksum), returns a MessageGuard borrowing zero-copy from the journal. On drop, MessageGuard advances the read cursor so segments behind that point can be reclaimed. Admin messages (Heartbeat, TestRequest, Logon, Logout, ResendRequest, SequenceReset, Reject) are handled internally — only app messages surface to the caller.

FixWriter — frames admin and app messages into a WriteBuf, appends to the outbound journal, flushes to transport. Typed write methods per admin message — no intermediate enum.

Abso1ut3Zer0:
Closing this in favor of #464.

The original design had FixReader scanning the journal for FIX message boundaries — network bytes landing directly in the mmap'd segment. Working through multiplexed sessions (#462, #463) surfaced a problem: raw TCP bytes from multiple sessions can't be interleaved in a shared journal. Framing has to happen *before* the journal, not after.

The revised architecture (see discussion #457):
- **FixFramer** (#464) handles TCP → complete FIX messages on the hot path, with incremental header decode as bytes arrive
- **Journal** is an archival side-channel — `append(session_id, complete_msg)` after framing, not the message read path
- **FixWriter** concept folds into the session layer — the user calls generated encoders directly (see #460)

FixReader as described here no longer fits the architecture. The framing half becomes FixFramer (#464), the journal-reading half is just `SegmentedLog::read_next()` on a logger thread.

PR #461, opened 2026-06-06
feat(fix-engine): Session<D: FixDictionary> routing layer

### LINKED ISSUE #458: feat(shm): shared conductor + Arc ownership for SegmentedLog
Follow-on to #454 (merged).

### Problem

Current `SegmentedLog` spawns one conductor thread per log. A venue with N sessions spawns N background threads just for segment cleanup. The conductor should be a shared resource that manages all journals in a venue — one thread, many journals.

### Design

**Conductor as standalone entity:**

```rust
// One conductor for the venue — single background thread
let conductor = Conductor::spawn(config);

// Create journals that register with the conductor
let inbound_journal = conductor.create_journal(journal_config)?;
let outbound_journal = conductor.create_journal(journal_config)?;

// Shared ownership for FixReader / FixWriter
let reader_journal: Arc<Journal> = inbound_journal.clone();
let writer_journal: Arc<Journal> = outbound_journal.clone();

// Session disconnects — journal removed, segments reclaimed
conductor.remove_journal(journal_id);
```

**Conductor responsibilities:**
- Single background thread processing `CleanRequest`s from all journals via MPSC channel
- Segment rotation and reclamation across all managed journals
- Journal lifecycle — create new journals, remove completed ones
- Graceful shutdown — drain pending cleans on drop

**Journal ownership:**
- `Arc<Journal>` handles for shared ownership across FixReader/FixWriter and across threads
- Current `&mut self` append stays for the single-writer (per-session) case
- Journal holds a sender handle to its conductor for clean requests

### What changes from #454

| Current (#454) | Target |
|---|---|
| `SegmentedLog::open()` spawns conductor internally | `Conductor::spawn()` is explicit, journals register with it |
| One conductor thread per log | One conductor thread per venue |
| `ConductorHandle` is private, per-log | `Conductor` is public, injectable, multi-journal |
| Owned `SegmentedLog` | `Arc<Journal>` for shared ownership |

ChronoCoders:
Adds `Session<D: FixDictionary>` — a thin codec-aware wrapper over `SessionState`.

- Decodes inbound buffer via `D::Header::decode`
- Validates SenderCompID / TargetCompID against `SessionConfig`; disconnects on mismatch
- Routes by raw MsgType bytes to typed `SessionState` handlers (Logon, Logout, Heartbeat, TestRequest, ResendRequest, SequenceReset, Reject)
- Returns `(Out, Some(header))` for app messages so callers can decode the body
- Determines acceptor vs initiator from `state != LogonSent` (no separate flag)

9 integration tests using a minimal `MockDict` / `MockHeader`.

Closes #458 prep. Related: #459 (FixReader/FixWriter), #460 (encoding context).

### DIFF
--- a/nexus-fix-engine/src/framework.rs
+++ b/nexus-fix-engine/src/framework.rs
@@ -1,7 +1,9 @@
 use core::marker::PhantomData;
 use std::time::Instant;
 
-use nexus_fix_codec::{FixDictionary, FixHeader, find_tag, parse_fix_bool, parse_fix_seqnum, parse_fix_uint};
+use nexus_fix_codec::{
+    FixDictionary, FixHeader, find_tag, parse_fix_bool, parse_fix_seqnum, parse_fix_uint,
+};
 
 use crate::{Out, SessionState, State};
 
@@ -20,7 +22,10 @@ impl CompId {
         }
         let mut bytes = [0u8; COMP_ID_CAP];
         bytes[..s.len()].copy_from_slice(s);
-        Some(Self { bytes, len: s.len() as u8 })
+        Some(Self {
+            bytes,
+            len: s.len() as u8,
+        })
     }
 
     pub fn as_bytes(&self) -> &[u8] {
@@ -80,7 +85,11 @@ pub struct Session<D: FixDictionary> {
 
 impl<D: FixDictionary> Session<D> {
     pub fn new(state: SessionState, config: SessionConfig) -> Self {
-        Self { state, config, _dict: PhantomData }
+        Self {
+            state,
+            config,
+            _dict: PhantomData,
+        }
     }
 
     pub fn state(&self) -> &SessionState {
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0:
Looks good — merging so you can keep building toward #459.

**Notes for the next iteration:**

1. **`Out` should go away when the writer arrives.** Right now it's a small queue of pending admin messages, which is what we decided against in the architecture comment. When `Session<D>::recv()` gets `&mut FixWriter` in #459, the pattern should be: `SessionState` returns response data (test_req_id, heartbeat interval, etc.) → `Session<D>` immediately calls `writer.write_heartbeat(test_req_id)` → no intermediate struct to drain. `recv()` returns just `Option<D::Header<'buf>>` for app messages + the event.

2. **u64→u32 seq truncation** — `header.msg_seq_num()` returns `u64`, cast to `u32` here. Will sort itself out when `SessionState` moves to `u64` for sequence numbers.

3. **`find_tag(buf, 0, tag)` rescans from 0** for each admin field (e.g., Logon scans twice for tags 108 and 141). Fine for cold-path admin messages, just noting the header decoder from codegen does single-pass.

4. **`framework.rs` naming** — minor, but something more descriptive like `routing.rs` might fit better since this is the dispatch/routing layer.

None of these are blockers for this PR.

Abso1ut3Zer0 (approved):

PR #463, opened 2026-06-06
improvements to seglog + session multiplexing + sequential scanning

Abso1ut3Zer0:
## Summary

- **`Frame<'buf>`**: zero-copy view of a committed record with `session_id()`, `offset()`, `payload()` — enables session multiplexing in the journal
- **Session-tagged writes**: `append(session_id, payload)` stores session_id in the frame header (bytes 4..8, previously reserved)
- **Sequential scan API**: `write_pos()`, `read_start()`, `read_next()` for monotonically increasing global offset scanning across segment rotations
- **Slot init order fix**: `(current=0, prev=2, standby=1)` so rotation visits slots 0→1→2→0, enabling clean `segment_number % 3` mapping
- **Epoch widened to u64**: prevents `write_pos()` wrapping (slot_gen stays u32 for LogOffset validation)
- **Alignment debug_assert in `read_next`**: documents the frame-alignment invariant

## Design notes

Frame header layout is now `[commit_len: u32][session_id: u32][payload...]`. SegmentedLog remains single-threaded SPSC — session_id is a tag for demuxing on the read side, not a concurrency mechanism. A single owner thread writes all sessions; a separate logger/archiver thread tails via `read_next()`.

See discussion #457 for the revised FIX engine architecture where the journal is an archival side-channel (not on the message read path), and #464 for the FixFramer design that replaces the framing half of the original FixReader concept.

## Test plan

- [x] Session ID roundtrip (write + read preserves session_id)
- [x] Session ID survives rotation
- [x] Sequential scan returns session_id per frame
- [x] Frame offset matches global position
- [x] Single segment scan, cross-rotation scan, resume after append
- [x] Evicted segment returns None
- [x] write_pos monotonicity (single segment + across rotation)
- [x] Slot ordering is sequential (0→1→2)
- [x] Edge cases: empty log, empty payloads, variable-size records
- [x] Cursor matches write_pos after full drain
- [x] read_start advances after rotation
- [x] All 40 nexus-shm tests pass, clippy clean, fmt clean

### DIFF
--- a/nexus-shm/src/lib.rs
+++ b/nexus-shm/src/lib.rs
@@ -21,5 +21,5 @@ pub use journal::{
 pub use lock::Liveness;
 pub use pod::Pod;
 pub use region::MapOptions;
-pub use seglog::{LogOffset, SegmentedLog, SegmentedLogError};
+pub use seglog::{Frame, LogOffset, SegmentedLog, SegmentedLogError};
 pub use segment::{Segment, Status};
--- a/nexus-shm/src/seglog.rs
+++ b/nexus-shm/src/seglog.rs
@@ -34,10 +34,11 @@ fn session_id_ptr(ptr: *mut u8) -> *mut u32 {
 ///
 /// Provides access to the session tag, global offset, and payload bytes
 /// without copying from the underlying mmap'd segment.
+#[repr(C)]
 pub struct Frame<'buf> {
-    session_id: u32,
-    offset: u64,
     payload: &'buf [u8],
+    offset: u64,
+    session_id: u32,
 }
 
 impl<'buf> Frame<'buf> {
--- a/nexus-shm/src/seglog.rs
+++ b/nexus-shm/src/seglog.rs
@@ -142,6 +142,32 @@ unsafe impl Send for Slot {}
 /// active becomes readable, clean standby becomes active. The hot-path append
 /// never blocks on cleaning; the conductor must finish before the *next*
 /// rotation (size segments with enough headroom for the expected message rate).
+///
+/// # Global offset addressing
+///
+/// Sequential reads use a monotonically increasing `u64` position that spans
+/// all segments. The physical slot is derived via modular arithmetic:
+///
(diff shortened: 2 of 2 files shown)


Abso1ut3Zer0:
## Future design notes for ShmJournal expansion

A few things surfaced while working on this that are worth tracking:

### SegmentedLog stays SPSC

SegmentedLog is single-threaded `&mut self` — no atomics on the write path, no CAS. Session multiplexing via `session_id` is purely a tagging concern. If we ever need MPMC (multiple sessions writing from different threads + multiple readers), that's a different struct (`MultiplexedJournal`) with CAS-based space claiming and subscriber position tracking. Don't mix these concerns into SegmentedLog.

### Concurrency story across the stack

- **Segment** enforces single-writer via OFD lock — solid for single-process ownership. `attach()` allows multiple readers implicitly. Cross-process reads work via `MAP_SHARED` + atomic `commit_len` (Acquire/Release).
- **SegmentedLog** is `!Sync` (raw pointers) which is the right constraint. If we need to move it between threads (create on main, move to IO thread), add explicit `unsafe impl Send`.
- **Cross-process Reader** — the Acquire on `commit_len` is sound on x86. ARM would need stronger fences. Worth noting but not blocking.
- **Segment discovery race** — Reader polling for new segment files is racy. A cross-process notification mechanism (eventfd, atomic segment count in shared control region) would be needed for reliable tailing.

### Journal is archival, not on the message path

Per discussion #457: the hot path is `FixFramer → Session<D>`. The journal append is a fire-and-forget memcpy into mmap'd memory. A logger thread tails the journal on the cold path. This means SegmentedLog's SPSC model is the right fit — single writer (session thread), single reader (logger thread).

### Backpressure (for future MPMC)

If we ever build the MPMC variant: each reader publishes its position atomically, rotation checks `min(all reader positions)` before evicting. Not a CAS on every read/write — just a check at rotation time. This is the Aeron pattern. Not needed for the current SPSC design.

PR #465, opened 2026-06-06
feat(fix-engine): Message<D> enum, FixAdminMsg trait, admin codegen

### LINKED ISSUE #460: feat(fix-engine): FixDictionary encoding context

ChronoCoders:
Closes #460.

Replaces `(Out, Option<Header>)` with a typed `Message<'buf, D>` enum. Callers own encoding; the engine owns protocol state.

**nexus-fix-codec**
- `FixAdminMsg<'buf>` trait: `fn decode(buf: &'buf [u8]) -> Self`
- `FixDictionary` extended with 7 GATs: `Logon`, `Logout`, `Heartbeat`, `TestRequest`, `ResendRequest`, `SequenceReset`, `Reject` — each bound on `FixAdminMsg`

**nexus-fix-codegen**
- New `admin.rs` emitter — always generates 7 admin decoder structs per dictionary
- Protocol-required fields hardcoded; venue-specific fields from XML merged in
- `emit_mod` wires in `pub mod admin` and sets all 7 associated types on `Dict`

**nexus-fix-engine**
- `Message<'buf, D>` enum: 10 variants (`LogonRequest`, `LogonAcknowledged`, `LogoutRequest`, `LogoutAcknowledged`, `Heartbeat`, `TestRequest`, `ResendRequest`, `SequenceReset`, `Reject`, `Application`, `Disconnected`)
- `Session::on_message` returns `Result<Message<'buf, D>, SessionError>`
- `Out`/`AdminMsg` remain on `SessionState` for direct state-machine users

9 framework tests updated to match the new API.

### DIFF
--- a/nexus-fix-codec/src/dict.rs
+++ b/nexus-fix-codec/src/dict.rs
@@ -10,7 +10,7 @@ use nexus_ascii::AsciiTextStr;
 /// typed accessor methods to read fields.
 pub trait FixAdminMsg<'buf>: Sized {
     /// Construct the decoder from a raw FIX message buffer.
-    fn decode(buf: &'buf [u8]) -> Self;
+    fn decode(buf: &'buf [u8]) -> Result<Self, crate::DecodeError>;
 }
 
 /// Dictionary-level knowledge for a specific FIX version.
--- a/nexus-fix-codec/src/dict.rs
+++ b/nexus-fix-codec/src/dict.rs
@@ -2,11 +2,22 @@ use crate::field::FieldView;
 use crate::types::FixTimestamp;
 use nexus_ascii::AsciiTextStr;
 
+/// Zero-copy decoder for a session-level admin message.
+///
+/// Implemented by every generated admin message type in `admin::*`.
+/// The session framework calls `decode` to construct the decoder and hands it
+/// to the caller via [`Message`](crate::Message); the caller then uses the
+/// typed accessor methods to read fields.
+pub trait FixAdminMsg<'buf>: Sized {
+    /// Construct the decoder from a raw FIX message buffer.
+    fn decode(buf: &'buf [u8]) -> Self;
+}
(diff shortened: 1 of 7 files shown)


Abso1ut3Zer0 (commented):
Looks good overall — the `Message<D>` enum is the right shape and the codegen structure is clean.

One thing to address: the new admin decoders in `admin.rs` don't follow the same pattern as the existing message decoders. The regular codegen gives every message (admin and app) a `header()` method via `HeaderDecoder` — so you get full header + body access on any decoded message. The new admin decoders are simpler structs that just hold `buf` and expose a few hardcoded protocol fields via `find_tag`. No header delegation, no venue-specific fields from the XML.

Since the `Message` enum carries `D::Logon<'buf>`, `D::SequenceReset<'buf>`, etc. as the caller's only handle on the message, those decoders should give the caller the same access they'd get from any other decoded message — including header fields like seq num, CompIDs, sending time. The admin decoders should follow the same emit pattern as `messages.rs`.

Abso1ut3Zer0:
The admin message path in `on_message` parses fields twice — once via `find_tag`/`parse_fix_uint` to feed `SessionState` handlers, then again via `D::Logon::decode(buf)` etc. to build the typed decoder. Not a correctness issue and admin messages are cold path, but worth a cleanup pass where the `SessionState` handlers take the typed decoder directly instead of raw field values. Can tackle separately.

PR #488, opened 2026-06-09
feat: archive evicted segments before zeroing

### LINKED ISSUE #479: feat: segment archival in conductor cleanup thread
## Summary

## Context

## Requirements

## Open questions

ChronoCoders:
Closes #479

Adds opt-in segment archival to the conductor cleanup thread. Before zeroing an evicted segment, the conductor walks its committed frames and writes them as a dense blob to `{session_dir}/archive/seg_{epoch}.dat`.

**API change:** `ConductorBuilder::archive(bool)` — off by default.

**Archive format:** packed committed frames — `[commit_len:u32][session_id:u32][payload...]` — no inter-frame padding. Readable with the same `commit_len`-walk as recovery.

**Error handling:** archival I/O errors are silently skipped; the zero + ready-store always runs.

Punted from #479: compression, retention policy.

### DIFF
--- a/nexus-platform/src/lib.rs
+++ b/nexus-platform/src/lib.rs
@@ -15,4 +15,4 @@ pub mod mapping;
 pub use file_lock::FileLock;
 pub use lease::{Liveness, ProcessLease};
 pub use mapped_file::MappedFile;
-pub use mapping::{Advice, MapError, MappedFileOptions, Mapping, Protection, Sharing};
+pub use mapping::{Advice, MapError, MapOptions, Mapping, Protection, Sharing};
--- a/nexus-platform/src/lib.rs
+++ b/nexus-platform/src/lib.rs
@@ -15,4 +15,4 @@ pub mod mapping;
 pub use file_lock::FileLock;
 pub use lease::{Liveness, ProcessLease};
 pub use mapped_file::MappedFile;
-pub use mapping::{Advice, MapError, MapOptions, Mapping, Protection, Sharing};
+pub use mapping::{Advice, MapError, MappedFileOptions, Mapping, Protection, Sharing};
--- a/nexus-platform/src/mapped_file.rs
+++ b/nexus-platform/src/mapped_file.rs
@@ -3,7 +3,7 @@ use std::num::NonZeroUsize;
 use std::os::fd::OwnedFd;
 use std::path::Path;
 
-use crate::mapping::{MapError, MapOptions, Mapping, Protection, Sharing};
+use crate::mapping::{MapError, MappedFileOptions, Mapping, Protection, Sharing};
 
 /// RAII file-backed memory mapping. Unmaps on drop.
 ///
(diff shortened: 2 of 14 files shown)


Abso1ut3Zer0:
Thanks for picking this up. I've been thinking about the archival model and I think there's a cleaner approach than copying frames to a separate file. Let me lay out the design I'd like to see.

## Core idea: the segment file IS the archive

Instead of walking frames and writing them to a new file, we stop reusing segment files. When a segment is evicted:

1. `fsync` the segment (durability boundary)
2. Rename `seg{i}.dat` → `archive/seg_{epoch}.dat` (atomic on same FS, mmap follows inode)
3. Drop the old `Segment` (munmap — file persists on disk as archive)
4. Create a fresh `seg{i}.dat` for the replacement

No copy step, no partial write risk, no `BufWriter`, same frame format on disk.

## Swap mechanism

The conductor needs to hand the new segment back to the `SegmentedLog`. A three-state machine on a shared struct replaces the current `AtomicBool` ready flag:

```rust
#[repr(u8)]
enum SwapState {
    Clean   = 0,  // new segment available, SegmentedLog can rotate
    Dirty   = 1,  // evicted segment waiting to be sent to conductor
    Pending = 2,  // sent to conductor, awaiting replacement
}

pub(crate) struct SegmentSwap {
    state: AtomicU8,
    inner: UnsafeCell<MaybeUninit<SwapPayload>>,
}

struct SwapPayload {
    segment: Segment,
    data: *mut u8,
}

// SAFETY: access serialized by `state` — never concurrent.
// Clean: SegmentedLog reads (after Acquire on state)
// Dirty: SegmentedLog writes, then conductor reads (after send through channel)
// Pending: inner is uninit, nobody touches it
unsafe impl Sync for SegmentSwap {}
unsafe impl Send for SegmentSwap {}
```

The `state` field provides Acquire/Release barriers. No mutex needed.

## Rotation flow

```rust
fn rotate(&mut self) -> Result<(), LogError> {
    // Can only rotate if conductor has provided a clean segment.
    if self.swap.state() != SwapState::Clean {
        return Err(LogError::StandbyNotReady);
    }

    // Take the replacement segment from the swap.
    // SAFETY: state == Clean (Acquire) guarantees payload is initialized
    // and conductor is done writing.
    let (new_seg, new_data) = unsafe { self.swap.take() };

    // Evict old_prev: swap its segment for the fresh one.
    let old_prev = self.prev;
    let evicted = std::mem::replace(&mut self.slots[old_prev].segment, new_seg);
    self.slots[old_prev].data = new_data;

    // Park the evicted segment in the swap for the conductor.
    // SAFETY: we just took from swap (now uninit), writing before
    // publishing Dirty (Release).
    unsafe { self.swap.store_dirty(evicted) };

    // Rotate indices.
    self.prev = self.current;
    self.current = self.standby;
    self.standby = old_prev;
    self.cursor = 0;
    self.epoch += 1;
    self.slot_gen[self.current] = (self.epoch as u32) & EPOCH_MASK;
    self.manifest.set_epoch(self.epoch);

    // Best-effort: try to flush the evicted segment to conductor now.
    self.try_flush_dirty();

    Ok(())
}

/// Attempt to send a dirty (evicted) segment to the conductor for archival.
///
/// Called after rotation and on each `append()`. If the channel is full,
/// the segment stays in the swap as Dirty and we retry next time — no
/// undo, no error propagation.
fn try_flush_dirty(&mut self) {
    if self.swap.state() != SwapState::Dirty {
        return;
    }

    // SAFETY: state == Dirty (Acquire) guarantees the evicted segment
    // is in the swap and we own it.
    let (evicted, _) = unsafe { self.swap.take() };

    let request = CleanRequest {
        segment: Some(evicted),
        segment_size: self.segment_size,
        epoch: self.epoch.saturating_sub(1),
        swap: Arc::clone(&self.swap),
        seg_path: seg_path(&self.session_dir, self.standby as u8),
        hints: self.hints,
        archive_dir: self.archive_dir.clone(),
    };

    // Transition Dirty → Pending before send. If send fails,
    // we get the request back and restore Dirty.
    self.swap.set_state(SwapState::Pending);

    match self.tx.try_send(request) {
        Ok(()) => {
            // Sent. Conductor will archive, create replacement,
            // and publish Clean when done.
        }
        Err(std::sync::mpsc::TrySendError::Full(returned)) => {
            // Channel full (other sessions). Put segment back, retry later.
            let segment = returned.segment.expect("just sent it");
            unsafe { self.swap.store_dirty(segment) };
        }
        Err(std::sync::mpsc::TrySendError::Disconnected(_)) => {
            panic!("conductor cleanup thread has exited unexpectedly");
        }
    }
}

pub fn append(&mut self, payload: &[u8]) -> Result<LogOffset, LogError> {
    // Opportunistically flush any dirty segment before appending.
    self.try_flush_dirty();

    let body = payload.len();
    let foot = footprint(body);
    if foot > self.segment_size {
        return Err(LogError::RecordTooLarge {
            max: self.segment_size.saturating_sub(FRAME_HDR),
        });
    }
    if self.cursor + foot > self.segment_size {
        self.rotate()?;
    }
    // ... frame write logic unchanged ...
}
```

## Conductor work loop

The conductor becomes a proper work loop with a retry buffer, not a simple `for req in rx`:

```rust
fn conductor_main(rx: Receiver<CleanRequest>) {
    let mut pending: Vec<PendingCreate> = Vec::new();

    loop {
        // 1. Retry pending creates from prior ENOSPC failures.
        pending.retain_mut(|p| {
            match Segment::create(&p.seg_path, p.segment_size, p.hints) {
                Ok(seg) => {
                    // zero first commit_len, publish to swap
                    unsafe { p.swap.publish_clean(seg) };
                    false
                }
                Err(_) => true, // still failing, retry next round
            }
        });

        // 2. Non-blocking drain of new requests.
        let mut drained = false;
        while let Ok(req) = rx.try_recv() {
            drained = true;
            process_request(req, &mut pending);
        }

        // 3. Block only when idle. Sleep if retrying.
        if pending.is_empty() {
            match rx.recv() {
                Ok(req) => process_request(req, &mut pending),
                Err(_) => break,
            }
        } else if !drained {
            std::thread::sleep(Duration::from_millis(250));
        }
    }
}

fn process_request(req: CleanRequest, pending: &mut Vec<PendingCreate>) {
    if let Some(segment) = req.segment {
        let _ = segment.sync();
        // rename to archive dir...
        drop(segment);
    }
    match Segment::create(&req.seg_path, req.segment_size, req.hints) {
        Ok(seg) => unsafe { req.swap.publish_clean(seg) },
        Err(_) => pending.push(PendingCreate { /* ... */ }),
    }
}
```

The only realistic runtime failure is ENOSPC. The conductor retries automatically — no user-facing error handling needed. The `SegmentedLog` just sees `StandbyNotReady` backpressure until disk frees up.

## What changes from this PR

- **No frame-copying archive** — the segment file persists as-is
- **`CleanRequest` carries `Segment` ownership** (not a raw pointer) — `Option<Segment>` to support retry-only requests
- **`SegmentSwap` replaces `Arc<AtomicBool>`** — three-state machine for the handoff
- **Conductor loop gains retry buffer** — drain pending, drain channel, block only when idle
- **`Slot` holds `Segment` directly** (not `_segment`) — ownership moves in and out during rotation

## What stays the same

- `ConductorBuilder::archive(bool)` API
- Archive directory structure (`{session_dir}/archive/`)
- The frame format on disk (identical — the segment files ARE the archive)
- Recovery logic (same commit_len walk)

This is a bigger change than the current PR, so it might make sense to open a design issue for tracking and implement incrementally. Happy to discuss any of the above.

PR #489, opened 2026-06-09
refactor: MappedFileOptions builder, MapHints for nexus-shm

### LINKED ISSUE #486: refactor MappedFile and SharedMemory to use builder options pattern
## Context

Current `MappedFile` constructors (`create`, `open`, `open_readonly`, `from_file`) take a plain `MapOptions` struct and scatter protection/sharing/offset configuration across multiple methods with different signatures. `from_file` alone has 6 parameters. This doesn't follow the `std::fs::OpenOptions` convention and won't scale cleanly to `SharedMemory` (#482).

## Problem

- `MapOptions` is too thin — just two bools (`pretouch`, `huge_pages`)
- Protection and sharing mode are implicit in which constructor you call rather than explicit in configuration
- No builder pattern — full-control usage (`from_file`) has too many positional parameters
- `SharedMemory` will need the same config surface plus type-specific options (sealing, named vs anonymous vs fd-attach), compounding the problem

## Proposal

Follow the `std::fs::OpenOptions` pattern. Each type gets its own options builder returned by `::options()`:

```rust
// MappedFile — full control via builder
let mf = MappedFile::options()
    .read_write()
    .shared()
    .pretouch(true)
    .offset(4096)
    .create(path, len)?;

// MappedFile — convenience shortcuts stay (use default options internally)
let mf = MappedFile::create(path, len)?;
let mf = MappedFile::open(path)?;

// SharedMemory — named
let shm = SharedMemory::options()
    .read_write()
    .shared()
    .create_named(name, len)?;

// SharedMemory — anonymous (Linux only)
let shm = SharedMemory::options()
    .read_write()
    .shared()
    .allow_sealing(true)
    .create_anonymous(len)?;

// SharedMemory — attach to received fd (anonymous or named)
let shm = SharedMemory::options()
    .read_only()
    .shared()
    .attach(fd, len)?;
```

### Type structure

- `MappedFileOptions` — returned by `MappedFile::options()`, terminals: `create`, `open`, `open_readonly`, `from_file`
- `SharedMemoryOptions` — returned by `SharedMemory::options()`, terminals: `create_named`, `create_anonymous`, `attach`
- Both embed a shared internal config struct for the overlapping fields (protection, sharing, pretouch, huge_pages)
- `MapOptions` struct goes away — its fields are absorbed into the builders
- `Protection` and `Sharing` enums stay as public types

## Depends on

- Should be done before #482 (SharedMemory) so the builder pattern is established first

## Scope

- Refactor `MappedFile` constructors to use `MappedFileOptions` builder
- Keep existing convenience constructors as shortcuts
- Update tests to exercise both builder and convenience paths
- Update nexus-shm call sites if needed

Abso1ut3Zer0:
Closes #486

## Summary

- Introduces `MappedFileOptions` builder following `std::fs::OpenOptions` convention (`&mut self` setters, `&self` terminals) for full control over mapped file creation
- Simplifies `MappedFile::create/open/open_readonly` to parameterless convenience methods that delegate to the builder with defaults
- Makes `MapOptions` internal (`pub(crate)`) — no longer part of the public API
- Adds `MapHints` to `nexus-shm` as the crate-owned type for pretouch/huge_pages, replacing the leaked `MapOptions` re-export
- Renames `JournalConfig.map` → `JournalConfig.hints` throughout

## Test plan

- [x] `cargo test --workspace` — all passing
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] New tests: `options_builder_pretouch`, `options_builder_readonly_private`, `options_builder_create_with_offset`
- [x] Existing tests updated to parameterless convenience methods
- [x] Copilot review feedback addressed: `open()` with offset fix, `create()` no-truncate fix, `MapHints` doc fix

### DIFF
--- a/nexus-platform/src/lib.rs
+++ b/nexus-platform/src/lib.rs
@@ -14,5 +14,5 @@ pub mod mapping;
 
 pub use file_lock::FileLock;
 pub use lease::{Liveness, ProcessLease};
-pub use mapped_file::MappedFile;
-pub use mapping::{Advice, MapError, MapOptions, Mapping, Protection, Sharing};
+pub use mapped_file::{MappedFile, MappedFileOptions};
+pub use mapping::{Advice, MapError, Mapping, Protection, Sharing};
--- a/nexus-platform/src/mapped_file.rs
+++ b/nexus-platform/src/mapped_file.rs
@@ -126,7 +126,9 @@ impl MappedFileOptions {
             .offset
             .checked_add(len.get() as u64)
             .ok_or(MapError::OutOfBounds)?;
-        file.set_len(total)?;
+        if total > file.metadata()?.len() {
+            file.set_len(total)?;
+        }
         self.map_file(file, len)
     }
 
@@ -140,7 +142,11 @@ impl MappedFileOptions {
         } else {
             OpenOptions::new().read(true).open(path)?
         };
-        let len = NonZeroUsize::new(file.metadata()?.len() as usize).ok_or(MapError::EmptyFile)?;
+        let file_len = file.metadata()?.len();
+        let remaining = file_len
+            .checked_sub(self.offset)
+            .ok_or(MapError::OutOfBounds)?;
+        let len = NonZeroUsize::new(remaining as usize).ok_or(MapError::EmptyFile)?;
         self.map_file(file, len)
     }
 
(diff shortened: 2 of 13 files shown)


Abso1ut3Zer0:
Duplicate of #490

PR #493, opened 2026-06-09
feat(seglog): SegmentSwap 3-state archival redesign (#479)

### LINKED ISSUE #479: feat: segment archival in conductor cleanup thread

ChronoCoders:
Supersedes #488. Implements the redesign from Michaels review comment.

## Changes

**SegmentSwap** — lock-free 3-state cell (Clean/Dirty/Pending) replacing Arc<AtomicBool>. Standby slot is always None; its segment lives exclusively in the swap, eliminating the double-open OwnerActive conflict.

**rotate()** — installs the swap Clean segment into slots[standby] (becomes new current), evicts slots[prev] to swap as Dirty.

**try_flush_dirty()** — sends CleanRequest with the evicted Segment; restores Dirty if channel is full (back-pressure, no block).

**Conductor** — fsync + atomic rename archival (rename follows inode; existing mmaps unaffected), retry buffer for ENOSPC.

**Segment** — stores path: PathBuf; exposes path() and sync().

77/77 tests pass, clippy -D warnings clean.

Closes #479

### DIFF
--- a/nexus-platform/src/lib.rs
+++ b/nexus-platform/src/lib.rs
@@ -15,4 +15,4 @@ pub mod mapping;
 pub use file_lock::FileLock;
 pub use lease::{Liveness, ProcessLease};
 pub use mapped_file::MappedFile;
-pub use mapping::{Advice, MapError, MappedFileOptions, Mapping, Protection, Sharing};
+pub use mapping::{Advice, MapError, MapOptions, Mapping, Protection, Sharing};
--- a/nexus-platform/src/lib.rs
+++ b/nexus-platform/src/lib.rs
@@ -15,4 +15,4 @@ pub mod mapping;
 pub use file_lock::FileLock;
 pub use lease::{Liveness, ProcessLease};
 pub use mapped_file::MappedFile;
-pub use mapping::{Advice, MapError, MapOptions, Mapping, Protection, Sharing};
+pub use mapping::{Advice, MapError, MappedFileOptions, Mapping, Protection, Sharing};
--- a/nexus-platform/src/mapped_file.rs
+++ b/nexus-platform/src/mapped_file.rs
@@ -3,7 +3,7 @@ use std::num::NonZeroUsize;
 use std::os::fd::OwnedFd;
 use std::path::Path;
 
-use crate::mapping::{MapError, MapOptions, Mapping, Protection, Sharing};
+use crate::mapping::{MapError, MappedFileOptions, Mapping, Protection, Sharing};
 
 /// RAII file-backed memory mapping. Unmaps on drop.
 ///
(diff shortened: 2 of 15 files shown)


Abso1ut3Zer0:
## Review

Two PRs landed on main since this branch was cut that change the foundation you're building on. The good news is they give you cleaner building blocks — the design here is solid and the rebase should make some things simpler.

### What's on main now

**#491 — `SharedMemory` in nexus-platform.** POSIX shared memory primitive (`/dev/shm`) alongside `MappedFile`. Both wrap the same inner `Mapping` type.

**#494 — `Segment` takes `impl Into<Mapping>`.** `Segment` now holds `Mapping` directly instead of `MappedFile`, and `create`/`attach` accept anything that converts to a `Mapping` (via `From<MappedFile>` and `From<SharedMemory>`). Segment is now backing-agnostic. There's also `Segment::total_size(data_len)` for sizing and a `MappingTooSmall` bounds check on both constructors.

The journal and seglog on main already use the new pattern — they create a `MappedFile` explicitly and pass it into `Segment`. See `journal/mod.rs::file_create` and `seglog/mod.rs::file_create` for the pattern.

### What needs to change on rebase

**1. `path: PathBuf` moves off `Segment` into `Slot` and `CleanRequest`.**

Segment is backing-agnostic now — no paths, no names. The path is a seglog concern. The conductor needs the path for two things: archive rename and replacement segment creation. Both can get it from `CleanRequest` (which already carries `seg_path`) rather than from the segment.

In `Slot`, add a `path: PathBuf` field alongside `data`. The conductor's archive rename uses `req.seg_path` (or a stored original path) instead of `segment.path()`.

**2. Conductor creates replacement segments using `MappedFile` + `Segment::create`.**

```rust
let total = Segment::total_size(req.segment_size)?;
let mf = MappedFile::create(&req.seg_path, total)?;
let seg = Segment::create(mf, req.segment_size, hints)?;
```

This is the same pattern the journal and seglog already use on main. The conductor owns the file creation — Segment just takes the mapping.

**3. `MapOptions` is `pub(crate)` in nexus-platform — can't import it from nexus-shm.**

Keep using `MapHints` (your crate's type) and convert at the `MappedFile` creation boundary:
```rust
let mut opts = MappedFile::options();
opts.pretouch(hints.pretouch).huge_pages(hints.huge_pages);
opts.create(&path, total)?
```

**4. Old segment lifecycle — just drop it.**

Dropping the evicted `Segment` does all the cleanup: marks control block `Dead`, calls `munmap`, closes the fd, releases the OFD lock. No additional cleanup needed. The file persists on disk (by design for crash recovery). So the conductor flow is:

1. `segment.mapping().sync()` (fsync dirty pages — `Mapping` exposes `sync()`)
2. If archiving: `rename()` the file (atomic, follows inode — existing mmap unaffected)
3. `drop(segment)` — munmap + close fd + release OFD lock
4. `MappedFile::create` at the slot path — fresh file (if renamed) or reuse (if not)
5. `Segment::create(mf, ...)` — claims OFD lock on new fd, writes control block

No OFD lock conflict because the old fd is closed before the new one opens.

### Design review — the core is good

The SegmentSwap state machine (Clean/Dirty/Pending) is correct. Standby-lives-in-swap eliminating the double-open conflict is the right fix. fsync + atomic rename for archival is sound. `try_flush_dirty` on `append` gives natural back-pressure without blocking.

### Smaller items

- **`set_state(u8)` is too permissive** — raw state setter bypasses the state machine invariants. Replace with `mark_pending()` or inline it at the single call site in `try_flush_dirty`.
- **Error silencing on archival** — `let _ = segment.sync()` and `let _ = std::fs::rename(...)` silently drop I/O failures. For archival, that's worth tracking (counter, flag, something the caller can observe). Silent data loss is hard to debug at 3am.
- **`unsafe impl Send for Segment` is redundant** — `Mapping` already has `unsafe impl Send/Sync`, so Segment auto-derives it through its fields.

ChronoCoders:
Superseded by #496 (rebased on main after #489/#491/#494).

PR #496, opened 2026-06-09
feat(seglog): segment archival with SegmentSwap 3-state machine (#479)

### LINKED ISSUE #479: feat: segment archival in conductor cleanup thread

ChronoCoders:
Closes #479. Rebased on main after #489/#491/#494.

## What this does

Replaces the `ready:AtomicBool` + inline-zeroing conductor with a `SegmentSwap` 3-state machine (Clean/Dirty/Pending) that carries evicted `Segment` ownership across the conductor boundary.

**State machine:**
```
Clean  ──(rotate takes)──▶ [uninit] ──(store_dirty)──▶ Dirty
Dirty  ──(try_flush sends)──▶ Pending
Pending ──(conductor publishes)──▶ Clean
```

**Key design decisions:**
- Standby slot holds no `Segment` — its segment lives exclusively in the swap until conductor publishes a replacement
- `rotate()`: takes Clean replacement from swap → installs in standby slot; evicts prev → swap as Dirty
- `try_flush_dirty()`: sends Dirty segment to conductor (Pending); retries on next append if channel full
- Conductor fsyncs then renames evicted segment to archive dir before creating replacement (atomic rename, inode follows mmap)
- `Drop`: spins on `SWAP_PENDING` (not a flag) so segments are live until conductor is done

**Adapted to #489/#491/#494:**
- `path` moves off `Segment` into `Slot.path` and `CleanRequest.seg_path`
- `Segment::create(impl Into<Mapping>)` via `file_create() -> MappedFile` pattern
- `MapOptions` → `MapHints`

## Review items addressed

- `set_state(u8)` removed; replaced with typed `mark_pending()`
- `unsafe impl Send for Segment` removed (auto-derives via `Mapping`)
- Archival errors degrade gracefully: sync failure skips rename, still creates replacement; TODO comment flags missing observability counter

### DIFF
--- a/nexus-shm/src/seglog/conductor.rs
+++ b/nexus-shm/src/seglog/conductor.rs
@@ -1,10 +1,17 @@
+use std::cell::UnsafeCell;
 use std::io::{Read as _, Seek, SeekFrom, Write as _};
+use std::mem::MaybeUninit;
+use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicU8, Ordering};
 use std::thread::JoinHandle;
+use std::time::Duration;
 
-use nexus_platform::FileLock;
+use nexus_platform::{FileLock, MappedFile};
+
+use crate::MapHints;
+use crate::segment::Segment;
 
 use super::frame::commit_len_ptr;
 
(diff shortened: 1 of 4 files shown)


Abso1ut3Zer0:
The SegmentSwap state machine is the right design — ownership transfer is much cleaner than the old raw-pointer approach. A few things to address:

**Bug: conductor thread hangs on shutdown with pending creates**

`conductor_main` only breaks at `rx.recv() -> Err`, guarded by `pending.is_empty()`. If the channel disconnects while `pending` is non-empty (e.g. persistent ENOSPC), the loop runs forever — `try_recv()` returns `Disconnected`, `pending` isn't empty, so it sleeps and retries instead of breaking. The old `for req in rx` loop broke naturally on disconnect. Need to check for disconnect in the `try_recv` loop or in the retry branch.

**Archive epoch off-by-one**

`epoch: self.epoch.saturating_sub(1)` — after `rotate()` increments to E+1, the evicted segment was prev at epoch E-1. File gets named `seg_E.dat` but contains epoch E-1 data. Should be `saturating_sub(2)` or capture the epoch before rotate increments it.

**Nits**
- Doc comments were stripped from public types (`Conductor`, `ConductorBuilder`, their methods). These need to stay — especially the directory layout docs.
- `pending.retain` with `is_none_or(|seg| { ...; false })` is hard to parse. A `match` with explicit `true`/`false` would be clearer.
- `unsafe impl Send for Slot` safety comment is now stale — `data` can be `null_mut()` for the standby slot.

Abso1ut3Zer0 (approved):

PR #500, opened 2026-06-10
feat(conductor): wakeup-free MPSC + adaptive backoff (#499)

### LINKED ISSUE #499: nexus-journal: replace conductor sync_channel with wakeup-free MPSC + adaptive backoff
## Summary

Replace the `std::sync::mpsc::sync_channel` that carries `CleanRequest`s from
writer sessions to the conductor with a **wakeup-free** MPSC handoff that the
conductor drains via **adaptive backoff** (spin → `yield_now` → capped sleep).
This removes the futex from the writer's rotation path entirely and collapses
the ms-scale `StandbyNotReady` stall toward the cost of the conductor's actual
work (~µs).

## Problem

Today the writer hands off via `SyncSender::try_send` and the conductor parks in
`Receiver::recv()`. std's bounded channel uses an internal mutex + condvar, so
`try_send` must **wake** the parked conductor with a `FUTEX_WAKE`. Two costs:

1. The rotation-triggering `append()` pays a futex-wake syscall (+ the channel
   mutex) on the writer thread.
2. More importantly, the time until the conductor publishes a clean segment is
   then dominated by **wakeup + scheduler-placement latency**, which on a
   contended / non-isolated core spikes to milliseconds — the writer sees
   `StandbyNotReady` and the record's effective insert latency becomes ms.

## Evidence (`examples/rotating_tail.rs`, ext4/nvme, pinned, turbo off)

- After the reuse-in-place change the conductor's per-rotation work is ~zero
  syscalls (zero a header word + publish), yet ms-scale stalls persist under
  small-segment / high-rate sizing — so the stall is **wakeup, not work**.
- Free-core test isolates it as scheduling:

  | cpuset | stall p50 | stall max |
  |--------|-----------|-----------|
  | `-c 0,2` (writer + conductor share 2 cores) | ~0.5 ms | ~3–4 ms |
  | `-c 0-11` (conductor gets a free core) | ~0.13 ms | ~0.3–0.9 ms |

  The multi-ms spike is the parked conductor waiting for the scheduler to run it.
- At the default 4 MiB segments the conductor keeps up and stalls are **zero**;
  the stall only appears under small segments + high append rate.

## Proposal

1. **Drop the std channel.** Replace it with a bounded **wakeup-free MPSC**
   carrying `CleanRequest` (many writer-sessions → one conductor, so MPSC;
   bounded to preserve the existing `try_send`-full → retry-on-next-append
   backpressure). `CleanRequest` owns a `Mapping` (an fd) + an `Arc`, so the
   queue must carry **owned, non-`Pod`** items. `nexus-queue::mpsc` is a strong
   candidate — it's a bounded lock-free MPSC ring buffer with `T: Send`,
   `push -> Result<(), Full<T>>`, `pop -> Option<T>`, and disconnect detection.
   The queue primitive already exists in the workspace.

2. **Conductor drains with adaptive backoff:** `pop` loop; when empty, three-phase
   backoff: fast spin → crossbeam `Backoff::snooze` → exponential `park_timeout`
   (e.g. 1ms → 2ms → 4ms → ... → configurable cap, default ~3s). Reset to min on
   successful drain. The conductor is a **non-pinned, slow-path** thread — the cap
   bounds idle CPU while keeping pickup latency low. `nexus-channel` already
   implements this exact pattern for SPSC; the conductor needs the MPSC equivalent.

3. **Writer enqueue wakes the conductor:** After a successful push, the writer
   conditionally unparks the conductor (same conditional-wake pattern
   nexus-channel uses — check the `parked` flag, only syscall when the conductor is
   actually sleeping). This is critical: with a long backoff cap (seconds), the
   conductor could be deep in a `park_timeout` when the writer pushes a
   `CleanRequest`. Without a wake, the conductor won't process it until the sleep
   expires, and the writer hits `StandbyNotReady` on the next rotation.

   ```
   try_flush_dirty:
       queue.push(request);
       swap.mark_pending();
       if conductor_parked.load(Acquire) {
           conductor_thread.unpark();
       }
   ```

   The unpark is a single atomic store when the conductor is already awake (no
   syscall). The `futex(WAKE)` only fires when the conductor is genuinely sleeping,
   which is exactly when we need it.

4. **Shutdown must be immediate.** Use `park_timeout` instead of `sleep` so `Drop`
   can interrupt. `Drop` sets a closing flag, then calls `unpark` on the conductor
   handle. The conductor wakes immediately, drains remaining requests, and exits.

   This matters in containerized environments (Docker, k8s): the orchestrator sends
   SIGTERM and then SIGKILL after a grace period (default 30s in k8s, 10s in
   Docker). If graceful shutdown hangs because the conductor is mid-sleep at a long
   cap, the process eats a SIGKILL and skips handlers, flush, cleanup — exactly the
   scenario where you lose data or leave stale lock files.

   ```
   Conductor::drop:
       closing.store(true, Release)
       conductor_thread.unpark()
       conductor_thread.join()

   Conductor main loop:
       loop {
           drain_queue();
           if closing.load(Acquire) { break; }
           parked.store(true, Release);
           park_timeout(backoff_duration);
           parked.store(false, Relaxed);
           backoff_duration = min(backoff_duration * 2, max_sleep);
       }
   ```

## While here (optional)

The `msync` gate means evicted-segment dirty pages now ride the kernel's
background writeback. If sustained write-rate ever approaches disk bandwidth, a
non-blocking `Mapping::sync_async()` (`MS_ASYNC`, already in `nexus-platform`) on
the conductor is a zero-block release valve.

## Scope

Internal conductor / handoff change — no public API change. Latency refinement
for small-segment / high-rate use; the default 4 MiB sizing is already
stall-free.

ChronoCoders:
Closes #499.

Replaces `std::sync::mpsc::sync_channel` with `nexus_queue::mpsc::ring_buffer` and rewrites the conductor work loop with adaptive backoff + conditional unpark.

## What changed

**`conductor.rs`**
- Channel: `sync_channel` → `mpsc::ring_buffer` (lock-free, no futex on push)
- `Conductor` gains `closing: Arc<AtomicBool>`, `parked: Arc<AtomicBool>`, `wake_thread: Thread`
- `Drop`: sets `closing = true`, unparks thread, joins — immediate shutdown (no sleep expiry wait)
- `conductor_main`: adaptive backoff — drain → if idle: `parked.store(true)` → re-check (race close) → `park_timeout` with exponential backoff (1ms → ... → 3s cap)

**`mod.rs`**
- `SessionResources` / `RotatingJournal`: `SyncSender` → `Producer`, gain `parked` + `wake_thread`
- `try_flush_dirty`: after successful push, conditionally unparks conductor (`parked.load(Acquire)` — single atomic check, syscall only when conductor is actually sleeping)

### DIFF
--- a/nexus-journal/Cargo.toml
+++ b/nexus-journal/Cargo.toml
@@ -14,6 +14,7 @@ workspace = true
 
 [dependencies]
 nexus-platform = { version = "0.1.0", path = "../nexus-platform" }
+nexus-queue = { version = "1.3.1", path = "../nexus-queue" }
 
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
--- a/nexus-journal/src/rotating/conductor.rs
+++ b/nexus-journal/src/rotating/conductor.rs
@@ -8,8 +8,8 @@ use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::thread::JoinHandle;
 use std::time::Duration;
 
-use nexus_platform::{FileLock, MappedFile, Mapping};
 use nexus_platform::MapHints;
+use nexus_platform::{FileLock, MappedFile, Mapping};
 use nexus_queue::mpsc;
 
 const LOCK_FILE: &str = "conductor.lock";
@@ -275,11 +275,7 @@ struct PendingCreate {
     hints: MapHints,
 }
 
-fn conductor_main(
-    rx: &mpsc::Consumer<CleanRequest>,
-    closing: &AtomicBool,
-    parked: &AtomicBool,
-) {
+fn conductor_main(rx: &mpsc::Consumer<CleanRequest>, closing: &AtomicBool, parked: &AtomicBool) {
     const MAX_SLEEP: Duration = Duration::from_secs(3);
     let mut sleep_dur = Duration::from_millis(1);
     let mut pending: Vec<PendingCreate> = Vec::new();
--- a/nexus-journal/src/rotating/conductor.rs
+++ b/nexus-journal/src/rotating/conductor.rs
@@ -4,13 +4,13 @@ use std::mem::MaybeUninit;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::thread::JoinHandle;
 use std::time::Duration;
 
 use nexus_platform::{FileLock, MappedFile, Mapping};
-
 use nexus_platform::MapHints;
+use nexus_queue::mpsc;
 
 const LOCK_FILE: &str = "conductor.lock";
 const DEFAULT_CLEAN_QUEUE_DEPTH: usize = 4;
@@ -160,13 +160,22 @@ impl ConductorBuilder {
     pub fn open(self) -> Result<Conductor, super::OpenError> {
         std::fs::create_dir_all(&self.dir)?;
 
-        let (tx, rx) = std::sync::mpsc::sync_channel(self.clean_queue_depth);
-        let thread = std::thread::spawn(move || conductor_main(&rx));
+        let (tx, rx) = mpsc::ring_buffer(self.clean_queue_depth);
+        let closing = Arc::new(AtomicBool::new(false));
+        let parked = Arc::new(AtomicBool::new(false));
+
+        let closing_c = Arc::clone(&closing);
+        let parked_c = Arc::clone(&parked);
+        let thread = std::thread::spawn(move || conductor_main(&rx, &closing_c, &parked_c));
+        let wake_thread = thread.thread().clone();
 
         Ok(Conductor {
             dir: self.dir,
-            tx: Some(tx),
+            tx,
             thread: Some(thread),
+            closing,
+            parked,
+            wake_thread,
             archive: self.archive,
         })
     }
@@ -178,8 +187,11 @@ impl ConductorBuilder {
 /// creates fresh replacements. Drop to shut the thread down gracefully.
 pub struct Conductor {
     dir: PathBuf,
-    tx: Option<std::sync::mpsc::SyncSender<CleanRequest>>,
+    tx: mpsc::Producer<CleanRequest>,
     thread: Option<JoinHandle<()>>,
+    closing: Arc<AtomicBool>,
+    parked: Arc<AtomicBool>,
+    wake_thread: std::thread::Thread,
     archive: bool,
 }
 
@@ -225,8 +237,16 @@ impl Conductor {
         &self.dir
     }
 
-    pub(crate) fn sender(&self) -> std::sync::mpsc::SyncSender<CleanRequest> {
-        self.tx.as_ref().expect("conductor shut down").clone()
+    pub(crate) fn sender(&self) -> mpsc::Producer<CleanRequest> {
+        self.tx.clone()
+    }
+
+    pub(crate) fn parked(&self) -> Arc<AtomicBool> {
+        Arc::clone(&self.parked)
+    }
+
+    pub(crate) fn wake_thread(&self) -> std::thread::Thread {
+        self.wake_thread.clone()
     }
 
     pub(crate) fn archive(&self) -> bool {
@@ -236,7 +256,8 @@ impl Conductor {
 
 impl Drop for Conductor {
     fn drop(&mut self) {
-        drop(self.tx.take());
+        self.closing.store(true, Ordering::Release);
+        self.wake_thread.unpark();
         if let Some(t) = self.thread.take() {
             let _ = t.join();
         }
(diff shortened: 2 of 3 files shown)


Abso1ut3Zer0:
Solid direction — lock-free handoff, `parked` flag to gate the syscall, `park_timeout`/`unpark` for prompt shutdown. The structure is what #499 asked for. Three things to address before this merges, one of which is a correctness bug.

### 1. Lost-wakeup race in the park/unpark handshake (blocking)

This is a StoreLoad ordering problem across two independent atomics, and Release/Acquire doesn't cover it.

Consumer (`conductor_main`):
```rust
parked.store(true, Release);   // A
// ... rx.pop() re-check         B  (Acquire load of the queue)
park_timeout(sleep_dur);
```
Producer (`try_flush_dirty`):
```rust
self.tx.push(request);         // C  (Release store into the queue)
if self.parked.load(Acquire)   // D
```

`parked` and the queue's head/tail are separate atomics, and the queue is pure Release/Acquire internally (`pop` → `turn.load(Acquire)`, `push` → `turn.store(Release)`). Release/Acquire is exactly the barrier that permits a store→load reorder, so this interleaving is legal:

1. Producer's `load(parked)` floats ahead of its `push` → reads `false`
2. Consumer's `pop` floats ahead of its `store(parked = true)` → reads empty
3. Consumer stores `parked = true`, parks
4. Producer pushes — nobody unparks it

`park_timeout` bounds it, so it's not a hang — but the backoff caps at 3s and only ramps while idle. Worst case: the system goes quiet, `sleep_dur` ramps to 3s, and the first rotation after the quiet period hits this race. The standby cleanup is then delayed up to 3s and surfaces on the hot path as `StandbyNotReady` on the next rotation — i.e. the unpark fails to fire in exactly the situation it exists for.

**The fix I'd suggest is to delete the `parked` flag entirely rather than add barriers to it.** Step back to what it buys: `try_flush_dirty` is rotation-only (cold path), and the flag exists solely to skip `unpark()` when the conductor is awake. But `std::thread::unpark()` on a running thread is already syscall-free — std's parker does `state.swap(NOTIFIED, Release)` and only issues `futex_wake` if the previous state was `PARKED`. So the flag is optimizing a cost that doesn't exist on the path that matters, and in doing so it re-implements std's own park token as an external atomic and gets the ordering wrong.

Drop the flag and unpark unconditionally after a successful push:

```rust
// producer (try_flush_dirty), after a successful push:
self.tx.push(request)?;
self.wake_thread.unpark();        // unconditional

// consumer (conductor_main):
loop {
    // drain queue;
    if closing { break; }
    if had_work { sleep_dur = ONE_MS; continue; }
    park_timeout(sleep_dur);       // std token absorbs the race
    if closing { break; }
    sleep_dur = (sleep_dur * 2).min(MAX_SLEEP);
}
```

This is correct with no fences and no flag. `push` is sequenced-before `unpark` on the producer; std's park token can't be lost (an `unpark` before `park` leaves a token that `park` consumes and returns on immediately — that's the std `Parker` contract). So `push →(po) unpark →(sync-with) park-returns →(po) re-drain`: by transitivity the consumer's re-drain must observe the pushed item. No lost wakeup, no lost request. Cost is one unconditional swap per rotation, which is negligible on a cold path. Keep `closing` and the adaptive backoff exactly as they are.

(crossbeam's `Parker`/`Unparker` would also be correct and would let you drop the flag the same way, but it pulls `crossbeam-utils` into the crate — which the extraction deliberately kept out — and the only ergonomic thing it adds is a `Clone + Send + Sync` unpark handle, which `std::thread::Thread` already is. Not worth the dependency here.)

### 2. The pending-retry path lost its backoff and now busy-spins

Previously:
```rust
} else if !drained {
    std::thread::sleep(Duration::from_millis(250));
}
```
Now:
```rust
if drained || !pending.is_empty() {
    sleep_dur = Duration::from_millis(1);
    continue;   // no sleep
}
```

When `pending` is non-empty (a `file_create` failed — ENOSPC, EMFILE), this `continue`s straight back into the `pending.retain(...)` retry with zero delay, pegging a core and retrying the failing syscall as fast as the loop turns. `pending` only fills on real I/O failure, so this is the disk-full / fd-exhaustion path specifically — the retry needs its own backoff (reuse `sleep_dur`, or a fixed interval). The comment that documented the old behavior was removed along with it.

### 3. The disconnect assert can't detect what its message claims

```rust
assert!(!self.tx.is_disconnected(), "conductor cleanup thread has exited unexpectedly");
```

`Producer::is_disconnected()` is `Arc::strong_count(&shared) == 1` — true only when this producer is the last handle alive. The `Conductor` holds its own `tx` clone and every sibling `RotatingJournal` holds one, so when the consumer thread actually exits this still returns `false` whenever another producer exists. The old `sync_channel` returned `Disconnected` the moment the in-thread `Receiver` dropped, which reliably detected thread death; this check only fires in the single-handle case. Either the message should be toned down or we need a real liveness signal for the consumer thread.

### Minor

- Related: nothing ties a `RotatingJournal`'s lifetime to its `Conductor` (it owns cloned `tx`/`wake_thread`). With the old code, dropping the conductor early would hang in `join()`; now the thread exits and journals push into a consumerless queue — combined with #3 that's either a hot-path panic or silent forever-`StandbyNotReady`. Pre-existing, but the failure mode changed here, so it's a good moment to document the "conductor outlives journals" contract or enforce it.

#1 is the one that has to change before merge. #2 and #3 are regressions against behavior the old `sync_channel` gave for free. Happy to pair on any of these.

Abso1ut3Zer0 (approved):

PR #501, opened 2026-06-11
fix(codec): FieldReader::next_data_field + codegen error propagation (#429)

### LINKED ISSUE #429: fix-codec: length-prefixed data fields break SOH scanner on embedded 0x01
## Summary

`FieldReader` is a pure SOH-delimited scanner. FIX *data* fields are length-prefixed and may legally contain embedded SOH (`0x01`) and `=` bytes in their value. When they do, the scanner cuts the field at the first embedded SOH and silently mis-parses everything after it.

The checksum is unaffected (it sums every byte regardless) — the corruption is **silent at the structural level**.

## Reproduction

Wire (SOH shown as `\x01`), XmlData pair `212=Len` / `213=<value>` where the value contains a SOH:

```
...212=8\x01213=AB\x01CDEF\x0110=123\x01
        ^len=8        ^embedded SOH inside the 8-byte value
```

Tag 213's value should be the 8 bytes `AB\x01CDEF`. Instead the scanner:

1. Reads `213=AB` (stops at embedded `\x01`), yielding truncated value `AB`
2. Resumes at `CDEF\x01` — no `=`, so `next_field` returns `None` or mis-attributes a bogus tag
3. Rest of message is lost

`validate_checksum` still passes because the PSADBW accumulation sums every byte regardless of field structure.

## Existing test gap

`reader.rs` line 775 (`binary_value`) parses a `96=...` (RawData) field, but only with a value containing **no** embedded SOH — exactly the case that works today. The embedded-SOH case is untested and broken.

## Standard data-tag pairs affected

| Length tag | Data tag | Field |
|----------:|---------:|-------|
| 90 | 91 | SecureDataLen / SecureData |
| 95 | 96 | RawDataLength / RawData |
| 212 | 213 | XmlDataLen / XmlData |
| 348 | 349 | EncodedTextLen / EncodedText |
| 358 | 359 | EncodedIssuerLen / EncodedIssuer |
| 360 | 361 | EncodedSecurityDescLen / EncodedSecurityDesc |
| … | … | rest of the `Encoded*` family |

### Option A — Length-tag aware reader

Teach the reader the Length→Data association (static table or caller-configurable). When `next_field` parses a length tag, capture `N`; when the next tag is the associated data tag, consume exactly `N` bytes without SOH-scanning.

- **Pro:** transparent to callers, Iterator keeps working
- **Con:** per-field check on hot path (must be well-predicted not-taken branch), SIMD mask-cache coherence after literal consume

### Option B — Caller-driven escape hatch (recommended)

Expose `next_data_field(len: usize)` or `read_raw(len)`. The caller — which just parsed the Length field — calls it to consume exactly `N` bytes literally.

- **Pro:** zero overhead on the common no-data-field path, no built-in tag table
- **Con:** pushes correctness onto caller, composes less cleanly with Iterator

### Recommendation

**Option B first.** The dominant workload (crypto order entry / market data) has no data fields — don't regress the common path for a rare feature. Layer Option A on top later if venue traffic warrants it.

## Design constraints

- Raw payload bytes (including embedded SOH/`=`) **must** be in the FIX checksum
- `soh_mask`, `mask_base`, `scan_pos`, `field_start` must be left consistent after a literal consume
- Data fields can appear inside repeating group entries
- Malformed input (`N` > remaining, `N = 0`, non-numeric length) must return a decode error, never read OOB

## Test plan

- [ ] Data field with embedded SOH
- [ ] Data field with embedded `=`
- [ ] Data field as last field before `10=`
- [ ] Multiple data fields in one message
- [ ] Data field inside a repeating group
- [ ] Checksum correctness over message with embedded SOH payload
- [ ] Malformed length: `N` > remaining, `N = 0`, non-numeric length

ChronoCoders:
Closes #429.

**What**

- `FieldReader::next_data_field(len)` — consumes the next field as exactly `len` raw bytes without SOH-scanning; returns `DecodeError::Truncated` for `len == 0`, `len > remaining`, or missing terminating SOH
- Codegen `data_body` now calls `next_data_field` via `?` instead of silently clamping with `.min(buf.len().saturating_sub(vstart))`

**Tests** — 9 new cases in `reader::tests`: embedded SOH, embedded `=`, last field before `10=`, multiple pairs, inside repeating group, checksum coverage, zero len, N > remaining, missing terminator

### DIFF
--- a/nexus-fix-codec/src/error.rs
+++ b/nexus-fix-codec/src/error.rs
@@ -2,6 +2,7 @@ use core::fmt;
 
 /// Error during FIX message decoding.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum DecodeError {
     /// Message is too short to contain required header fields.
     Truncated,
@@ -15,6 +16,8 @@ pub enum DecodeError {
     MissingBodyLength,
     /// Checksum validation failed.
     Checksum(ChecksumError),
+    /// DATA field declared a zero or inconsistent length.
+    InvalidLength,
 }
 
 impl fmt::Display for DecodeError {
@@ -26,6 +29,7 @@ impl fmt::Display for DecodeError {
             Self::MissingBeginString => f.write_str("missing or misplaced BeginString (tag 8)"),
             Self::MissingBodyLength => f.write_str("missing or misplaced BodyLength (tag 9)"),
             Self::Checksum(e) => write!(f, "checksum: {}", e),
+            Self::InvalidLength => f.write_str("DATA field declared invalid length"),
         }
     }
 }
@@ -187,6 +191,10 @@ mod tests {
         assert_eq!(
             DecodeError::Checksum(ce).to_string(),
             "checksum: expected 178, computed 042"
+        );
+        assert_eq!(
+            DecodeError::InvalidLength.to_string(),
+            "DATA field declared invalid length"
         );
     }
 
--- a/nexus-fix-codec/src/reader.rs
+++ b/nexus-fix-codec/src/reader.rs
@@ -68,6 +68,29 @@ impl<'a> FieldReader<'a> {
         }
     }
 
+    #[inline]
+    pub fn next_data_field(&mut self, len: usize) -> Result<RawField, crate::DecodeError> {
+        if len == 0 {
+            return Err(crate::DecodeError::Truncated);
+        }
+        let field_start = self.field_start;
+        let (tag, tag_len) = parse_tag(self.buf.get(field_start..).unwrap_or(b""));
+        if tag_len == 0 {
+            return Err(crate::DecodeError::InvalidTag);
+        }
+        if self.buf.get(field_start + tag_len) != Some(&b'=') {
+            return Err(crate::DecodeError::MissingSeparator);
+        }
+        let value_start = field_start + tag_len + 1;
+        let data_end = value_start + len;
+        if self.buf.get(data_end) != Some(&b'\x01') {
+            return Err(crate::DecodeError::Truncated);
+        }
+        let span = FieldSpan::new(value_start as u32, len as u32);
+        self.resync_after_data(data_end + 1);
+        Ok(RawField { tag, value: span })
+    }
+
     /// Resume scanning at `data_end`, having consumed a length-prefixed DATA
     /// field whose value may contain SOH bytes.
     ///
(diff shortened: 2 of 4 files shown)


Abso1ut3Zer0:
Good fix. The old codegen silently clamped a declared DATA length to the buffer (`.min(buf.len().saturating_sub(vstart))`), so a bad length decoded to a quietly-shortened span — a malformed wire message turning into a plausible-looking order. Reading exactly `len` bytes, requiring the terminating SOH, and propagating via `?` is the correct contract, and the roundtrip flip (`..._does_not_panic` → `..._errors`) captures it. The embedded-SOH / embedded-`=` / checksum-coverage tests cover the real cases.

I checked the arithmetic against the framing model too: this codec is sans-IO — the session framework frames the message via BodyLength and hands `decode` a complete buffer, so `next_data_field` always runs on a fully-delimited message. `len` arrives as `n as usize` from `parse_tag`'s `u32`, so on 64-bit `value_start + len` can't overflow `usize`; out-of-range just yields `None` → error. Memory-safe on the supported target. Two things:

### 1. Public method needs a doc comment (would hold merge on this)

`pub fn next_data_field` ships with no docs, while every sibling (`next_field`, `resync_after_data`, `verify_checksum`) is fully documented. It's `pub` because generated code in other crates calls it, and the contract is non-obvious: `len` is the raw byte count *excluding* the terminating SOH, that SOH is required, and `len == 0` is rejected. Please document it — it's a shipped public API with subtle behavior.

### 2. Give the bad-length case its own error variant (recommended)

Erroring on `len == 0` is correct — FIX forbids empty field values (`SessionRejectReason = 4`, "tag specified without a value") and a conformant peer never sends `95=0`, so the old clamp-to-empty was swallowing a protocol violation. That part's right.

I'd return a distinct variant for it rather than `Truncated`, though — something like `InvalidLength` (or `MalformedData`). It's not a correctness issue here (with the frame already delimited upstream, both a zero length and an overrunning length are just "malformed message, reject"), but a separate value is worth having for diagnostics: a venue-integration engineer reading logs, or a session layer mapping decode failures to the right `SessionRejectReason`, gets a precise signal — "peer declared a zero/inconsistent data length" vs. the generic "truncated." Cheap to add, and it ages well as more malformed-input cases get distinguished. While there, `#[non_exhaustive]` on `DecodeError` would keep future variant additions from being breaking changes.

Minor, folds into the above: a `let remaining = self.buf.len().saturating_sub(value_start); if len > remaining { ... }` pre-check makes the bound explicit instead of leaning on 64-bit `usize` width.

### Not this PR

`next_data_field` returns `tag` but codegen discards it — no check that the field after the length field is the schema's expected DATA tag. Pre-existing positional-trust design, not a regression; noting it for the record only.

---

Net: correct, well-tested fix. Add the doc comment (#1); the dedicated error variant (#2) is a recommended refinement, not a blocker.

PR #502, opened 2026-06-12
feat(fix-engine): MessageStore + SessionStore persistence traits (#408)

### LINKED ISSUE #408: nexus-fix-engine: sans-IO session layer

ChronoCoders:
Closes #408.

Replaces `MessageStore`/`StoreFactory` traits and `store.rs` with `FixJournal` backed by `RotatingJournal` instead of `AppendOnlyJournal`.

**Changes**
- `FixJournal::open(dir, window)` -- opens or recovers via `Conductor` + `RotatingJournal`; window must be a power of two
- `recover()` -- infallible scan from `read_start()` to restore `next_outbound`
- `store(seq, timestamp, msg)` -- encodes `[seq:u32 le][ts:u64 le][msg]`, stores `LogOffset` in ring
- `resend(begin, end)` -> `ResendPlan::Replay(Frame)` on ring hit, `GapFill` on miss/rotation
- `ResendPlan<'a>` enum replaces separate `resend()` + `resend_is_aged_out()` methods
- Correct drop order: `journal` before `_conductor` so `RotatingJournal` can spin-wait during drop


Abso1ut3Zer0:
The instinct to put persistence behind a clean seam is right, and the journal is the correct backing store. But I'd reshape this fairly significantly before it lands — the traits are abstraction we don't need, and the `retrieve` path gives up everything the journal does well. Concretely:

### 1. Drop the `MessageStore` / `SessionStore` traits — one concrete `FixJournal`

Each trait has exactly one implementor and no second caller asking for one, and the journal-backed version is testable directly against a tmpdir. That's a struct with extra ceremony, not an abstraction. Collapse it to a single concrete `FixJournal` that owns the journal writer/reader (plus the resend index below) and exposes inherent methods — `store`, `resend`, `next_seq`, startup recovery. No `type Error` associated type; it just returns `AppendOnlyJournalError`. `store.rs` goes away, `persist.rs` becomes `FixJournal`.

Extracting a trait later, if a real second backend ever appears, is a non-breaking change — so there's no cost to waiting and a real cost (indirection + associated types on every call) to minting it now.

### 2. Fold session sequence state into the journal — delete `SessionStore`/`FileSessionStore`

`FileSessionStore` writes `(next_inbound, next_outbound)` with `std::fs::write` — truncate-then-write, no fsync, no atomic rename. That's the one piece of state that must survive a crash intact (get it wrong and the session resumes at the wrong seqnum → resend storm / forced logout), and this is the least safe way to store it.

But it shouldn't be a separate file at all. The seqnum already rides in each record header, so journaling a message *is* persisting its seqnum — durable for free via the journal's msync. So:
- **hot path:** `next_outbound` is an in-memory counter, bumped per send
- **durable:** the journal append you're already doing
- **startup:** reseed the counter from the last record's seqnum; have the manifest cache `last_seq` so that's O(1) rather than a scan to the tail

No separate file, no second source of truth, no torn-write failure mode. `SessionStore` and `FileSessionStore` both delete.

### 3. `retrieve` must be lazy and zero-copy

This is the main one. `Reader::read_range` already returns a lazy, borrowing range, and the impl throws that away:
```rust
range.map(|rec| Ok(StoredMsg { bytes: rec.payload().to_vec(), .. }))
     .collect::<Vec<_>>()
     .into_iter()
```
Every record gets a heap copy, and a `ResendRequest` with `EndSeqNo=0` ("everything from BeginSeqNo") materializes the entire history into a `Vec` up front — off an mmap built to stream. The `impl Iterator` return type advertises laziness it doesn't deliver, and `StoredMsg { bytes: Vec<u8> }` bakes the allocation into the type.

Yield borrowed bytes (`&[u8]` / the journal's `ReadRecord<'_>`) and keep it lazy. The reason for the `.collect()` was unifying the error vs map `match` arms — solve that with `-> Result<impl Iterator<Item = Result<…>>, Error>`: setup error out front, per-item `Result` only for genuine mid-stream record faults.

### 4. O(1) resend lookup via an offset ring

`read_range` keyed by seqnum has to locate `begin` by scanning. A pre-allocated `Box<[LogOffset]>` indexed by `seq & mask` (size a power of two) gives O(1) seek: `store` records the offset, `resend` jumps straight to it.

One correctness catch — `ring[seq & mask]` only holds that seq's offset while it's within the last `N` written; after `seq + N` is stored the slot is overwritten, so a lookup for an aged-out seq returns a *stale offset for a newer message*. The record header makes this cheap to catch: seek → read → verify `header.seq == requested`. Match ⇒ resend; mismatch ⇒ it's aged out ⇒ the session sends `SequenceReset-GapFill`, which is the correct FIX response anyway. That also turns the ring size into an explicit "how far back will we truly resend" policy knob. Store `LogOffset` (global), not a raw byte offset, so entries survive segment rolls.

### 5. Journal API this needs

The append journal doesn't expose these yet, so they're part of the work:
- `commit` returns the `LogOffset` where the record landed (to populate the ring)
- `read_at(LogOffset) -> ReadRecord<'_>` for the O(1) lookup
- `last_seq()` (manifest-cached) for O(1) startup reseed

### Smaller
- `FixHeader { seq, timestamp: 0 }` discards the send time. FIX PossDup resend requires `OrigSendingTime (122)` = the original sending time — store it, or a correct resend can't be reconstructed.
- `rec.header().seq as u32` silently narrows u64→u32; the rest of the codebase avoids silent truncation.
- No tests on a persistence/crash-recovery layer — this is exactly where you want round-trip store/resend, range boundaries (`begin==end`, empty, aged-out), and save→reseed-after-restart coverage.
- `lib.rs` doc can stay scoped to "the *session state machine* is sans-IO" while the crate also ships the IO adapters (TCP transport, `FixJournal`) — same split as nexus-web, no contradiction.

If a trait belongs anywhere it's the transport adapter — but nexus-net already ships `WireStream`/`ParserSink` for that seam, so parameterize over those rather than minting a new one.

Net: delete the two traits, one concrete `FixJournal`, seqnum state lives with the journal, and `resend` streams borrowed bytes through the offset ring. Happy to talk through any of these.

Abso1ut3Zer0:
This is a strong revision — it hits everything from the last round: traits gone in favor of a concrete `FixJournal`, `resend` returns the lazy borrowing `ReadRange` (no eager `Vec`, no per-message `to_vec`), the offset ring with verify-seqnum-on-read is exactly right, `next_outbound` derives from the journal, and the timestamp is stored. Tests green, clippy clean. The structure is what I want.

One architectural change before it lands — and it simplifies more than it adds: **build `FixJournal` on `RotatingJournal` instead of `AppendOnlyJournal`.**

The reason is the hot path. Journaling an outbound message sits on the send path, and `AppendOnlyJournal` rolls segments inline — an `mmap` syscall on the critical path (the ~ms-scale spike). `RotatingJournal` exists to move exactly that off-path: `append()` is a pointer-swap, and the Conductor does the `mmap` on a background thread. The "never evicts so replay always has full history" rationale also goes away once you accept a **bounded replay window** — which is the standard FIX model: replay within a window, `SequenceReset-GapFill` everything older (Artio's `ReplayIndex` is precisely a bounded circular seq→position index over a durable log).

And it composes more cleanly than the AppendOnly version. `RotatingJournal::read(offset) -> Option<Frame>` already returns `None` once an offset rotates out of the live slots, so:
- the live slots *are* the replay horizon (the window),
- the ring stores `LogOffset` (from `append`), and
- `resend` collapses to one `read()`: `Some(frame)` + matching seq ⇒ replay; `None`/seq-mismatch ⇒ gap-fill.

That makes the `resend` / `resend_is_aged_out` split disappear — same `read()`, O(1), no scan fallback (the inconsistency from before). Recovery comes from the manifest instead of the `last_seq` scan, and turning on the Conductor's archival gives the full audit trail off the replay path.

Concretely:
- `store` → `append()` (returns `LogOffset` for the ring); Conductor becomes part of setup.
- ring holds `LogOffset`; `resend` does `read(ring[seq & mask])`.
- have `resend` return an explicit plan — `Replay(range)` vs `GapFill { from, to }` — so the session can't treat an aged-out range as replayable, and it handles the straddle case (request begins below the window, ends inside it).
- make the window size configurable rather than a hardcoded `8192`.
- the AppendOnly `read_from`/`peek_header`/`last_seq` additions can come back out.

One smaller thing regardless of the above: `next_inbound` is in-memory only (`recover()` only restores outbound) — document who restores it after a restart, typically the session at Logon via `set_next_inbound`.

This is the journal you built for exactly this hot path, and the replay semantics fall out of `read()→Option` for free. Solid work getting the structure right — the retarget is the last piece.

ChronoCoders:
Switched to RotatingJournal (Conductor manages rotation off-path, append() returns LogOffset for the ring). AppendOffset, read_from, peek_header, last_seq removed from nexus-journal. set_next_inbound has a doc comment noting caller restores from Logon's NextExpectedMsgSeqNum.

Abso1ut3Zer0:
This is close — the RotatingJournal retarget is right, `FixJournal` owning the `Conductor` handles the lifetime, and `resend` collapsing to a single `read()→Option` with the verify-seqnum check is exactly what I wanted. Two gaps left before it lands:

1. **`resend` needs to handle the range, not just `begin`.** Right now a `begin`-miss gap-fills the entire `[begin, end]`, so a request that straddles the window edge (begin below the window, end inside) — or where `begin` is a hole but later seqnums are stored — gap-fills messages it could actually replay. Either make it a per-seqnum resolver (drop `end`; `GapFill` is a single seq; the session loops `begin..=end` and coalesces consecutive gaps into one `SequenceReset`), or a range walker that returns the replay/gap-fill split. I'd lean per-seqnum — keeps the journal a dumb store and the FIX coalescing in the session. Either way, add a straddle test (some stored, some not, mixed in one range).

2. **`store` should archive the already-formatted message, not re-frame it.** Right now it allocates a `Vec` and prepends `seq + timestamp` to the body — but by the time you're journaling, the message has already been formatted in the `FrameWriter` to be sent, and that formatted slice already carries `MsgSeqNum` (tag 34) and `SendingTime` (tag 52). So the prepend is re-framing data that's already framed, and the `Vec` is a heap allocation on the send path — which is most of what we moved to RotatingJournal to avoid.

   When the message is formatted you already know the slice that forms it, so just archive that slice directly: `journal.append(&slice)` on IO success, and index `ring[seq & mask] = offset` with the seqnum the send path already assigned. Drop the `timestamp` param and the prepend entirely — the cold paths (`recover`, resend verify) read the seqnum back via `find_tag(34)`, and the time is already tag 52. That removes the hot-path allocation at the *source* (no scratch buffer, no double copy — the only copy is the formed slice into the segment, which is irreducible). Storing the raw wire message is also what lets a resend reconstruct `OrigSendingTime` (122) / `PossDupFlag` correctly, so it's the right source of truth too.

Everything else is in good shape — these two are the last things.

ChronoCoders:
Both done. resend() is now per-seqnum — takes a single seq, returns Replay(Frame) or GapFill(u32); the session loops begin..=end and coalesces. store() archives the raw wire message directly (no Vec, no prepend); recover() and the resend verify both parse tag 34 via find_tag/parse_fix_seqnum. Added straddle test (stores seqs 1, 3, 5; verifies mixed Replay/GapFill across 1..=5).

Abso1ut3Zer0 (approved):
Approving — this caps off the whole arc nicely: concrete `FixJournal` on `RotatingJournal`, hot-path `append` of the raw formatted message (no re-framing, no allocation), per-seqnum `resend` with the verify-seqnum check, and a straddle test proving interleaved holes resolve correctly. 32 tests green, clippy clean. Both gaps from the last round are properly fixed at the source.

Three minor, non-blocking follow-ups captured in #515 (one cleanup PR for the batch), so we can move on:
- document the `store(seq, msg)` contract (`seq` must equal the message's `MsgSeqNum`)
- `open()`'s single-session `sessions_on_disk().first()` assumption — document or take an explicit id
- `recover()`'s per-frame scan — possible manifest-cached `last_seq` later

None of those block. Nice work seeing this through the several rounds.

PR #503, opened 2026-06-13
feat(shm): ShmSlot + ShmRingBuffer cross-process primitives (#393, #392)

### LINKED ISSUE #392: nexus-shm: ShmRingBuffer — cross-process SPSC ring buffer
Depends on: #390 (foundation layer)

## Overview

Cross-process SPSC ring buffer over shared memory. The cross-process
version of nexus-queue.

## Key differences from nexus-queue
- Backing memory is mmap'd (POSIX shm `/dev/shm` or file)
- Sequence-based so a dead reader or writer is detectable
- Recovery path for half-written slots when writer dies mid-claim
- Two-tier liveness from foundation

## Design decisions
- SPSC first. MPSC is a separate type if a caller materializes.
  No generic parameterized over topology.
- Pod types only, index-based addressing
- Backpressure model TBD: overwrite (lossy) or block. Probably both
  variants, caller chooses.

## API sketch
```rust
let (tx, rx) = shm_ring_buffer::<Order>(path, capacity)?;

// Producer
if let Some(mut claim) = tx.try_claim() {
    claim.write(order);
    claim.commit();
}

// Consumer
if let Some(msg) = rx.try_read() {
    process(msg);
}
```

## References
- Aeron — `/dev/shm` IPC transport
- nexus-queue — existing SPSC ring buffer patterns
- Agrona — ManyToOneRingBuffer `unblock()` for dead-producer recovery

## Design doc
`docs/design/nexus-shm.md`

### LINKED ISSUE #393: nexus-shm: ShmSlot — cross-process seqlock slot
Depends on: #390 (foundation layer)

## Overview

Cross-process seqlock slot for "latest value wins" patterns. Writer
always overwrites, reader gets the most recent consistent value.

## Use cases
- Config propagation (admin process → trading process)
- Book snapshots (gateway → strategy)
- Any "latest value wins" cross-process state

## Seqlock with shadow copy

Reader logic:
1. Version even → read value, verify version unchanged after read
2. Version odd (mid-write) → check atomic status (~1 cycle)
3. ALIVE → retry, writer is genuinely mid-write
4. DEAD → shadow copy + staleness flag
5. SIGKILL coverage → `peer_alive_kernel()` opt-in

Shadow = last successfully read even-version value. Explicit memcpy,
not a Copy trait bound.

Retry cost note: seqlock retry copies the whole struct per attempt.
Retry cost scales with size. Fine for low-write-rate config and book
snapshots. Worth documenting if a slot is ever both large and hot.

## Design decisions
- Pod types only (no Copy bound)
- POSIX shm (`/dev/shm`) or file backing
- Two-tier liveness from foundation

## References
- nexus-slot — existing single-process conflation slot
- nexus-queue — existing SPSC patterns

## Design doc
`docs/design/nexus-shm.md`

ChronoCoders:
Closes #393, closes #392.

## What

Two cross-process shared-memory primitives over nexus-shm Segment.

### Pod trait (nexus-shm::pod)
Marker trait for types safe in shared memory: flat repr, no heap, no Drop.

### ShmSlot (closes #393)
Cross-process seqlock slot for latest-value sharing.
- ShmSlotWriter::create / ShmSlotReader::attach
- Version counter and payload on separate cache lines (offset 64)
- Writer: odd version = mid-write, even = committed
- SlotRead: Fresh / Stale / Empty — Stale returned when writer dies mid-write
- write takes &self — safe behind Arc when writes are externally serialized

### ShmRingBuffer (closes #392)
Cross-process SPSC ring buffer.
- ShmRingWriter::create / ShmRingReader::attach
- head/tail each on own cache line; data at offset 192
- capacity stored in mmap at offset 128 (read on attach, must be power of two)
- try_push returns false when full; try_pop returns None when empty
- Both sides cache the other end's counter to reduce atomic loads (same as nexus-queue SPSC)
- Crash recovery: tail advances only after full write, so a mid-write crash leaves the slot invisible to the consumer

21 tests total, clippy clean.

### DIFF
--- a/nexus-shm/Cargo.toml
+++ b/nexus-shm/Cargo.toml
@@ -17,6 +17,9 @@ workspace = true
 crossbeam-utils.workspace = true
 nexus-platform = { version = "0.1.0", path = "../nexus-platform" }
 
+[target.'cfg(loom)'.dependencies]
+loom = "0.7"
+
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
 
--- a/nexus-shm/src/error.rs
+++ b/nexus-shm/src/error.rs
@@ -11,6 +11,8 @@ pub enum ShmError {
     HugePagesUnavailable(std::io::Error),
     OwnerActive,
     Os(std::io::Error),
+    ElemSizeMismatch { written: usize, expected: usize },
+    CorruptHeader,
 }
 
 impl fmt::Display for ShmError {
@@ -28,6 +30,10 @@ impl fmt::Display for ShmError {
             Self::HugePagesUnavailable(e) => write!(f, "huge pages unavailable: {e}"),
             Self::OwnerActive => write!(f, "segment already owned by a live process"),
             Self::Os(e) => write!(f, "{e}"),
+            Self::ElemSizeMismatch { written, expected } => {
+                write!(f, "element size mismatch: segment has {written}, reader expects {expected}")
+            }
+            Self::CorruptHeader => write!(f, "segment header is corrupt"),
         }
     }
 }
--- a/nexus-shm/src/error.rs
+++ b/nexus-shm/src/error.rs
@@ -31,7 +31,10 @@ impl fmt::Display for ShmError {
             Self::OwnerActive => write!(f, "segment already owned by a live process"),
             Self::Os(e) => write!(f, "{e}"),
             Self::ElemSizeMismatch { written, expected } => {
-                write!(f, "element size mismatch: segment has {written}, reader expects {expected}")
+                write!(
+                    f,
+                    "element size mismatch: segment has {written}, reader expects {expected}"
+                )
             }
             Self::CorruptHeader => write!(f, "segment header is corrupt"),
         }
--- a/nexus-shm/src/lib.rs
+++ b/nexus-shm/src/lib.rs
@@ -3,16 +3,19 @@
 //! Implements the foundation layer: the segment control block, the
 //! mmap-backed [`Segment`], and two-tier liveness (atomic status + OFD lock).
 //! Also provides [`ShmSlotWriter`] / [`ShmSlotReader`] for cross-process
-//! latest-value sharing via seqlock.
+//! latest-value sharing via seqlock, and [`ShmRingWriter`] / [`ShmRingReader`]
+//! for cross-process SPSC messaging.
 
 pub(crate) mod control;
 mod error;
 pub mod pod;
+mod ring;
 mod segment;
 mod slot;
 
 pub use error::ShmError;
 pub use nexus_platform::{Liveness, MapHints, MappedFile};
 pub use pod::Pod;
+pub use ring::{ShmRingReader, ShmRingWriter};
 pub use segment::{Segment, Status};
 pub use slot::{ShmSlotReader, ShmSlotWriter, SlotRead};
--- a/nexus-shm/src/lib.rs
+++ b/nexus-shm/src/lib.rs
@@ -2,11 +2,17 @@
 //!
 //! Implements the foundation layer: the segment control block, the
 //! mmap-backed [`Segment`], and two-tier liveness (atomic status + OFD lock).
+//! Also provides [`ShmSlotWriter`] / [`ShmSlotReader`] for cross-process
+//! latest-value sharing via seqlock.
 
 pub(crate) mod control;
 mod error;
+pub mod pod;
 mod segment;
+mod slot;
 
 pub use error::ShmError;
 pub use nexus_platform::{Liveness, MapHints, MappedFile};
+pub use pod::Pod;
 pub use segment::{Segment, Status};
+pub use slot::{ShmSlotReader, ShmSlotWriter, SlotRead};
--- a/nexus-shm/src/pod.rs
+++ b/nexus-shm/src/pod.rs
@@ -1,0 +1,23 @@
+/// Marker trait for types safe to place in shared memory.
+///
+/// # Safety
+/// The type must have no heap pointers, no `Drop`, and a stable
+/// binary representation (`repr(C)` or `repr(transparent)`).
+/// Any bit pattern that fits in `size_of::<Self>()` bytes must be valid.
+pub unsafe trait Pod: Sized + 'static {}
+
+unsafe impl Pod for u8 {}
+unsafe impl Pod for u16 {}
+unsafe impl Pod for u32 {}
+unsafe impl Pod for u64 {}
+unsafe impl Pod for u128 {}
+unsafe impl Pod for i8 {}
+unsafe impl Pod for i16 {}
+unsafe impl Pod for i32 {}
+unsafe impl Pod for i64 {}
+unsafe impl Pod for i128 {}
+unsafe impl Pod for f32 {}
+unsafe impl Pod for f64 {}
+unsafe impl Pod for usize {}
+unsafe impl Pod for isize {}
+unsafe impl<T: Pod, const N: usize> Pod for [T; N] {}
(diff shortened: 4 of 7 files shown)


Abso1ut3Zer0:
Solid amount of careful work here — the layout (cache-line-separated counters, power-of-two masking, cached far-end counters) is right, and the **ShmRingBuffer's SPSC ordering is correct**: writer publishes the payload then `tail.store(Release)`, reader does `tail.load(Acquire)` then reads, so there's a real happens-before and the non-atomic slot copy isn't a race; slot reuse on wrap is gated by the head release/acquire too. The seqlock is where the problems are.

## Blocking — 1. Seqlock fences are wrong-directional (`slot.rs`)

```rust
// writer
ver.fetch_add(1, Release);   // even→odd
copy(value → dst);           // payload
ver.fetch_add(1, Release);   // odd→even
```
The first `fetch_add`'s `Release` orders operations *before* it — it does nothing to stop the payload store (after it) from hoisting **above** the odd-mark. A seqlock writer needs a `fence(Release)` *between* the odd-mark and the payload write. Symmetrically on the reader:
```rust
let v1 = ver.load(Acquire);
copy(src → buf);
let v2 = ver.load(Acquire);   // wrong direction
```
`Acquire` on the `v2` load doesn't keep the payload read *above* it, so the read can sink below `v2` and the `v1==v2` check ends up validating stale data. The reader needs a `fence(Acquire)` *between* the payload read and the `v2` load.

On x86 (TSO) neither reorders, which is why the 21 tests pass — but this tears on ARM/weak memory. The non-atomic `ptr::copy` payload accesses racing the writer are also a formal data race (UB) independent of arch. The verified fence pattern already exists in `nexus-slot` (atomic payload words + correct fences); this should mirror it rather than re-derive.

## Blocking — 2. Mid-write crash detection uses the wrong liveness tier (`slot.rs`)

`read()`'s mid-write branch (`v1 & 1 == 1`) checks `self.segment.status()`, which is tier-1 — and its own doc says *"Alive may be stale if the owner died without cleanup; consult `peer_liveness`."* But an odd version *only* persists when a writer is killed between the two increments — exactly the crash tier-1 can't see. So a writer killed mid-write leaves the version odd, `status()` keeps returning `Alive`, and the reader **spins forever**. The `Stale` fallback — the feature this type exists for — never fires for a real crash. That path must consult `peer_liveness()` (the OFD lock), not `status()`. It's also untested (see #6).

## Medium

**3. No writer/reader element-size agreement (slot + ring).** Neither stores `size_of::<T>()`. The ring validates `capacity` on attach but not element size; the slot validates nothing about `T`. A reader attaching with a larger `T` than the writer created reads past the mapping → OOB/UB. Store `size_of::<T>()` (ideally a small layout tag) in the header and verify on attach, plus check the data region covers `DATA_OFFSET + capacity * elem_size`. (See the forward note below — this may resolve differently if the payload becomes byte-oriented.)

**4. Reader panics on corrupt shared memory (`ring.rs` attach).** `assert!(capacity.is_power_of_two() …)` on a value read from untrusted shm imposes a crash policy on the caller. That's the one genuinely hostile input here — return `ShmError`, don't `assert!`. (Writer-side asserts on the local caller's capacity are fine as programmer-error checks.)

**5. `write(&self)` + `Sync` on `ShmSlotWriter` is misuse-prone.** `Sync` plus a `&self` mutator means `arc.write()` compiles from N threads, which corrupts the seqlock; "serialize externally" is unenforceable for a `Sync` type. Drop the `Sync` impl (keep `Send`) so sharing `&Writer` across threads doesn't compile — the type system then enforces single-writer. (Or take `&mut self`.)

**6. No concurrency or crash tests.** All 21 tests are single-threaded and sequential — writer and reader stepped by hand in one thread. The SPSC concurrency, the seqlock racing, and the mid-write `Stale` path have no coverage; the `writer_drop` tests drop *gracefully* (even version), so the odd-version crash path is never exercised. Add a threaded producer/consumer torture test and a fork+SIGKILL-mid-write test, plus loom where feasible (matching nexus-slot).

## Low
- `DATA_OFFSET + capacity * size_of::<T>()` in `create` can overflow `usize` for a huge capacity → `checked_mul` → `Err`.
- `Pod for usize/isize` is a cross-process footgun if the two ends differ in pointer width — same-arch is assumed; worth a doc caveat.
- Read does `src→buf→shadow` (two copies of `T`); a buffer swap avoids the second for large `T`.
- A single writer could use `Relaxed` load + `Release` store instead of `fetch_add` RMW.

## Forward direction (not blocking this PR)

Worth not over-indexing on the `T: Pod` typed API. The fixed-`T`-slot model assumes both ends are Rust and agree on a compile-time struct, but the real cross-process / cross-language users send **opaque bytes with the schema defined out-of-band** — a C++ engine sending C structs, or an SBE/FIX-encoded stream, often variable-length. For those, the fundamental primitive is "move these N bytes," not "move this `T`."

So the more fundamental form is a **byte-oriented frame ring** (length-prefixed / claim-based, like `nexus-logbuf`'s `WriteClaim`/`ReadClaim`), with the typed `T: Pod` as a thin zero-copy convenience on top. That also dissolves #3 — if the primitive guarantees frame integrity and the app owns the schema/versioning (SBE template id, FIX msg type), there's no `T` for the two ends to disagree on. And these should stay pure transport in `nexus-shm`, consumed by higher layers (the FIX engine) the way it consumes `nexus-net` for TCP — not coupled to them. Happy to spec this out as a follow-up.

## Process note
This branch is stacked on #502 — it carries `persist.rs`/`store.rs`, which are still under the `FixJournal` redesign. Don't merge before #502 resolves, or rebase off that work first.

---
Net: the ring is sound; the **seqlock tears on weak memory (#1) and its crash-recovery doesn't survive a crash (#2)** — both blockers — and #3–6 are the cross-process-safety and test gaps I'd want closed before it lands.

ChronoCoders:
Addressed all blocking and medium issues. Summary of what's on the branch:

**Blocking**
- Seqlock fences corrected: `fence(Release)` between odd-mark and payload write (writer), `fence(Acquire)` between payload read and v2 load (reader); both `fetch_add` calls use `Relaxed`
- Crash detection now uses `peer_liveness()` (OFD lock) not `status()` — `Stale` fires correctly on a mid-write kill

**Medium**
- Element size stored in header for both slot and ring; `ShmSlotReader::attach` / `ShmRingReader::attach` return `ElemSizeMismatch` on disagreement
- `ShmRingReader::attach` returns `ShmError::CorruptHeader` on non-power-of-two capacity (no more panic)
- `Sync` not implemented on `ShmSlotWriter` (explicit comment explains why)
- Tests: threaded seqlock torture + mid-write crash simulation (slot), threaded SPSC producer/consumer torture (ring), loom ordering verification for ring's tail/head Release-Acquire handshake

**Low**
- `capacity * size_of::<T>()` uses `checked_mul` in ring create
- `usize`/`isize` cross-process caveat added to pod.rs

**Not addressed (forward direction per your note)**
- Byte-oriented frame ring — agreed, out of scope for this PR

ChronoCoders:
Rebased onto main to resolve conflicts from #502 and #511 landing. CI green. Re-approval needed after the force-push.

Abso1ut3Zer0 (approved):
Approving — the rebase came out clean (now pure nexus-shm, no fix-engine baggage), and the substance is solid:

- **Seqlock fences correct** — `fence(Release)` between odd-mark and payload write, `fence(Acquire)` between payload read and the re-check load; verified line-by-line.
- **Crash detection via `peer_liveness()`** (OFD lock), with a mid-write-crash test for the `Stale` path.
- **All earlier findings addressed** — elem-size stored + `ElemSizeMismatch` on attach, corrupt capacity → `CorruptHeader` (not panic), `Sync` dropped from the writer, `checked_mul` on the size calc.
- **Tests** — 28 pass including the new `threaded_spsc_torture`; `loom_ring` passes under `--cfg loom` (tail/head Release-Acquire ordering). Ring SPSC ordering confirmed correct; usize/isize Pod caveat documented.

One tracked follow-up, not blocking: #514 — the `ShmSlot` seqlock payload should move to volatile/atomic accesses (defined, rather than benign-but-UB). The fences make it correct on the supported targets today; #514 makes it correct by the model. Post-merge cleanup.

Nice work seeing the cross-process primitives through.

PR #504, opened 2026-06-13
feat(collections): v2 Wave 0 — slab types behind feature flag

### LINKED ISSUE #319: nexus-collections v2: slab-backed types migration (Wave 0)
## Summary

Major version bump (v1 → v2) for nexus-collections. Existing slab-backed types (List, Heap, BTree, RBTree) move behind a `slab` feature flag and relocate from crate root to `slab::*` submodule. This is the breaking change that enables the v2 expansion.

**This is the prerequisite for all other nexus-collections v2 items.**

## Migration

```rust
// v1
use nexus_collections::List;

// v2
use nexus_collections::slab::List;
```

```toml
# v1
nexus-collections = "1.x"

# v2
nexus-collections = { version = "2.0", features = ["slab"] }
```

## What moves

| v1 location | v2 location |
|---|---|
| `nexus_collections::List` | `nexus_collections::slab::List` |
| `nexus_collections::Heap` | `nexus_collections::slab::Heap` |
| `nexus_collections::BTree` | `nexus_collections::slab::BTree` |
| `nexus_collections::RBTree` | `nexus_collections::slab::RBTree` |

## Feature flag layout

```toml
[features]
default = ["std"]
slab = ["dep:nexus-slab"]
queue = []
cache = ["alloc"]
filter = ["alloc"]
bitmap = []
# ... (other families added as items land)
```

## Build order

**Wave 0** — do this first, before any new types land.

*From nexus-collections v2 roadmap.*

ChronoCoders:
Closes #319

Bumps nexus-collections to v2.0.0. Moves all slab-backed types (List, Heap, BTree, RbTree) to `nexus_collections::slab::*` behind an opt-in `slab` feature flag.

**Migration:**
```toml
# before
nexus-collections = "1.x"

# after
nexus-collections = { version = "2.0", features = ["slab"] }
```
```rust
// before
use nexus_collections::list::{List, ListNode};

// after
use nexus_collections::slab::list::{List, ListNode};
// or primary types at slab root:
use nexus_collections::slab::{List, ListNode};
```

- `nexus-slab` is now an optional dep, activated by the `slab` feature
- Sub-modules (`slab::list`, `slab::btree`, etc.) expose iterator/cursor/entry types
- All integration tests and benchmarks updated to v2 paths; `required-features = ["slab"]` added
- Builds clean with and without the feature; all 20 tests pass

Prerequisite for all other nexus-collections v2 items.

### DIFF
--- a/nexus-collections/CHANGELOG.md
+++ b/nexus-collections/CHANGELOG.md
@@ -10,6 +10,30 @@ contained.
 
 ## [Unreleased]
 
+### Breaking
+
+- **v2: slab types moved to `nexus_collections::slab::*`** behind the `slab` feature flag
+  (enabled by default). Sub-modules expose associated types:
+  `slab::list`, `slab::btree`, `slab::rbtree`, `slab::heap`, `slab::compare`.
+
+  ```toml
+  # unchanged for most users (slab is default)
+  nexus-collections = "2.0"
+
+  # opt-out of slab to avoid pulling nexus-slab
+  nexus-collections = { version = "2.0", default-features = false }
+  ```
+
+  ```rust
+  // before
+  use nexus_collections::list::{List, ListNode};
+
+  // after
+  use nexus_collections::slab::list::{List, ListNode};
+  // or primary types at slab root:
+  use nexus_collections::slab::{List, ListNode};
+  ```
+
 ## [1.1.4] and earlier
 
 Earlier history is not documented in this CHANGELOG. See git history
--- a/nexus-collections/Cargo.toml
+++ b/nexus-collections/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "nexus-collections"
-version = "1.1.5"
-description = "Slab-backed intrusive collections for low-latency systems"
+version = "2.0.0"
+description = "Feature-gated collections for low-latency systems"
 readme = "README.md"
 edition.workspace = true
 rust-version.workspace = true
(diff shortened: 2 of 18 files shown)


Abso1ut3Zer0:
The reorg mechanics are clean: every module consistently `#[cfg(feature = "slab")]`-gated, `slab::*` re-exports the primary types with sub-modules (`slab::btree`, `slab::heap`, …) for the associated types, and tests/benches updated with `required-features = ["slab"]`. It builds in all four feature combinations (default, `--no-default-features`, `slab`, `--no-default-features --features slab`), tests pass, clippy clean. The structure is the right shape for the v2 feature-families vision.

A few things to resolve — none about the code logic:

### 1. The `std` feature gates nothing

`std = []` is declared and defaulted, but there's no `#[cfg(feature = "std")]` anywhere and the gated modules use `std::` unconditionally (8 refs in `list.rs` alone). `--no-default-features --features slab` compiles fine, which confirms the feature is a no-op. It also diverges from the workspace convention where `std` actually does something — `nexus-slab` and `nexus-stats-core` both have `std = ["alloc"]` and are genuinely no_std-capable. Either make the crate actually no_std (core/alloc + gate the std usage, like `nexus-slab`) or **drop the `std` feature** so it doesn't advertise capability it doesn't have. For Wave 0 I'd just drop it.

### 2. Version bump + CHANGELOG — how releases work here

This repo uses a Keep-a-Changelog `## [Unreleased]` model, and the version number is **not** hand-edited in feature PRs. `tools/release.sh` (cargo-release) bumps `Cargo.toml` and renames `## [Unreleased]` → `## [x.y.z]` at release time, deriving the new version from the accumulated changelog entries. So the flow for this PR is:

- **Revert the `1.1.5 → 2.0.0` bump** in `Cargo.toml` (and the description edit if you want to keep that release-time too) — leave the version where it is on `main`.
- **Add a `## [Unreleased]` entry** to `nexus-collections/CHANGELOG.md` describing the breaking reorg, with the migration snippet. That snippet is currently only in the PR description, so it'll otherwise be lost — and the changelog content is exactly what the release tooling reads to cut the (major) version. The CHANGELOG is presently untouched by this PR.

Net: the entry under `[Unreleased]` *is* how you signal "this is the breaking 2.0.0 change"; the tooling does the bump.

### 3. Default build is an empty crate — intended?

With `slab` off by default and being the only content today, `cargo add nexus-collections` yields a crate with zero public types, and existing 1.x users face a double migration: enable the feature *and* rewrite every import path. Is empty-default the intended v2 model (opt-in families, don't pull `nexus-slab` unless used), or should `slab` be default-on until other families land, so Wave 0 is just a path change? Both are defensible — worth a deliberate decision rather than letting it fall out of the gating, and the README/migration should state whichever you pick.

---

Clean refactor overall — the substance is #1 (drop or wire up the `std` feature) and #2 (changelog entry instead of the manual version bump). #3 is a design call worth making explicit.

Abso1ut3Zer0 (approved):

PR #511, opened 2026-06-14
feat(shm): public Segment file API, generation, owner_pid (#474)

### LINKED ISSUE #474: feat(shm): public API for shared memory regions
After extracting nexus-platform (#472) and nexus-journal (#473), nexus-shm becomes the crate for working with shared memory. The internal primitives (`Segment`, `Mapping`, `MapOptions`) should become a public, ergonomic API.

## Scope

- **Region creation** — file-backed mmap, `/dev/shm`, tmpfs, anonymous mappings
- **Configuration** — size, huge pages (MAP_HUGETLB), pretouch (MAP_POPULATE), permissions
- **Pod-safe typed access** — read/write structured data with the `Pod` boundary enforced
- **Liveness/ownership** — two-tier model (atomic status + file lock via nexus-platform)
- **ShmSlot** — already exists, cross-process single-slot conflation

## Current state

- `region.rs` — `Mapping` struct wrapping `memmap2`, `MapOptions` for pretouch/huge pages
- `segment.rs` — `Segment` composes `Mapping` + control block
- `pod.rs` — `Pod` trait (plain-old-data safety boundary)
- `lock.rs` — `Liveness` two-tier liveness checking
- `control.rs` — segment control block metadata

These are all `pub(crate)` or partially public. After extraction they become the core public surface of nexus-shm.

## Future data structures (separate issues)

- `ShmQueue` — cross-process SPSC ring buffer
- `ShmLogBuf` — cross-process byte ring buffer
- Shared-memory allocator for dynamic regions

## Dependency chain

```
nexus-platform    (FileLock, mmap, OS primitives)
       ↑
   nexus-shm      (Pod, regions, ShmSlot, liveness)  ← this issue
       ↑
  nexus-journal   (SegmentedLog, Conductor, manifest, frames)
```

Related: #472 (nexus-platform), #473 (nexus-journal).

ChronoCoders:
Closes #474.

## What

- `Segment::create_file(path, data_len, hints)` — create without manually constructing `MappedFile`
- `Segment::attach_file(path)` — attach without manually constructing `MappedFile`
- `Segment::generation() -> u64` — control block generation counter
- `Segment::owner_pid() -> u32` — PID of the segment owner
- Re-export `MappedFile` from `nexus_shm` crate root for callers that still need the low-level handle

Three new tests: file-API roundtrip, generation increment on recreate, owner PID.

### DIFF
--- a/nexus-shm/src/control.rs
+++ b/nexus-shm/src/control.rs
@@ -84,6 +84,10 @@ impl ControlBlockInner {
         self.generation.load(Ordering::Acquire)
     }
 
+    pub(crate) fn owner_pid(&self) -> u32 {
+        self.owner_pid.load(Ordering::Acquire)
+    }
+
     pub(crate) const fn data_len(&self) -> u64 {
         self.data_len
     }
--- a/nexus-shm/src/lib.rs
+++ b/nexus-shm/src/lib.rs
@@ -8,5 +8,5 @@ mod error;
 mod segment;
 
 pub use error::ShmError;
-pub use nexus_platform::{Liveness, MapHints};
+pub use nexus_platform::{Liveness, MapHints, MappedFile};
 pub use segment::{Segment, Status};
--- a/nexus-shm/src/segment.rs
+++ b/nexus-shm/src/segment.rs
@@ -1,7 +1,8 @@
 use std::num::NonZeroUsize;
+use std::path::Path;
 use std::sync::atomic::AtomicU32;
 
-use nexus_platform::{Liveness, Mapping, ProcessLease};
+use nexus_platform::{Liveness, MappedFile, Mapping, ProcessLease};
 
 use crate::MapHints;
 
@@ -34,6 +35,21 @@ impl Segment {
     /// The total mapping size needed for a segment with `data_len` payload bytes.
     pub fn total_size(data_len: usize) -> Result<NonZeroUsize, ShmError> {
         HEADER.checked_add(data_len).ok_or(ShmError::SizeOverflow)
+    }
+
+    pub fn create_file(
+        path: impl AsRef<Path>,
+        data_len: usize,
+        hints: MapHints,
+    ) -> Result<Self, ShmError> {
+        let total = Self::total_size(data_len)?;
+        let mf = MappedFile::create(path.as_ref(), total)?;
+        Self::create(mf, data_len, hints)
+    }
+
+    pub fn attach_file(path: impl AsRef<Path>) -> Result<Self, ShmError> {
+        let mf = MappedFile::open(path.as_ref())?;
+        Self::attach(mf)
     }
 
     pub fn create(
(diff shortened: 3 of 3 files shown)


Abso1ut3Zer0:
Small and clean, and CI's green. The code's right — the issue is that this is a **public** shm API (the whole point of #474), and all four new public methods ship with no doc comments. A few of them have non-obvious semantics that need to be documented before this is really a public API:

### `create_file` semantics aren't self-evident
From `generation_increments_on_recreate`, calling it on an existing (dead) segment *re-incarnates* it and bumps the generation — but on a *live* segment it fails (the single-owner rule, same as `ShmSlotWriter::create`). That clobber-dead / fail-on-live-owner behavior needs to be in the doc, or a caller will assume it's either exclusive (`O_EXCL`) or unconditionally clobbering. Same for `attach_file` vs `attach`.

### `owner_pid()` is diagnostic-only — say so
PIDs are reused by the OS after a process exits, so `owner_pid` is **not** a liveness check: a new process can hold the same PID. The doc should frame it as informational/observability only and point at `peer_liveness()` for actual liveness. Without that caveat, someone will use it as a liveness gate and get burned by PID reuse.

### `generation()` — document the intended use
This should be documented as a coarse **incarnation marker** (detect "the region was recreated since I attached" across processes), explicitly *not* something to key per-access validity on. Make that intent clear in the doc so it doesn't get reached for as a slot-validity arbiter.

### Minor (not this PR)
Now that `create_file`/`attach_file` exist, `slot.rs` and `ring.rs` still do `MappedFile::create` + `Segment::create` inline — they could use the new helpers to dedup. Small follow-up.

Net: behavior is correct and tested; the gap is documentation. For a public API I'd want all four methods documented — particularly the `create_file` clobber/fail semantics and the `owner_pid`-isn't-liveness caveat — and then this is good to merge.

ChronoCoders:
Added doc comments to all four methods: create_file documents the clobber-dead/fail-on-live-owner semantics, attach_file notes the equivalent to MappedFile::open + Segment::attach, generation() is documented as an incarnation marker not a per-access validity check, and owner_pid() is marked diagnostic-only with a note that PIDs are reused and peer_liveness() should be used for actual liveness.

Abso1ut3Zer0 (approved):
LGTM — all four public methods are documented and the docs match the code (verified `create_file`'s contract: `create()` does return `ShmError::OwnerActive` on a live owner and bumps `generation` on re-incarnation). The `generation` "not a per-access validity check" and `owner_pid` "PIDs are reused, not a liveness check — use `peer_liveness()`" caveats are exactly right.

Two optional follow-ups, neither blocking:
- Add a test for the `OwnerActive` branch — it's in-process testable: create a live `Segment`, then `create_file` the same path and expect `OwnerActive`. That's the one part of `create_file`'s contract without coverage.
- `slot.rs`/`ring.rs` could now use `create_file`/`attach_file` instead of their inline `MappedFile::create` + `Segment::create`.

Abso1ut3Zer0 (approved):

PR #512, opened 2026-06-14
fix(channel): add SeqCst fence in notify_sender/notify_receiver (#428)

### LINKED ISSUE #428: nexus-channel: stress_capacity_one_high_volume hangs in CI (>60s)
## Summary

`tests::stress_capacity_one_high_volume` in `nexus-channel` occasionally hangs in CI, running for over 60 seconds without completing. The test passes locally but the CI environment (likely slower / more contended) exposes what may be a rare missed-wakeup or livelock in the capacity-1 park/unpark path.

## Location

`nexus-channel/src/lib.rs:1241`

## What the test does

- Creates a capacity-1 bounded SPSC channel
- Producer thread sends 10,000 messages sequentially
- Consumer thread receives and verifies FIFO order
- Every send/recv pair cycles through the park/unpark mechanism since the buffer is always full or empty

## Likely root cause

Capacity-1 is the worst case for the three-phase backoff (spin → yield → park). With a single slot, the producer and consumer are in lockstep — if a wakeup is missed or the parking flag check races with the unpark signal, the thread parks indefinitely. Under CI load (noisy neighbors, context switch delays), the timing window for this race widens.

## Investigation notes

- Check the `Condvar` / parking interaction for a missed-notify race
- The ordering on the atomic flag that gates parking is the prime suspect
- May also be worth checking if the spin/yield phase is too short under contention, causing both threads to park simultaneously

## Priority

Low — only observed in CI, does not affect correctness of the channel itself (the data path is fine). Flag for later investigation.

ChronoCoders:
Closes #428.

## Root cause

The ring buffer uses `Release/Acquire` fences + `Relaxed` atomics for `head`/`tail`. The park flags use `SeqCst`. These don't compose correctly.

After `pop()`, the head store (`Relaxed`, with only a `Release` compiler fence before it on x86) can still be in the CPU store buffer when `notify_sender` loads the parked flag. On x86 TSO, a `SeqCst` load is just a regular `MOV` — it doesn't drain the other thread's store buffer. So the notifier sees the flag as `false`, skips `unpark`, and the parker later sees a stale-full queue (the head update hasn't propagated) and sleeps. Nobody wakes it: deadlock.

## Fix

`fence(SeqCst)` at the top of `notify_sender` / `notify_receiver` issues an `MFENCE` on x86, flushing the store buffer before the flag load. This ensures the condition change (`head`/`tail` store) is globally visible before the flag is checked.

Combined with the `SeqCst` store in the park loop, the total-order argument holds:

- If the notifier's fence precedes the parker's `SeqCst` store: the notifier sees the flag and calls `unpark`.
- If the parker's `SeqCst` store precedes the notifier's fence: the parker sees the fresh head/tail on its condition check (the condition change was globally visible before the `SeqCst` MFENCE), so it doesn't park.

3-line change, no new allocations, no new syscalls on the uncontended fast path.

### DIFF
--- a/nexus-channel/Cargo.toml
+++ b/nexus-channel/Cargo.toml
@@ -18,6 +18,7 @@ crossbeam-utils.workspace = true
 criterion = { version = "0.5", features = ["html_reports"] }
 crossbeam-channel = "0.5"
 hdrhistogram = "7.5"
+loom = "0.7"
 rtrb = "0.3"
 
 [lints]
--- a/nexus-channel/src/lib.rs
+++ b/nexus-channel/src/lib.rs
@@ -223,6 +223,7 @@ impl<T> Sender<T> {
         // Park phase
         loop {
             self.shared.sender_parked.store(true, Ordering::SeqCst);
+            fence(Ordering::SeqCst);
 
             if self.producer.is_disconnected() {
                 self.shared.sender_parked.store(false, Ordering::Relaxed);
@@ -367,6 +368,7 @@ impl<T> Receiver<T> {
         // Park phase
         loop {
             self.shared.receiver_parked.store(true, Ordering::SeqCst);
+            fence(Ordering::SeqCst);
 
             if let Some(v) = self.consumer.pop() {
                 self.shared.receiver_parked.store(false, Ordering::Relaxed);
@@ -435,6 +437,7 @@ impl<T> Receiver<T> {
             }
 
             self.shared.receiver_parked.store(true, Ordering::SeqCst);
+            fence(Ordering::SeqCst);
 
             if let Some(v) = self.consumer.pop() {
                 self.shared.receiver_parked.store(false, Ordering::Relaxed);
--- a/nexus-channel/src/lib.rs
+++ b/nexus-channel/src/lib.rs
@@ -64,7 +64,7 @@
 use core::fmt;
 use std::mem::ManuallyDrop;
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, Ordering, fence};
 use std::time::{Duration, Instant};
 
 use crossbeam_utils::sync::{Parker, Unparker};
@@ -278,6 +278,7 @@ impl<T> Sender<T> {
 
     #[inline]
     fn notify_receiver(&self) {
+        fence(Ordering::SeqCst);
         if self.shared.receiver_parked.load(Ordering::SeqCst) {
             self.receiver_unparker.unpark();
         }
@@ -487,6 +488,7 @@ impl<T> Receiver<T> {
 
     #[inline]
     fn notify_sender(&self) {
+        fence(Ordering::SeqCst);
         if self.shared.sender_parked.load(Ordering::SeqCst) {
             self.sender_unparker.unpark();
         }
--- a/nexus-channel/tests/loom_channel.rs
+++ b/nexus-channel/tests/loom_channel.rs
@@ -1,0 +1,42 @@
+#![cfg(loom)]
+
+use loom::sync::Arc;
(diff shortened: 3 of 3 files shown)


Abso1ut3Zer0:
The diagnosis is exactly right — the Release/Acquire head/tail and the SeqCst park flags don't compose, so the notifier can read `parked == false` while its head/tail update is still in the store buffer. The notifier-side `fence(SeqCst)` is correct and necessary.

But the fix is **asymmetric**, and I think it only closes half the race. The parker side (all three park loops — sender ~225, receiver `recv` ~369, `recv_timeout` ~437) does:

```rust
self.shared.sender_parked.store(true, Ordering::SeqCst);   // A
// ...
match self.producer.push(val) { ... }                       // B: condition re-check, head.load(Acquire)
self.parker.park();
```

A SeqCst *store* followed by the condition load, with **no SeqCst fence between them**. In the abstract memory model a SeqCst store does not provide a StoreLoad barrier to a following non-SeqCst load, so `B` can hoist above `A` — which reopens the same missed-wakeup on the parker's side: it observes a stale-full queue *before* publishing `parked = true`, parks, and the notifier meanwhile reads `parked == false` and skips the unpark. The "parker sees the fresh head/tail on its condition check" step assumes `B` is ordered after `A`, and that's the part that isn't guaranteed.

It does work on the targets we ship — x86 (SeqCst store → `xchg`/`mfence`, a full barrier) and ARMv8 (`STLR`→`LDAR` aren't reordered) — but that's *hardware*-correct, not *model*-correct. Unlike a cross-process payload race, this is in-process standard concurrency where the abstract model is the contract, so I don't want to lean on architecture specifics here.

Two changes:

1. **Make it symmetric** — add `fence(SeqCst)` on the parker side too, between `*_parked.store(...)` and the `push`/`try_*` re-check, in all three loops. That's the textbook Dekker-with-fences construction and it's model-guaranteed rather than architecture-dependent.

2. **Add a loom test for the park/notify handshake.** This is the second missed-wakeup fix in this file, and the only thing that caught the first was a >60s CI hang — the worst possible detector. nexus-channel has no loom coverage today. Loom models SeqCst and would both prove the fix and settle definitively whether the parker-side fence is strictly required (passes without it ⇒ the asymmetric form is genuinely sufficient; finds the interleaving ⇒ it's needed). Either way it becomes a permanent regression guard, which is exactly what this primitive needs.

Right call on the notifier fence — just want the other side barriered and a model-checker behind it before this lands.

ChronoCoders:
Added fence(SeqCst) on the parker side between the parked.store and the push/try_* re-check in all three loops (sender ~225, recv ~369, recv_timeout ~437), making it the Dekker-with-fences construction you described. Also added a loom test (loom_channel.rs) modeling the park/notify handshake with loom atomics — passes under RUSTFLAGS='--cfg loom' cargo test.

Abso1ut3Zer0 (approved):
Approving — the symmetric `fence(SeqCst)` is correct in all three park loops, and I ran the loom test under `--cfg loom`: it passes, so the Dekker invariant is now model-verified, not just argued. Good work splitting the diagnosis from the fix.

Two follow-ups, neither blocking the merge:

1. **Wire the loom test into CI.** It's `#![cfg(loom)]`, so the normal checks compile it to nothing — right now it only runs if someone invokes `RUSTFLAGS="--cfg loom" cargo test -p nexus-channel --test loom_channel --release` by hand. A small dedicated loom job would make it re-runnable.

2. **Be aware it's a proof, not a regression guard.** It re-implements the handshake with its own loom atomics (since crossbeam's `Parker` isn't loom-instrumented — correctly noted in the comment), so it proves the fence *pattern* is sound but won't catch a fence being removed from `lib.rs` later. A true guard would need `lib.rs`'s atomics swapped to loom under `cfg(loom)` plus a parking stub — not worth it given the `Parker` blocker. The pattern proof is the high-value part and it's done.

PR #513, opened 2026-06-14
nexus-pool: Debug impls for pool and guard types

### LINKED ISSUE #507: nexus-pool: pool types missing Debug impl
## Summary

The public pool types in `nexus-pool` implement no `Debug`, so they can't be printed or nested inside `#[derive(Debug)]` structs.

## Details

None of these have a `Debug` impl (no derive, no manual impl anywhere in the crate):

- `nexus-pool/src/local.rs:138` — `BoundedPool<T>`
- `nexus-pool/src/local.rs:220` — `local::Pool<T>`
- `nexus-pool/src/sync.rs:183` — `sync::Pool<T>`
- `nexus-pool/src/local.rs:371` — `Pooled<T>`

## Approach

Prefer a **manual `Debug` impl** that prints structural info (capacity, available/in-use counts) **without requiring `T: Debug`** — pool contents are usually not interesting to print, and we don't want to force a `T: Debug` bound onto every caller. `finish_non_exhaustive()` is a good fit.

## Acceptance criteria

- [ ] All four types implement `Debug`
- [ ] No new `T: Debug` bound is added to the public API
- [ ] Output includes useful structural fields (capacity, availability), not pooled contents
- [ ] `cargo clippy --all-features --all-targets -- -D warnings` stays clean

ChronoCoders:
Closes #507

Adds manual `Debug` impls with no `T: Debug` bound to `local::BoundedPool`, `local::Pool`, `sync::Pool`, and both `Pooled` variants. Pool types report `available` (and `capacity` for `sync::Pool`); `Pooled` uses `finish_non_exhaustive`.

### DIFF
--- a/nexus-pool/src/local.rs
+++ b/nexus-pool/src/local.rs
@@ -7,6 +7,7 @@
 //! Both use LIFO ordering for cache locality.
 
 use std::cell::UnsafeCell;
+use std::fmt;
 use std::mem::{ManuallyDrop, MaybeUninit};
 use std::ops::{Deref, DerefMut};
 use std::rc::Rc;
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
LGTM — does exactly what #507 asked: manual `Debug` impls with no `T: Debug` bound, `finish_non_exhaustive()`, structural fields only (never the pooled values). Clean first contribution.

One tiny consistency nit, non-blocking: `BoundedPool`'s `Debug` prints only `available`, but it has a fixed capacity just like `sync::Pool` — adding a `capacity` field there too would make the two fixed-size pools print the same shape (`available` alone doesn't tell you 3-of-4 vs 3-of-1000). `local::Pool` correctly omits it since it's growable. Fold it in now or leave as a follow-up — either's fine.

Issue #521, opened 2026-06-14
nexus-async-fix-engine: tokio async adapter


Abso1ut3Zer0:
A tokio adapter that drives the sans-IO `Session` over async I/O, mirroring how `nexus-async-web` adapts the sans-IO `nexus-web`. Same session core, `.await` on socket read/write, tokio timers for heartbeat/TestRequest.

## Scope
- New crate `nexus-async-fix-engine` (name TBD).
- Reuse the sans-IO `Session`, `FrameReader`/`FrameWriter`; only the transport differs.
- Heartbeat/TestRequest/reconnect on tokio timers.

## Acceptance
- Interops with the same peers as the sync driver; behavioral parity.
- The conformance harness runs green against the async adapter too.

PR #535, opened 2026-06-15
feat(fix-engine): resend walk — gap-fill admin + coalesce (#520)

### LINKED ISSUE #520: nexus-fix-engine: correct resend — gap-fill admin messages + coalescing
On replaying a `ResendRequest [begin, end]`, admin messages must **never** be resent — they are gap-filled. `FixJournal::resend(seq)` already returns `Replay(frame) | GapFill(seq)` per seqnum; the session layer needs the walk on top.

## Scope (session layer)
Walk `begin..=end`, and for each seqnum:
- `GapFill(seq)` (hole / rotated out) → gap-fill
- `Replay(frame)` → classify by `MsgType` (tag 35): **admin** (Logon, Logout, Heartbeat, TestRequest, ResendRequest, SequenceReset) → gap-fill; **app** → resend with `PossDupFlag=Y` and `OrigSendingTime` (122) from the original `SendingTime`, re-framed (BodyLength/CheckSum recomputed).
- Coalesce consecutive gap-fills into a single `SequenceReset-GapFill` with `NewSeqNo`.

## Acceptance
- Replay walk implemented in the session.
- Tests: admin-skip, interior holes, straddle (below-window + in-window), PossDup re-framing, coalesced gap-fill spans, all-gap-fill range.

ChronoCoders:
Closes #520.

## What

`FixJournal::resend_range(begin, end, emit)` — the replay walk the session layer was missing.

Walks `begin..=end` (0 = unbounded, up to `next_outbound - 1`), and for each seqnum:
- **Hole / rotated-out** (`GapFill`) → gap-fill
- **Admin message** (Logon A, Logout 5, Heartbeat 0, TestRequest 1, ResendRequest 2, SequenceReset 4) → gap-fill
- **App message** → re-frame with `PossDupFlag(43)=Y` + `OrigSendingTime(122)` from stored `SendingTime(52)`; recomputed `BodyLength`/`CheckSum`
- Consecutive gap-fills coalesce into a single `ReplayItem::GapFill { seq, new_seq }`

`SessionState::on_resend_request` no longer emits a placeholder `SequenceReset`; it surfaces `Event::ResendRange` so the caller drives the walk via `resend_range`.

## Tests

- `resend_range_admin_skip` — all-admin range → single coalesced gap-fill
- `resend_range_interior_holes` — stored 1,3,5; range 1..5 → Replay/GapFill alternating, each gap is per-seqnum
- `resend_range_straddle_window` — window=4, stored 1..8; seqs 1..4 rotated out → one coalesced GapFill + 4 App
- `resend_range_poss_dup_reframe` — checks tag 43=Y, tag 122=original SendingTime, valid checksum
- `resend_range_coalesced_gapfill` — stored 1 and 5; range 1..5 → App, GapFill{2..5}, App
- `resend_range_all_gapfill` — all admin → single GapFill
- `resend_range_end_zero_means_all` — end=0 walks up to next_outbound-1

### DIFF
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -12,10 +12,6 @@ mod framework;
 #[cfg(unix)]
 pub mod persist;
 mod session;
-#[cfg(unix)]
-mod timestamp;
-#[cfg(unix)]
-pub mod transport;
 
 pub use frame::{
     FrameError, FrameReader, FrameReaderBuilder, FrameWriter, FrameWriterBuilder, ReadError,
@@ -24,5 +20,3 @@ pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
 pub use persist::{FixJournal, ReplayItem};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
-#[cfg(unix)]
-pub use transport::{Error as TransportError, FixConnection, FixConnectionBuilder};
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -17,4 +17,6 @@ pub use frame::{
     FrameError, FrameReader, FrameReaderBuilder, FrameWriter, FrameWriterBuilder, ReadError,
 };
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
+#[cfg(unix)]
+pub use persist::{FixJournal, ReplayItem, ResendPlan};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -12,11 +12,17 @@ mod framework;
 #[cfg(unix)]
 pub mod persist;
 mod session;
+#[cfg(unix)]
+mod timestamp;
+#[cfg(unix)]
+pub mod transport;
 
 pub use frame::{
     FrameError, FrameReader, FrameReaderBuilder, FrameWriter, FrameWriterBuilder, ReadError,
 };
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
-pub use persist::{FixJournal, ReplayItem, ResendPlan};
+pub use persist::{FixJournal, ReplayItem};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
+#[cfg(unix)]
+pub use transport::{Error as TransportError, FixConnection, FixConnectionBuilder};
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -13,9 +13,7 @@ pub enum ReplayItem<'a> {
     /// Coalesced gap-fill: encode as `SequenceReset(GapFillFlag=Y)` with
     /// `MsgSeqNum=seq` and `NewSeqNo=new_seq`.
     GapFill { seq: u32, new_seq: u32 },
-    /// Original stored bytes for the app message; the transport is responsible
-    /// for adding `PossDupFlag(43)`, `OrigSendingTime(122)`, and a fresh
-    /// `SendingTime(52)` before retransmitting.
+    /// Original stored bytes; transport reframes with `PossDupFlag` and fresh timestamps.
     App(&'a [u8]),
 }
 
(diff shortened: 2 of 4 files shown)


Abso1ut3Zer0:
The resend *walk* here is right — coalescing is correct (`[gs, x-1]` → `GapFill { seq: gs, new_seq: x }`, `new_seq = high+1` for a trailing gap), BodyLength/CheckSum recompute is correct, and the admin-skip / interior-holes / straddle / end=0 cases are well tested. Two things — and one is a layering change we landed after discussion:

**1. `is_admin_type` is missing MsgType `3` (Reject).** The session-level types are `0,1,2,3,4,5,A`; this omits `3`. Reject is administrative and must be gap-filled on resend — as written, a stored Reject is replayed as an app message. One-char fix: add `b"3"`. This is classification, so it **stays here in the journal**.

**2. `reframe` should move out of the journal into the transport (#536).** Re-framing is a wire concern — it inserts `PossDupFlag(43)`/`OrigSendingTime(122)` and, critically, needs a *fresh* `SendingTime(52)`, which the journal can't set (no clock). Cleaner split:
- `resend_range` does the walk and yields the **borrowed original bytes** (`App(&[u8])` off the mmap) + `GapFill` markers — ideally as an iterator;
- drop `reframe` and `ReplayItem::App(Vec<u8>)` (that's the per-message allocation);
- the transport reframes into its `FrameWriter` with a fresh `SendingTime` and batch-flushes (see #536).

This **supersedes my earlier SendingTime / PossDup-on-replay notes here** — they're addressed transport-side now. Full reshape is on #536.

Net: keep the walk + the Reject(`3`) fix; hand back borrowed originals instead of reframed `Vec`s, and let the transport own the wire framing.

Abso1ut3Zer0:
Both prior findings are addressed — Reject(`3`) is in the admin set, and `resend_range` now yields borrowed originals (`ReplayItem::App(&'a [u8])`, `reframe` gone). The layering is right and CI's green. One last API refinement before this is done:

**Return an iterator instead of taking `FnMut`.** `resend_range<'a, F: FnMut(ReplayItem<'a>)>(…, emit)` → `fn resend(&'a self, begin, end) -> impl Iterator<Item = ReplayItem<'a>>`. Two reasons, the second is the important one:

1. **Fallibility** — the consumer's per-item work (reframe → write into the `FrameWriter` → flush to the socket) is fallible. An `FnMut` can't propagate that out of the walk; you'd stash the error in the closure env and check after (or panic). An iterator lets the consumer `?` per item and `break`/batch-flush mid-stream.

2. **Async** — this is what makes it matter for any adapter later. An `FnMut` is synchronous, so you **can't `.await` inside it**. For the async adapter (#521), the per-item work is an async socket write — with a callback you'd be forced to either collect every item into owned buffers first (which copies the borrowed bytes and throws away the whole borrowed-`App(&[u8])` win) and await afterward, or `block_on` inside the closure (blocks the executor). The iterator inverts control: the walk stays sync (pure-CPU — read mmap, classify, coalesce, no IO) and the consumer drives it, choosing sync or `.await` in the loop body:
   ```rust
   for item in journal.resend(begin, end) { sink.write(reframe(item)).await?; }
   ```

Keep it a plain **`Iterator`, not a `Stream`** — the producer does zero IO, so async machinery there buys nothing; the async-ness belongs in the consumer where the socket is. Coalescing fits fine in `next()` (advance through a run of gaps, then yield the coalesced `GapFill`; emit a final one for a trailing gap), and since the items borrow the long-lived journal it's a clean `impl Iterator<Item = ReplayItem<'a>> + 'a`, not a lending-iterator problem.

Net: one iterator-returning `resend` backs both the sync transport (#536) and the async adapter (#521). With that, #535 is done.

Abso1ut3Zer0 (approved):
LGTM — every finding is closed:

- **Reject (`3`)** is in the admin set.
- `resend` yields **borrowed originals** (`ReplayItem::App(&[u8])`) — no reframe in the journal, no per-message `Vec`.
- **`resend(begin, end) -> impl Iterator<Item = ReplayItem<'_>>`** — I traced `next()` against the old callback: coalescing is identical (`App(1), GapFill{2→3}, App(3), GapFill{4→5}, App(5)` on stored `1,3,5`), the `deferred` slot correctly emits "coalesced GapFill then App" across two `next()`s, and trailing-gap / all-gap / empty all check out. Lifetimes are sound — items borrow the long-lived journal, so `deferred` holding a `ReplayItem<'a>` is fine (not a lending-iterator issue).

39 tests pass (full resend matrix), clippy clean, CI green incl. loom. The iterator is what lets one `resend` back both the sync transport (#536) and the future async adapter (#521). Nice work seeing this through the several rounds.

Issue #517, opened 2026-06-14
nexus-fix-engine: synchronous IO transport (drive the sans-IO session over TCP)


Abso1ut3Zer0:
The sans-IO session core is done (#408, closed): `Session::on_message` is driven by the caller over buffers, and `FrameReader`/`FrameWriter` work on `nexus_net::buf` — no sockets. There is no component that owns a socket and drives the loop. This issue adds the synchronous transport.

## Scope
A blocking (and/or mio-based) TCP driver that runs the session end-to-end:
- connect / accept, TLS optional (via nexus-net)
- read loop → `FrameReader` → `Session::on_message` → encode outbound via `FrameWriter` → write
- heartbeat + TestRequest timers, logon/logout sequencing
- reconnect with ResendRequest/replay on gap detection
- backpressure + clean shutdown

Mirrors how `nexus-web` pairs the sans-IO parser with an IO loop; keeps the sans-IO core untouched.

## Acceptance
- Connect and maintain a session against a peer; exchange app messages, heartbeats, graceful logout, reconnect-with-resend.
- No async; tokio support is a separate crate (see the async adapter issue).

Abso1ut3Zer0:
Delivered by #536 (synchronous TCP transport), refined by #539 and #548 (typed `recv() -> Message<'_, D>`). `FixConnection<S, D>` drives the sans-IO session end-to-end — connect/accept, read→frame→session→write, heartbeat/TestRequest timers, logon/logout, and resend-on-gap — and is exercised by the conformance suite (#540). Closing as completed.

Issue #518, opened 2026-06-14
nexus-fix-engine: multi-session support (N sessions per process / journal dir)


Abso1ut3Zer0:
Today `FixJournal::open` selects `sessions_on_disk().first()` — one session per directory. The `Conductor` already supports multiple sessions via `session_id`, so the journal layer can do it; the engine just hardcodes a single session.

## Scope
- Run multiple sessions in one process against the same conductor/journal directory, each with its own session files (manifest, lock, segments) and its own inbound/outbound seqnum state — **no shared session state**.
- `FixJournal` takes/derives an explicit session id rather than `.first()`.
- Recovery selects the correct session.

## Acceptance
- Open N `FixJournal`s under one dir with distinct session ids; concurrent sessions don't clobber each other's files or seqnums.
- Restart recovers each session's `next_outbound` independently.
- Test with ≥2 concurrent sessions in one process.

On0n0k1:
Hello. I created a Pull Request for the requested changes. Please let me know if there are changes I should make.

For the test to be created, you mentioned it needed to run "concurrently". I implemented it to just load both FixJournal twice to assure the output is the expected.

ChronoCoders:
Hey, saw you picked up #518 — nice. Let me know if you need a review or have questions on the session layer.

Abso1ut3Zer0:
Delivered by #570. `FixJournal` opens by explicit `session_id`, so concurrent sessions in one directory keep separate manifest, lock, segment, and seqnum state, and each recovers `next_outbound` independently. Collapsing the per-session conductors into one shared archival thread is tracked separately in #575.

PR #536, opened 2026-06-15
feat(fix-engine): synchronous TCP transport (#517)

ChronoCoders:
## Summary

- Add `transport.rs`: `FixConnection<S>` drives the sans-IO `SessionState` over any `Read + Write` stream
- Encode all admin messages (`Logon`, `Logout`, `Heartbeat`, `TestRequest`, `ResendRequest`, `SequenceReset`) via `FrameFormatter`
- Timer polling via 100 ms read timeout; `on_timeout()` fires heartbeats and test-request probes
- ResendRange events drive `FixJournal::resend_range`; gap-fills encoded as `SequenceReset(GapFillFlag=Y)`, app replays written raw
- `FixConnectionBuilder` provides `connect(addr, …)` (initiator) and `accept(stream, …)` (acceptor) constructors
- `send_app(seq, frame)` stores in journal and flushes to socket
- Wire up `mod timestamp` (previously orphan file)
- 3 integration tests covering initiator logon+logout, acceptor app-message delivery, and resend-request gap-fill

## Test plan

- [ ] `cargo test -p nexus-fix-engine` — all 79 tests pass (43 unit + 9 framework + 24 session + 3 transport)
- [ ] `cargo clippy -p nexus-fix-engine --tests` — clean
- [ ] CI lint/build/test

### DIFF
--- a/nexus-fix-engine/src/frame.rs
+++ b/nexus-fix-engine/src/frame.rs
@@ -7,6 +7,7 @@
 //! are invariant across all FIX versions.
 
 use nexus_net::buf::{ReadBuf, WriteBuf};
+use nexus_net::wire::ParserSink;
 
 const SOH: u8 = 0x01;
 
@@ -297,6 +298,18 @@ impl FrameReader {
         let len = data.len();
         self.buf.advance(len);
         len
+    }
+}
+
+impl ParserSink for FrameReader {
+    #[inline]
+    fn spare(&mut self) -> &mut [u8] {
+        self.buf.spare()
+    }
+
+    #[inline]
+    fn filled(&mut self, n: usize) {
+        self.buf.filled(n);
     }
 }
 
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -12,6 +12,10 @@ mod framework;
 #[cfg(unix)]
 pub mod persist;
 mod session;
+#[cfg(unix)]
+mod timestamp;
+#[cfg(unix)]
+pub mod transport;
 
 pub use frame::{
     FrameError, FrameReader, FrameReaderBuilder, FrameWriter, FrameWriterBuilder, ReadError,
@@ -19,4 +23,6 @@ pub use frame::{
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
 pub use persist::{FixJournal, ReplayItem, ResendPlan};
+#[cfg(unix)]
+pub use transport::{Error as TransportError, FixConnection, FixConnectionBuilder};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -23,6 +23,6 @@ pub use frame::{
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
 pub use persist::{FixJournal, ReplayItem, ResendPlan};
+pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
 #[cfg(unix)]
 pub use transport::{Error as TransportError, FixConnection, FixConnectionBuilder};
-pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -22,7 +22,7 @@ pub use frame::{
 };
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
-pub use persist::{FixJournal, ReplayItem, ResendPlan};
+pub use persist::{FixJournal, ReplayItem};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
 #[cfg(unix)]
 pub use transport::{Error as TransportError, FixConnection, FixConnectionBuilder};
--- a/nexus-fix-engine/src/lib.rs
+++ b/nexus-fix-engine/src/lib.rs
@@ -18,5 +18,5 @@ pub use frame::{
 };
 pub use framework::{CompId, Message, Session, SessionConfig, SessionError};
 #[cfg(unix)]
-pub use persist::{FixJournal, ReplayItem};
+pub use persist::{FixJournal, ReplayItem, ResendPlan};
 pub use session::{AdminMsg, DisconnectReason, Event, Out, SessionState, State};
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -252,7 +252,7 @@ mod tests {
 
         match j.resend_one(2) {
             ResendPlan::GapFill => {}
-            _ => panic!("expected GapFill(2)"),
+            ResendPlan::Replay(_) => panic!("expected GapFill"),
         }
 
         cleanup(&dir);
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -1,16 +1,23 @@
 use std::path::Path;
 
-use nexus_fix_codec::{find_tag, parse_fix_seqnum};
+use nexus_fix_codec::{
+    FieldReader, checksum, encode_fix_uint, find_tag, format_checksum, parse_fix_seqnum,
+};
 use nexus_journal::{Conductor, Frame, LogOffset, OpenError, RotatingJournal, WriteError};
 
-enum ResendPlan<'a> {
+pub enum ResendPlan<'a> {
     Replay(Frame<'a>),
-    GapFill,
+    GapFill(u32),
 }
 
-pub enum ReplayItem<'a> {
+/// Output of a single [`FixJournal::resend_range`] step.
+pub enum ReplayItem {
+    /// Coalesced gap-fill: encode as `SequenceReset(GapFillFlag=Y)` with
+    /// `MsgSeqNum=seq` and `NewSeqNo=new_seq`.
     GapFill { seq: u32, new_seq: u32 },
-    App(&'a [u8]),
+    /// Re-framed app message with `PossDupFlag(43)=Y` and
+    /// `OrigSendingTime(122)` from the original `SendingTime(52)`.
+    App(Vec<u8>),
 }
 
 pub struct FixJournal {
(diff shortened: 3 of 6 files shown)


Abso1ut3Zer0:
Solid transport — clean `connect`/`accept` builder, correct partial-read framing (the read timeout only fires when the socket is idle, so partial messages accumulate without loss), CompId check in the right direction, sans-IO session layering preserved (the session owns seq validation; `on_app` only fires on `Event::App`), and the gap-fill `SequenceReset` is encoded correctly (`PossDup=Y` + fresh `SendingTime` + `GapFill=Y`). CI green incl. loom. Findings:

**1. Drop `run()` — mirror the nexus-web / nexus-async-web decomposition (architectural, headline).** The session core is already sans-IO (`SessionState` + `FrameReader`/`FrameWriter`), but `run()` bolts a blocking, thread-per-session loop on top — which doesn't scale to **multi-session (#518)** (N sessions = N parked threads), contradicts this crate's own "the caller owns the transport, the clock, and the loop" contract, and can't compose into a reactor. The threading/loop is policy that belongs to the user, not the engine.

nexus-web already solved this exact problem — match it rather than invent a new API:
- the reader should impl **`nexus_net::wire::ParserSink`** (`spare`/`filled`/`next`) — the FIX `FrameReader` already has those method shapes, it just needs to implement the trait, so it plugs into the same byte-fill machinery;
- `FixConnection<S>` becomes the analog of nexus-web's **`Client<S>`**: caller-driven `recv()` / single-step + `send_app`/`logout`, with `wants_read()`/`wants_write()` readiness (like ws `Connecting::poll`) so the caller's reactor knows what to wait on — **not** a blocking `run()` with a hardcoded 100 ms tick;
- **tokio support is a separate `nexus-async-fix-engine` (#521)** that drives the same core via `nexus_net::wire::WireStream` (`poll_fill_into`/`poll_write`/…), exactly how `nexus-async-web` adapts `nexus-web`. Async never belongs in this crate.

Net: ship the pieces; the user composes them into their own engine/reactor (one thread, many sessions → #518) or uses the async adapter. A blocking loop, if anyone wants one, is an `examples/blocking_session.rs` recipe — not a library API.

**2. Resend path: borrow → reframe-into-`FrameWriter` → batch (folds four findings).** Today `do_resend` collects all `ReplayItem`s (each an owned `Vec<u8>` from `reframe`) then `write_all`s each app message directly. Reshape:
- the journal yields **borrowed originals** via an iterator (the #535 change — `reframe` leaves the journal);
- the transport reframes each app message into the `FrameWriter` (`PossDup`/`OrigSendingTime`/**fresh `SendingTime`** — the transport has the clock — recomputed `BodyLength`/`CheckSum`), synthesizes gap-fills into the same writer, and **batch-flushes**;
- the borrow conflict that forced the collect-all dissolves by destructuring `self` (`journal` `&`, `writer`/`stream` `&mut` — disjoint fields).

A resend isn't a verbatim retransmit (PossDup/OrigSendingTime are mandatory, which changes BodyLength/CheckSum), so there's one unavoidable transform per app replay — but it lands directly in the send buffer, batched out, no intermediate `Vec`. This single change removes the per-message `Vec`, the collect-all, the **two-write-paths ordering hazard** (admin via the writer vs app via direct `write_all` — safe today only because `flush_out` always drains first), and the **stale `SendingTime`** on app replays (inherited from #535 — fixed naturally here now that reframe has the clock).

**3. Framing errors swallowed — `Err(_) => {}`.** Won't infinite-loop (`FrameReader::next` self-advances via `pending_advance`), but malformed frames drop silently — no Reject, no log. Garbage is recoverable via the seq layer, but **`MessageTooLarge` should be fatal**: if a valid message exceeds your buffer you can never receive it, and silently skipping it desyncs the session. At minimum, distinguish the error variants.

**Minor:** repeated `find_tag` full-frame scans per inbound message (a single header parse would do); `encode_admin`'s `_now` is unused (it correctly uses `SystemTime::now()` for wall-clock tag 52 — just drop the param).

Net: framing and admin-encoding details are right. **#1 is the big one — drop `run()`, mirror nexus-web's `Client<S>` + `ParserSink`/`WireStream` decomposition so the user composes the pieces into their own engine (and #521 adapts it for tokio like nexus-async-web). This is the foundation for multi-session #518.** **#2** is the resend reshape (reframe moves out of #535's journal into here). **#3** is the framing-error fix. The Reject(`3`) classification fix stays on #535.

Abso1ut3Zer0:
Big reshape, and it's almost there — most of the prior review is addressed:

- **`run()` → `recv(now, on_app)`** (caller-driven) + `wants_read`/`wants_write` + **`ParserSink for FrameReader`** — the nexus-web `Client<S>` shape.
- **Framing errors fixed** — `MessageTooLarge` is fatal (`Err(FrameTooLarge)`), garbage skipped explicitly.
- **Resend reshaped** — `do_resend` borrows disjoint fields and reframes the borrowed `App(&orig)` straight into the `FrameWriter`, batch-flushing once. `reframe_app` writes a **fresh `SendingTime(52)`** + `PossDup(43)=Y` + `OrigSendingTime(122)` — the stale-time bug is gone now that the transport (which has the clock) owns the reframe.

Two things left, and both are about aligning to what nexus-web's `Client<S>` already settled on:

**1. Silent drop on writer-full during resend.** `reframe_app`/`encode_gap_fill` ignore `fmt.finish()` errors and `do_resend` only flushes at the very end, so a resend larger than the 64 KB writer fills up and silently drops messages. This resolves with the **#535 iterator switch** — once `resend` returns `impl Iterator`, `do_resend` becomes a `for` loop that can flush-when-full and `?`:
```rust
for item in journal.resend(begin, end) {
    if !fits(&self.writer, &item) { self.flush_writer()?; }
    write_item(&mut self.writer, item)?;   // propagate, don't silently skip
}
```
The `FnMut` closure can't do either, which is why it drops silently today.

**2. Route `send_app` through the `FrameWriter` — align to nexus-web's write path.** nexus-web's `Client` has one write path: `send_*` → `encode_*` into the `FrameWriter` → `flush_write_buf()` → `stream.write_all(writer.data())`. The FIX transport already does this for admin/resend, but `send_app` writes **directly** via `write_all`, bypassing the writer — a residual dual-path ordering hazard if the writer has pending bytes. Fix: `send_app` stores in the journal, then encodes into the writer + flushes, like every other send. One buffered-then-flush path, matching `Client`.

For reference, nexus-web's surface (worth matching, since it's the proven one):
- **read** — `recv() -> Result<Option<_>>` returns `Ok(None)` on `WouldBlock` (so on a non-blocking socket, `recv` *is* the try-read). The FIX `recv()` already matches — no separate `try_recv` needed.
- **write** — encode into the writer, blocking `flush` via `write_all`. No `try_write`.
- **async** — lives in the separate `nexus-async-web` crate via `WireStream` (`poll_fill_into`/`poll_write`/`poll_flush`), not as `try_*` on the sync client. The FIX equivalent is **#521** (`nexus-async-fix-engine`).

So no new non-blocking method surface to invent — just route every send through the `FrameWriter` (#2) and let the iterator fix the resend drop (#1). The sync blocking-loop path is solid today; the async/reactor story is #521 via `WireStream`, exactly like nexus-async-web.

Abso1ut3Zer0:
Both prior items are addressed — `do_resend` now drives the `resend` iterator with flush-when-full + `?`, and `send_app` routes through the `FrameWriter`. 👍

One new edge, surfaced by the buffered write path itself — **oversized frames (larger than the writer's capacity).** `FrameWriter` wraps a fixed-capacity `WriteBuf` (default 64 KB), so after a flush `spare()` is at most `capacity`:

**`send_app` panics on a frame bigger than the writer.**
```rust
if self.writer.remaining() < frame.len() { self.flush_writer()?; }
let spare = self.writer.spare();
spare[..frame.len()].copy_from_slice(frame);   // out of bounds if frame.len() > capacity → panic
```
The guard only handles "doesn't fit the *remaining* space"; after flushing, `remaining() == capacity`, but a `frame.len() > capacity` still slices past `spare`. FIX RawData/XMLData/EncodedText messages can exceed 64 KB, so it's reachable. (The old `write_all`-based `send_app` couldn't hit this — it's new to the buffered version.)

**`do_resend`'s retry silently drops the same case.** After the flush-then-retry, an item that still doesn't fit the (now empty) writer gets `.ok()`'d away.

Both want the same handling: after flushing, if it still doesn't fit, **`write_all` it directly** (you just flushed, so there's nothing to reorder against) or return an error — don't index a fixed `spare` (panic) or `.ok()` it (drop). It's the write-side mirror of the `MessageTooLarge → FrameTooLarge` you already handle on the read side; a small shared "flush, then buffer-or-write-direct" helper covers `send_app` and `do_resend` both.

(Mechanical, separate: rebase onto the merged #535 so its persist.rs commits drop out of the diff.)

Abso1ut3Zer0:
The oversized-frame handling is right now — `write_through` guards the copy with a `remaining() >= len` check and writes oversized frames directly (after flushing, so no reorder), and `do_resend`'s retry surfaces `FrameTooLarge` instead of `.ok()`-dropping. With the recv/ParserSink/wants_read reshape and the resend-into-writer path, this is basically there. One cleanup and one note:

**1. Drop the `persist.rs` change — it's outside this PR's scope and regresses docs.** The diff shows `persist.rs -14`: it deletes the `FixJournal::open` and `store` doc comments that are **already on `main`** (merged via #535). The `store` one is the contract doc — `seq` must equal the message's `MsgSeqNum(34)` — which is worth keeping. `persist.rs` belongs to #535 (merged); #536 shouldn't touch it. Please restore those comments / drop the `persist.rs` delta so the PR is transport-only. (It slips past CI because the crate doesn't `#![warn(missing_docs)]`, but it's a real doc loss.)

**2. Minor / known limitation (not blocking):** `do_resend` errors with `FrameTooLarge` on a replay larger than the writer capacity, because `reframe_app` builds the reframed message *into* the writer and can't exceed it — whereas `send_app` can direct-write already-formed bytes. So the writer capacity (default 64 KB) caps the max *resendable* message size, while *sending* one that big works. It's surfaced as an error (not silent or a panic), and >64 KB single FIX messages are rare, so this is fine to leave — just worth knowing that "can send" and "can resend" have different ceilings.

Otherwise this is good — drop the `persist.rs` doc removal and it's transport-only and ready.

Abso1ut3Zer0 (approved):
LGTM — this is the synchronous TCP transport (#517), landed in the right sans-IO `Client<S>` shape, and every finding across the rounds is closed:

- `run()` → caller-driven `recv` + `wants_read`/`wants_write` + `ParserSink for FrameReader`
- `MessageTooLarge` fatal, garbage skipped
- resend via the `resend` iterator → reframed into the `FrameWriter` with a fresh `SendingTime`, batch-flushed
- `send_app` through the writer; oversized frames handled without panic (`write_through` check-then-copy-or-direct) or silent drop (`do_resend` → `FrameTooLarge`)
- `persist.rs` docs restored — diff is transport-only

43 + 9 + 24 + 3 + 2 tests pass locally, clippy clean, CI green incl. loom.

Deferred follow-ups (none blocking) are catalogued in **#537** — the resendable-size cap, a `blocking_session.rs` example, the `dispatch` header-parse perf nit, and the note that reactor/non-blocking writes come with #521/#518. Worth adding `Closes #517` to the PR body on merge so the transport ticket closes.

Nice work seeing this through the reshape — sans-IO core, caller-driven loop, async cleanly deferred to #521.

Issue #505, opened 2026-06-13
nexus-stats: time-aware types should take integer timestamps (deprecate f64-timestamp variants)


Abso1ut3Zer0:
## Summary

Two time-aware types in `nexus-stats` take an `f64` **timestamp**, which is a precision hazard. They should take integer timestamps (`u64`/`i64`) like every other time-aware type in the crate (`EventRate`, `Hawkes`, `CoDel` all already do).

## Why f64 timestamps are wrong

`f64` has a 52-bit mantissa, so it represents integers exactly only up to ~4.5e15. Real clocks blow past that:

- epoch **nanoseconds** today ≈ 1.75e18 — ~500× past the exact-integer ceiling → ~256–512 ns of silent quantization
- epoch **microseconds** ≈ 1.75e15 — already brushing the ceiling
- even a monotonic `Instant` in nanos exceeds 2^52 after ~52 days of uptime

So `update(now.as_nanos() as f64)` silently buckets time — a correctness bug that only surfaces at scale/uptime. Absolute time is integer by nature.

## The two offenders

- **`LivenessF64`** (`nexus-stats-core/src/monitoring/liveness.rs:57`) — `update(&mut self, timestamp: f64)`. A `LivenessI64` already exists, so the fix is to **deprecate/remove the f64-timestamp variant** (keep the integer one).
- **`DecayAccumF64`** (`nexus-stats-control/src/frequency/decay_accum.rs`) — `score(&mut self, now: f64)` and the corresponding `update`. No integer variant exists; convert it to integer timestamps.

## The decay-math nuance (DecayAccum)

DecayAccum needs `f64` for `exp(-λ·Δt)`, but that does **not** require an f64 timestamp. Compute the delta in integer space and convert only the (small) delta:

```rust
let delta = now_u64 - last_u64;          // exact integer
let f = delta as f64;                     // small interval, converts precisely
value *= (-lambda * f).exp();
```

The `f64` only ever touches a small inter-event delta, never an absolute timestamp.

## Acceptance criteria

- [ ] `DecayAccum` takes integer (`u64`/`i64`) timestamps; integer delta internally, `f64` only for the `exp`
- [ ] `LivenessF64`'s f64-timestamp surface is deprecated/removed (keep `LivenessI64`)
- [ ] Tests updated to use integer timestamps
- [ ] No f64-timestamp surface remains in the time-aware API

[...]

PR #538, opened 2026-06-17
feat(stats)!: time-aware types should take integer timestamps (deprecate f64-timestamp variants) (#505)

On0n0k1:
Apologies for including all changes into a single commit. Next time will make the changes more precise.

- Remove LivenessF64 and its tests; LivenessI64 remains unchanged
- Add LivenessU64 alongside LivenessI64, using wrapping_sub as i64 for safe unsigned timestamp arithmetic and bit-shift EMA (same pattern as  EventRateU64)
- Replace DecayAccumF64 with DecayAccumU64 and DecayAccumI64; integer delta computed first, f64 only for the exp() call
- Update tests to integer timestamps for all affected types
- Add [Unreleased] breaking entries to nexus-stats-core and nexus-stats-control CHANGELOGs

### DIFF
--- a/nexus-stats-control/CHANGELOG.md
+++ b/nexus-stats-control/CHANGELOG.md
@@ -8,6 +8,17 @@ with the project-specific allowance that a minor bump may carry small,
 narrowly-scoped breaking changes when external blast radius is
 contained.
 
+## [Unreleased]
+
+### Removed
+
+- **`DecayAccumF64`** — f64 timestamps cause silent precision loss past 2^52; replaced by `DecayAccumU64` and `DecayAccumI64`.
+
+### Added 
+
+- **`DecayAccumU64`** — decaying accumulator with `u64` timestamps.
+- **`DecayAccumI64`** — decaying accumulator with `i64` timestamps.
+
 ## [2.0.0] — 2026-05-28
 
 ### Removed
--- a/nexus-stats-control/src/frequency/decay_accum.rs
+++ b/nexus-stats-control/src/frequency/decay_accum.rs
@@ -8,14 +8,14 @@
 /// - "How active has this been recently?"
 /// - Rate limiting with smooth backoff
 #[derive(Debug, Clone)]
-pub struct DecayAccumF64 {
+pub struct DecayAccumU64 {
     score: f64,
-    last_time: f64,
+    last_time: u64,
     decay_constant: f64, // ln(2) / half_life
     initialized: bool,
 }
 
-impl DecayAccumF64 {
+impl DecayAccumU64 {
     /// Creates a new decaying accumulator with the given half-life.
     ///
     /// `half_life` is in the same time units as the timestamps passed to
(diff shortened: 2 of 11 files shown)


ChronoCoders:
Build fails: `nexus-stats/examples/perf_stats.rs:17` still imports `LivenessF64` which is removed. That needs updating.

Correctness issue in `LivenessU64::is_alive_with`: `deadline_absolute` is `u64` but gets cast to `i64`. Values above `i64::MAX` wrap negative and the check always fails. Use `dt >= 0 && (dt as u64) <= absolute` instead.

Abso1ut3Zer0:
The stats work here is exactly right — `LivenessU64` faithfully mirrors `EventRateU64` (`wrapping_sub as i64`, `i128` acc, bit-shift EMA), `DecayAccumU64`/`I64` do the integer-delta-then-`exp()`, `F64` is gone with `I64` kept and `U64` added, and the CHANGELOG `Removed` entries with the precision rationale are spot on. Tests + clippy are clean on both stats crates. 

Just the one CI build failure to fix, and it's a missed reference, not your core change: the **umbrella `nexus-stats` crate's** `examples/perf_stats.rs` still uses the removed `LivenessF64`:
- line 17 — the `use` import
- line 241 — `LivenessF64::builder()`
- line 987 — the `"LivenessF64::update"` row label

Swap those to `LivenessI64` (or `LivenessU64`) and feed it integer timestamps in the bench loop. It slipped through because that example lives in `nexus-stats` (the re-export crate), separate from `nexus-stats-core` where you made the change — easy to miss.

Optional aside while you're here: `DecayAccumI64` uses `timestamp - self.last_time`; `wrapping_sub` like the `U64` variant (and `EventRate`) is a touch safer on overflow and keeps the two consistent.

Once CI's green I'll approve — this is good work, just that one example to update.

On0n0k1:
Hello. I included the changes @ChronoCoders @Abso1ut3Zer0 mentioned. Please let me know if any more changes are needed.

On0n0k1:
In `liveness.rs`, I had to change the tests in lines 455 and 584 because update was returning false. Had to change the asserts from expecting true, to expecting false. Below is the explanation I found for it. Please let me know if that is correct or if my changes have broken something.


After 10 events at interval 10, the smoothed interval has converged to ~10. Then update(200) arrives with dt = 110 (200 - 90, last event was at timestamp 90).

The liveness check inside update is dt <= smoothed * deadline_multiple, which is 110 <= 10 * 3 = 30 — false. The gap of 110 ticks is more than 3× the smoothed interval, so the detector correctly reports the source as dead at that moment.

The next update(210) has dt = 10, and by then the EMA has partially absorbed the large gap (smoothed rises to ~22), so 10 <= 22 * 3 = 66 — true, alive again.

ChronoCoders:
CI is green. Your explanation on the test changes is correct. Ready for @Abso1ut3Zer0.

Abso1ut3Zer0 (approved):
Thanks for taking this on! I pushed a small follow-up commit adjusting the markdown docs to match the new types — `LivenessF64` references swapped to `LivenessI64`/`LivenessU64`, the long-removed `F32`/`I32` variants dropped from the type list, and the builder examples updated to integer timestamps. The stats change itself looks good. Approving.

Issue #519, opened 2026-06-14
nexus-fix-engine: FIX interop conformance harness (containerized QuickFIX)


Abso1ut3Zer0:
Our equivalent of the Autobahn suite for `nexus-web`, but for the FIX session layer: a containerized **reference FIX engine (Python QuickFIX)** as the peer, our engine on the other side, driving a battery of exchanges and asserting conformance.

## Coverage
- Logon / Logout handshake, `NextExpectedMsgSeqNum`
- Heartbeat / TestRequest
- Sequence gap → ResendRequest → replay
- Gap-fill of admin messages (see the resend issue) + PossDup / OrigSendingTime
- SequenceReset (GapFill and Reset modes)
- Malformed / out-of-order input rejection

## Notes
- Podman-based, following the existing Autobahn setup precedent in the repo.
- Should run against both the sync transport and (later) the tokio adapter.

## Acceptance
- A repeatable suite (ideally CI-runnable) that exercises the full session lifecycle against QuickFIX and asserts our behavior matches.

PR #539, opened 2026-06-18
fix(fix-engine): transport follow-ups (#537)

### LINKED ISSUE #537: nexus-fix-engine transport: follow-ups (#517/#536)
Deferred items from the synchronous TCP transport (#517 / #536). The transport itself is done and correct; these are polish / known-limitation follow-ups.

### Resendable message size is capped by the writer
`do_resend` builds each reframed replay *into* the `FrameWriter`, so a replay larger than the writer capacity (default 64 KB) errors with `FrameTooLarge` — whereas `send_app` can direct-write an oversized *send*. So "can send" and "can resend" have different ceilings.
- [ ] Decide: document the cap (and make the capacity tunable for venues with large RawData/XML), or make `reframe_app` handle oversized (reframe into a temporary growable buffer / stream the field copy) so any sendable message is also resendable.

### Add a `blocking_session.rs` example
The blocking loop was deliberately kept out of the library (the transport is caller-driven `recv`/`send_app`/`wants_read`). Users need a worked recipe.
- [ ] `examples/blocking_session.rs` driving the `recv` loop on a blocking socket (initiator + acceptor).

### Inbound dispatch re-scans the frame per tag
`dispatch` calls `find_tag` repeatedly (49, 56, 34, 43, 35, then type-specific tags), each a full scan from offset 0.
- [ ] Parse the header once (single pass) instead of N `find_tag` scans. Perf nit, cold-ish path.

### Reactor / non-blocking writes (dependency note, not a task here)
The sync transport's write side (`flush_to`, `write_through`) is blocking-drain by design, matching nexus-web's `Client`. True non-blocking writes for a reactor (buffer + single non-blocking flush, `wants_write`-driven re-flush) land with the async adapter (#521) and multi-session (#518), via `WireStream` — not the sync transport.

Part of the FIX engine v1 buildout (#523). Transport ticket: #517.

ChronoCoders:
Closes #537.

Three items:

- Resendable cap: when reframe_app fails even after flushing (frame exceeds writer capacity), falls back to a heap buffer and writes through directly. Makes "can send" == "can resend" for oversized messages.
- Single-pass dispatch: replace five separate find_tag scans per inbound message with one FieldReader pass that collects all header and type-specific fields.
- examples/blocking_session.rs: minimal initiator + acceptor recipe driving the recv loop on a blocking socket.

### DIFF
--- a/nexus-fix-engine/examples/blocking_session.rs
+++ b/nexus-fix-engine/examples/blocking_session.rs
@@ -1,0 +1,129 @@
+#![cfg(unix)]
+
+//! Blocking session recipe: one initiator connects to one acceptor on localhost,
+//! sends a NewOrder, then logs out.
+//!
+//! Run with: cargo run --example blocking_session
+
+use std::net::TcpListener;
+use std::path::PathBuf;
+use std::time::{Duration, Instant};
+
+use nexus_fix_codec::{FrameFormatter, encode_fix_uint};
+use nexus_fix_engine::{
+    CompId, FixConnection, FixJournal, SessionConfig, SessionState, State,
+};
+
+const BEGIN: &[u8] = b"FIX.4.4";
+
+fn main() {
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let addr = listener.local_addr().unwrap();
+
+    let acceptor_dir = tmp_dir("acceptor");
+    let acceptor = std::thread::spawn(move || run_acceptor(listener, acceptor_dir));
+
+    let initiator_dir = tmp_dir("initiator");
+    run_initiator(addr, initiator_dir);
+
+    acceptor.join().unwrap();
+}
+
+fn run_acceptor(listener: TcpListener, dir: PathBuf) {
+    let (stream, _) = listener.accept().unwrap();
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
LGTM — all three #537 follow-ups, verified:

- **Resendable cap** — `do_resend`'s oversized branch reframes into a heap `tmp` (`orig.len()+512`, comfortable headroom) and `write_through`s it directly after flushing, so an oversized message is as resendable as it is sendable.
- **Single-pass dispatch** — `parse_frame_header` collects every field the old five `find_tag` scans used (35/34/49/56/43/108/141/112/7/16/45/36/123), same defaults, `None` on missing `35`/`34`. One pass instead of five.
- **`examples/blocking_session.rs`** — clean initiator+acceptor recipe driving the `recv` loop.

Diff is clean against main now that #536 landed (transport + example only). Tests pass (43+9+24+3+2), clippy clean, example builds.

One negligible note (not worth changing): dispatch is now parse-then-CompId-check, so a frame that's both wrong-CompId *and* missing seq/MsgType is skipped rather than triggering a disconnect — an extremely narrow edge the conformance harness (#519) would catch if it ever mattered.

PR #540, opened 2026-06-20
feat(fix-engine): FIX conformance harness (#519)

### LINKED ISSUE #519: nexus-fix-engine: FIX interop conformance harness (containerized QuickFIX)

ChronoCoders:
Closes #519.

Python FIX 4.4 peer (`tests/fixtures/fix_peer.py`, pure stdlib) runs as an acceptor subprocess. Rust integration test (`tests/fix_conformance.rs`) connects our `FixConnection` as the initiator and drives five scenarios:

- `logon_logout` — basic session setup and teardown
- `heartbeat` — peer sends TestRequest, engine replies with matching Heartbeat
- `resend` — peer sends ResendRequest for an app message, engine replays with PossDup=Y and OrigSendingTime
- `gap_fill` — peer sends SequenceReset GapFill, engine advances expected seq and accepts Logout at new seq
- `seq_reset` — peer sends SequenceReset Reset, engine jumps expected seq and accepts Logout at new seq

New `verify-fix` CI job: Python 3.12, no extra deps (stdlib only), runs `cargo test -p nexus-fix-engine --test fix_conformance`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,6 +105,17 @@ jobs:
     - name: Verify Rust matches LightGBM
       run: cargo test -p nexus-inference --features loader-lightgbm --test lightgbm_integration
 
+  verify-fix:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@1.94.0
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+    - name: FIX conformance tests (Python peer + Rust engine)
+      run: cargo test -p nexus-fix-engine --test fix_conformance
+
   loom:
     runs-on: ubuntu-latest
     steps:
--- a/nexus-fix-engine/tests/fix_conformance.rs
+++ b/nexus-fix-engine/tests/fix_conformance.rs
@@ -37,7 +37,9 @@ fn spawn_peer(scenario: &str) -> (std::process::Child, u16) {
 
 fn connect(port: u16, dir: &PathBuf) -> FixConnection<TcpStream> {
     let stream = TcpStream::connect(("127.0.0.1", port)).unwrap();
-    stream.set_read_timeout(Some(Duration::from_secs(10))).unwrap();
+    stream
+        .set_read_timeout(Some(Duration::from_secs(10)))
+        .unwrap();
     FixConnection::from_parts(
         stream,
         SessionState::new(Duration::from_secs(30)),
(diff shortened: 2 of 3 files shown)


Abso1ut3Zer0 (approved):
Approving — this is a genuinely useful harness. Driving the real `FixConnection` initiator path against an independent stdlib Python acceptor over TCP is a much stronger bar than the in-crate unit tests, and the ephemeral-port-over-stdout handshake keeps it flake-free. I ran all five scenarios locally (Python 3.14) and confirmed they aren't vacuous: making the resend scenario's `35=4` branch fatal still passes, which proves the engine really replays the app message with `43=Y` + `122` rather than gap-filling.

A few cleanup items for a follow-up (none blocking):

1. **Tighten the resend assertion** (`fix_peer.py:103`). For `begin=end=2` — a single known app message — the intent is to verify a *replay*, but the `if msg.get("35") == "4": break` escape also accepts a gap-fill, so a future regression (engine gap-filling an app message) would pass silently. Dropping the escape for this scenario makes it require `43=Y`.

2. **Kill the peer on a Rust-side panic.** `Child` isn't killed on drop, so a panicking test leaves the python peer running until its 10s socket timeout. A small kill-on-drop guard (or `child.kill()` on teardown) makes failures cleaner.

3. **Add messages to the bare asserts** in `gap_fill`/`seq_reset` (`assert msg.get("35") == "5"`) so a failure is diagnosable.

Broader adversarial/negative conformance — unbounded ResendRequest end-seq, backward/zero SequenceReset, bad checksum, garbage framing — is the natural next step. I'm folding that into a consolidated FIX-engine audit so it lands as a tracked follow-up rather than scope-creeping this PR.

Issue #544, opened 2026-06-20
nexus-fix-engine: adopt nexus-net wire seam; thin async adapter


Abso1ut3Zer0:
**Decided in the 2026-06-20 design discussion.** Build the FIX transport on the `nexus-net` wire seam, the way `nexus-web`/`nexus-async-web` already compose, and make the async adapter thin.

**Depends on the reader/writer re-layer issue** (the `MessageReader<D>`/`MessageWriter<D>` work) — this builds on top of it.

## Why
`nexus-web` owns the codec + connection core (`Client<S>`) once; `nexus-async-web` reuses it and only swaps the IO driver via a generic `AsyncReadAdapter<S>` (`WireStream`), copying no protocol/encoding code, and `WireStream::poll_fill_into` reads straight into the parser's buffer (no intermediate copy).

The FIX engine never adopted this: `FrameReader` already impls `nexus_net::ParserSink`, but neither `FixConnection` nor the async adapter (#541) references `WireStream` — both hand-roll their IO loop. #541 additionally copies the P&L-critical encode helpers (timestamp / civil-date / reframe / gap-fill) from the engine's private modules and reintroduces an 8 KiB intermediate read buffer.

## Target
- `FixConnection<S: WireStream, D>` over the `MessageReader<D>` / `MessageWriter<D>` layer; `poll_fill_into` fills `MessageReader` copy-free.
- Encode helpers live in the engine (shared), not per-crate copies.
- Promote `nexus-async-web`'s `AsyncReadAdapter<S>` into `nexus-net` behind a tokio feature so web and FIX share it.
- `nexus-async-fix-engine` becomes a thin `WireStream` adapter + timer wiring only.

## Effect on #541
This **reworks #541's internals** (which merged as the reference async adapter). Its conformance tests guard the behavior through the refactor.

## Acceptance
Async crate contains no protocol/encode code; one IO driver; conformance green sync + async; no per-message `to_vec` or 8 KiB intermediate on the inbound path. Relates to #521. Part of #523.

Abso1ut3Zer0:
Depends on #542 (the `MessageReader<D>`/`MessageWriter<D>` re-layer) — this builds the transport on the wire seam on top of that layering.

Abso1ut3Zer0:
Unblocked — #542 landed (#548 merged), so the typed `MessageReader<D>`/`MessageWriter<D>` layering this builds on is in place. Ready to pick up.

Current state to rework from:
[...]

PR #548, opened 2026-06-21
feat(fix-engine): typed MessageReader/MessageWriter; delete Session<D>

### LINKED ISSUE #542: nexus-fix-engine: re-layer typed reader/writer (MessageReader<D>/MessageWriter<D>); delete Session<D>
**Decided in the 2026-06-20 design discussion.** Re-layer the FIX reader/writer to mirror `nexus-web`, and delete `Session<D>`.

## Problem
`Session<D>::on_message` (`framework.rs`) calls each `SessionState` handler but **discards the returned `Out`** — the only channel for outbound admin replies — so as an acceptor it never emits a Logon ack, Heartbeat echo, ResendRequest-on-gap, or Logout. It conflates two concerns (decode + protocol) in one wrapper, and the README documents a third API that exists on neither `Session<D>` nor `FixConnection`.

## Target layering (mirrors nexus-web's `FrameReader` → `Client::recv()`)
- **`SessionState`** — protocol core, no `D`. Unchanged.
- **Low-level byte primitives:** `FrameReader` / `FrameWriter` — framing (`8=…10=` + checksum) / outbound byte buffer. Dictionary-free; reused by the journal and codec.
- **User-level typed pair (new):** `MessageReader<D>` / `MessageWriter<D>`, built on the byte primitives, symmetric.
  - `MessageReader<D>::next() -> Option<Message<'_, D>>` — the single reader the user touches (no more "frame, then decode separately").
  - `MessageWriter<D>` — encode a typed message → frame bytes (standard header + framing + checksum; body from user/codegen).
  - `MessageReader<D>` impls `nexus_net::ParserSink` by delegating `spare()`/`filled()` to its inner `FrameReader`, so the wire-seam refactor can fill it with no extra copy.
- **Transport** holds `MessageReader<D>` + `MessageWriter<D>` + `SessionState`; on each frame it **matches the typed `Message` to drive `SessionState`** (the variant *is* the dispatch — removes the hand-rolled header re-parse), auto-flushes `Out`, and returns the `Message`. `recv() -> Option<Message<'_, D>>`, mirroring `Client::recv()`.
- **Delete `Session<D>`** — its decode half becomes `MessageReader<D>`, its protocol half is just `SessionState`. BYO-IO callers compose the two.

## Note
FIX diverges from WS deliberately: session mechanics (seq check, gap-fill, TestRequest reply, logout handshake) are **auto-handled** (Out flushed by the transport); the surfaced message is *additive* for app logic, not "you must reply."

## Acceptance
- `MessageReader<D>`/`MessageWriter<D>` exist, symmetric; `FrameReader`/`FrameWriter` remain as byte primitives.
- One dispatch path (via the typed `Message`); `Session<D>` removed.
- Outbound replies actually emitted; tests assert them.
- README rewritten to the real API.

Naming (`MessageReader`/`MessageWriter`) is tentative. Paired with the wire-seam refactor (filed alongside) — that PR builds the transport on `nexus_net::WireStream` on top of this layering. Part of #523.

ChronoCoders:
Fixes #542.

**Problem:** `Session<D>::on_message` called every `SessionState` handler but discarded the returned `Out`, so the engine never sent Logon ack, Heartbeat echo, ResendRequest-on-gap, or Logout as acceptor.

**Changes:**

- `MessageReader<D>` — wraps `FrameReader` + owned frame buffer; implements `ParserSink`; dictionary-aware via `D::Header`
- `MessageWriter<D>` — wraps `FrameWriter`; `encode_admin` uses `D::BEGIN_STRING` (no stored `begin_string` field)
- `FixConnection<S, D>` — generic over dictionary; `begin_string` field removed; `from_parts` / builder `connect` / `accept` drop the `begin_string` arg
- `recv(&mut self, now) -> Result<Option<Message<'_, D>>, Error>` — returns typed message directly; `Out` auto-flushed on every frame
- `do_resend` extracted as free function (split-borrow safe while frame slice is live)
- `Error::Protocol(SessionError)` variant for decode failures
- `Session<D>` deleted
- Tests rewritten; example updated

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -116,6 +116,17 @@ jobs:
     - name: FIX conformance tests (Python peer + Rust engine)
       run: cargo test -p nexus-fix-engine --test fix_conformance
 
+  verify-async-fix:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@1.94.0
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+    - name: Async FIX conformance tests (Python peer + tokio adapter)
+      run: cargo test -p nexus-async-fix-engine --test async_conformance
+
   loom:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ members = [
 	"nexus-fix-codegen",
 	"nexus-fix-codegen-tests",
 	"nexus-fix-engine",
+	"nexus-async-fix-engine",
 	"nexus-platform",
 ]
 
--- a/nexus-async-fix-engine/Cargo.toml
+++ b/nexus-async-fix-engine/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "nexus-async-fix-engine"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+description = "Tokio async adapter for the sans-IO FIX session layer"
+repository.workspace = true
+keywords = ["fix", "protocol", "session", "async", "tokio"]
+categories = ["network-programming", "finance", "asynchronous"]
+
+[lints]
+workspace = true
+
+[dependencies]
+nexus-fix-engine = { path = "../nexus-fix-engine" }
+nexus-fix-codec = { path = "../nexus-fix-codec" }
+tokio = { version = "1", features = ["io-util", "net", "rt", "time"] }
+
+[dev-dependencies]
+tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -1,0 +1,601 @@
+//! Tokio async adapter for the sans-IO FIX session layer.
+//!
+//! [`AsyncFixConnection`] drives [`SessionState`] over a tokio `AsyncRead +
+//! AsyncWrite` stream, mirroring [`FixConnection`](nexus_fix_engine::FixConnection)
+//! with `.await` on socket I/O and `tokio::time` for heartbeat/TestRequest timers.
+
+#![cfg(unix)]
+
+use std::io;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+use nexus_fix_codec::{
+    FieldReader, FrameFormatter, encode_fix_uint, find_tag, parse_fix_bool, parse_fix_seqnum,
+    parse_fix_uint,
+};
+use nexus_fix_engine::{
+    AdminMsg, CompId, DisconnectReason, Event, FixJournal, FrameError, FrameReader, FrameWriter,
+    Out, ReplayItem, SessionConfig, SessionState, State,
+};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use tokio::time::timeout_at;
+
+const TS_LEN: usize = 21;
+
+/// Error from [`AsyncFixConnection`].
+#[derive(Debug)]
+pub enum Error {
+    Io(io::Error),
+    FrameTooLarge(usize),
+}
+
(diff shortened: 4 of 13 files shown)


Abso1ut3Zer0 (changes requested):
Really nice work — this faithfully implements #542: typed `recv() -> Message<'_, D>`, `Out` auto-flushed every frame, `do_resend` extracted cleanly with split borrows, `MessageReader/Writer<D>` layered on the byte primitives via `ParserSink`, and `Session<D>` gone. The disjoint-field borrow that lets `recv` flush before returning the borrowed `Message` is handled correctly, and the reused inbound `Vec` is a nice improvement over the old per-message `to_vec`. One correctness issue to fix before merge, plus a couple of small things.

### 🔴 Out-of-sequence app messages are surfaced as `Message::Application`

In the app branch (`transport.rs:448-462`), `recv` calls `on_app`, flushes the `Out`, checks only for `Disconnected`, then returns `Message::Application` **unconditionally**. But on a sequence **gap** (seq > expected), `on_app` queues a `ResendRequest` and returns with **no `Event::App`** — the session has rejected that message and asked for retransmission. The caller still receives `Message::Application` and would act on an out-of-order message (and again when it's resent). For an order/fill stream that's a real hazard.

Both prior paths gated this on `Event::App` — and the async adapter still does, at `nexus-async-fix-engine/src/lib.rs:270` (`Some(Event::App { .. }) if is_app => ...`). Collapsing dispatch into `recv()` dropped the gate on the sync side only.

**Fix:** surface `Message::Application` only when the session accepted the message, and otherwise move on to the next frame rather than returning it. Concretely, fold the dispatch into the read loop and, in the app branch:

```rust
_ => {
    let out = self.state.on_app(seq, poss_dup, now);
    // … encode admin + flush …
    if let Some(Event::Disconnected { reason }) = out.event() {
        return Ok(Some(Message::Disconnected { reason }));
    }
    if matches!(out.event(), Some(Event::App { .. })) {
        return Ok(Some(Message::Application { header: D::Header::decode(frame) }));
    }
    continue; // gap → ResendRequest already sent; don't surface, process next frame
}
```

(Admin branches should keep surfacing unconditionally — that's intended, additive info for the caller.)

**Test:** none of the conformance scenarios exercise an inbound gap, so this slipped through. Could you add one — e.g. a Rust integration test over an in-memory duplex stream: log on, feed an app message with a seq gap (expected 2, send 5), then assert (a) a `ResendRequest` was written and (b) `recv` did **not** yield `Message::Application` for seq 5. A `gap_fill`-style conformance scenario works too.

### Minor

- `transport.rs:347-352` (Logout→disconnect): decodes `D::Logout` then `let _ = msg;` and discards it — pointless work, and it turns a decode failure into an error on a disconnect path. Just return `Message::Disconnected { reason }`.
- `transport.rs:309`: a frame missing tag 35 falls through to the app branch (`raw_type = b""`) rather than `SessionError::MissingMsgType`. Worth erroring explicitly for parity with the missing-34 handling just above.

The encode/timestamp/reframe duplication across `framework.rs`/`transport.rs` and the async crate is expected — that's #544's job, out of scope here. Thanks again, this is close.

ChronoCoders:
All three fixed:

- **Gap gate**: app branch now checks `Event::App` before returning `Message::Application`; gaps return `Ok(None)` so the caller loops to the next frame.
- **Logout-disconnect decode**: removed the pointless `D::Logout::decode` + `let _ = msg` on the disconnect path.
- **Missing tag 35**: now returns `Protocol(MissingMsgType)` instead of falling to the app branch.

New transport test: `inbound_gap_sends_resend_request_and_suppresses_app_message` — sends a gap app (seq 5, expected 2), asserts no `Message::Application` and state == `Resending` (confirms ResendRequest was queued).

Abso1ut3Zer0 (approved):
Re-reviewed — all three items addressed cleanly. The gap gate now suppresses out-of-sequence app messages (returns `Ok(None)`), the Logout-disconnect path returns `Disconnected` directly, and missing tag 35 surfaces `Protocol(MissingMsgType)`. The new `inbound_gap_sends_resend_request_and_suppresses_app_message` test guards the gap behavior well (asserts both no surfaced app and `Resending` state). Nice work — approving.

Issue #546, opened 2026-06-20
nexus-fix-codec: parser/encoder overflow hardening (parse_tag, decimal encode)


Abso1ut3Zer0:
Codec parser/encoder overflow hardening from the 2026-06 audit.

- **`parse_tag` overflow** (`nexus-fix-codec/src/reader.rs`): `tag = tag * 10 + d` is unchecked on peer bytes → debug panic / release wrap for tags with >10 digits. Memory-safe, but wrong. Cap to a sane digit count (FIX tags are ≤ 5 digits) or use `saturating_*`.
- **Decimal encode overflow — verify before changing.** A sub-review flagged that `FixDecimal` uses `10u64.pow(scale)` and i64/i128 rescales with unchecked `.pow()`, which could silently corrupt a serialized price for large scale/mantissa. This is P&L-critical, so **gate any change on a `proptest`/`cargo fuzz` round-trip** over encode ↔ parse first; if confirmed, switch to `checked_pow` + a scale bound. Not asserted as a bug yet — verification step is part of this issue.

Part of #523.

On0n0k1:
About `Decimal encode overflow`, I found that there are unchecked powers in `/nexus-fix/codec/types.rs` on lines `138`, `1557` and `1560`. But `FixDecimal::parse` has a check for overflows on lines 77 and 80. If the type is created with the "parse" function, the check will not be necessary.

The properties of `FixDecimal` are public, meaning it is possible to create an instance with invalid attributes like `FixDecimal { mantissa: 20, scale: 255 }`. How about changing the properties to be public only inside the crate? That way a library user will have to use `parse` to get this type.

```
pub struct FixDecimal {
     pub(crate) mantissa: i64,
     pub(crate) scale: u8,
}
```

On0n0k1:
I created a Pull Request with only the change for `parse_tag overflow`. Waiting for your feedback.

Abso1ut3Zer0:
## Verification (the "verify before changing" step) — E13 is real

Read the code; the decimal-encode overflow is confirmed, not speculative. Three spots, all unchecked `.pow()`:

[...]

Issue #543, opened 2026-06-20
nexus-fix-engine: session/transport hardening + adversarial conformance


Abso1ut3Zer0:
Peer-facing hardening gaps found in the 2026-06 audit. **Each sub-item ships with a red→green adversarial scenario** added to the conformance harness (#540 + the async equivalent), which today covers only positive / forward-direction flows.

## Items
- **Resend hang / u32 overflow.** `persist.rs` `resend()` takes `high = end` unclamped; `EndSeqNo` (tag 16) is peer-controlled. A huge `end` makes `ResendIter` loop ~4B times and `self.seq += 1` overflows at `u32::MAX` → debug panic / **release infinite loop**. Fix: `let high = end_clamped.min(self.next_outbound.saturating_sub(1));` (also resolves `begin > end` / over-range to an empty iterator). Test: `ResendRequest(7=1, 16=4000000000)`.
- **No inbound checksum validation.** `frame.rs` frames purely on peer `BodyLength`; the `10=XXX` trailer is never validated and the transport never checks tag 10, so a wrong `BodyLength` desyncs the stream silently. Fix: validate trailer format + value at the frame boundary (the codec already has a checksum helper). Test: corrupt checksum / wrong BodyLength.
- **Silent protocol-violation swallow.** The transport drops `Garbage` frames (`{}`) and silently discards frames missing tag 34/35 (`Ok(None)`); a misbehaving peer is invisible at 3am. Fix: a reject/garbage counter or surfaced event. Test: garbage stream, missing-34 frame.
- **SequenceReset Reset-mode unvalidated.** `session/mod.rs` sets `next_inbound = new_seq` with no guard; peer `SequenceReset(123=N, 36=0)` zeros inbound and `36 < expected` rewinds it (the GapFill branch already guards). Fix: `new_seq != 0 && new_seq >= next_inbound`. Test: backward / zero reset.

## Acceptance
Each fix + its adversarial scenario green in **both** the sync (#540) and async conformance harnesses. Part of #523.

PR #549, opened 2026-06-22
fix(fix-codec): guard parse_tag against overflow on malformed peer bytes

On0n0k1:
Related to #546

This PR is associated to  `parse_tag overflow` in the issue.

### DIFF
--- a/nexus-fix-codec/src/reader.rs
+++ b/nexus-fix-codec/src/reader.rs
@@ -239,9 +239,10 @@ pub fn parse_tag(buf: &[u8]) -> (u32, usize) {
     let mut tag = 0u32;
     let mut i = 0;
     while i < buf.len() && buf[i] >= b'0' && buf[i] <= b'9' {
-        tag = tag
-            .saturating_mul(10)
-            .saturating_add((buf[i] - b'0') as u32);
+        if i >= 5 {
+            return (u32::MAX, i);
+        }
+        tag = tag * 10 + (buf[i] - b'0') as u32;
         i += 1;
     }
     (tag, i)
--- a/nexus-fix-codec/src/reader.rs
+++ b/nexus-fix-codec/src/reader.rs
@@ -713,6 +713,21 @@ mod tests {
     }
 
     #[test]
+    fn parse_tag_overflow() {
+        assert_eq!(parse_tag(b"99999999999=x"), (u32::MAX, 5));
+    }
+
+    #[test]
+    fn overflowed_tag_dropped_by_field_reader() {
+        let msg = b"8=FIX.4.4\x013400000=1\x0135=A\x01";
+        assert!(find_tag(msg, 0, 3_400_000).is_none());
+        // Also none because iterator returns infinite for each tag after the large one
+        assert!(find_tag(msg, 0, 35).is_none());
+        // Since 8 is found before the large tag, it successfully retrieves the value
+        assert!(find_tag(msg, 0, 8).is_some());
+    }
+
+    #[test]
     fn swar_mask_conversion() {
         // Single match at byte 0: bit 7 set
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0:
Thanks for the fix! Saturating to `u32::MAX` is the right move — it can't collide with a real tag (1–99999), so a garbage tag just becomes an ignored unknown field instead of wrapping onto a real one.

Couple of tests to lock it in and then this is good to go:
1. A `parse_tag` unit test for the overflow — e.g. `assert_eq!(parse_tag(b"99999999999=x"), (u32::MAX, 11));`
2. One showing the error still propagates: a frame with an overflowed required tag (say `34`) should surface `Err(Error::Protocol(SessionError::MissingMsgSeqNum))` from `recv()`.

ChronoCoders (commented):
The saturating fix prevents the panic but still consumes all digits of a runaway field — a 20-digit tag still advances `i` to 20 before returning `u32::MAX`. Since FIX tags are at most 5 digits, a digit-count cap is tighter and avoids producing a sentinel tag value entirely:

\`\`\`rust
while i < buf.len() && buf[i] >= b'0' && buf[i] <= b'9' {
    if i >= 5 {
        return (u32::MAX, i);
    }
    tag = tag * 10 + (buf[i] - b'0') as u32;
    i += 1;
}
\`\`\`

Same `u32::MAX` sentinel for callers, same "unknown tag, skip it" behavior downstream — but the unchecked arithmetic is preserved on the hot path for all valid inputs, and malformed fields are rejected at the earliest point.

On0n0k1:
> The saturating fix prevents the panic but still consumes all digits of a runaway field — a 20-digit tag still advances `i` to 20 before returning `u32::MAX`. Since FIX tags are at most 5 digits, a digit-count cap is tighter and avoids producing a sentinel tag value entirely:
> 
> ```rust while i < buf.len() && buf[i] >= b'0' && buf[i] <= b'9' { if i >= 5 { return (u32::MAX, i); } tag = tag * 10 + (buf[i] - b'0') as u32; i += 1; } ```
> 
> Same `u32::MAX` sentinel for callers, same "unknown tag, skip it" behavior downstream — but the unchecked arithmetic is preserved on the hot path for all valid inputs, and malformed fields are rejected at the earliest point.

If we are checking the index value during the loop, we can just check the length of the collection before starting the loop. As in the following:


```
    if buf.len() >= 5 {
        // The loop will always return i as 5 in that case
        return (u32::MAX, 5);
    } 
    while i < buf.len() && buf[i] >= b'0' && buf[i] <= b'9' {
        tag = tag * 10 + (buf[i] - b'0') as u32;
        i += 1;
    }
```

While checking the index during the loop could trigger compiler optimizations. I am not certain if it does unless I implement a benchmark testing it locally first.

On0n0k1:
Hello. I added 2 unit tests for your latest suggestion. As can be seen from the unit tests, if one of the tags is large, the tags before it can be found. Every tag after the large tag will return u32::MAX though. Just confirming if this behavior is acceptable.

Abso1ut3Zer0:
Did a bit of digging into how production engines (QuickFIX, Artio) handle this. The pattern is: parse the whole message and bubble the offending tag up to the session layer — a Reject carrying the bad tag id + a reason code — rather than letting one malformed field break the rest of the parse. That lines up with the saturating approach: consume all the digits, return the `u32::MAX` sentinel, and keep going, so the rest of the message stays parseable.

The `i >= 5` early-return is the opposite — it stops mid-tag, so `FieldReader` halts and everything after the bad field disappears (your test shows `35` going missing). So let's go back to the saturating version. The "reject the bad tag with its id" behavior is a session-layer concern — I'll track that under #543.

On0n0k1:
Understood. Since the requirements changed, I am closing this PR.

Issue #547, opened 2026-06-20
nexus-fix-engine: conformance harness cleanups (#540 follow-ups)


Abso1ut3Zer0:
Follow-up cleanups for the conformance harness landed in #540.

- **Tighten the resend assertion** (`tests/fixtures/fix_peer.py`). For `begin = end = 2` — a single known app message — the scenario should require a *replay*, but the `if msg.get("35") == "4": break` escape also accepts a gap-fill, so a future regression (engine gap-filling an app message) would pass silently. Drop the escape for this scenario so it requires `43=Y`.
- **Kill the peer on a Rust-side panic.** `Child` isn't killed on drop, so a panicking test leaves the Python peer running until its 10s socket timeout. Add a kill-on-drop guard / teardown `child.kill()`.
- **Add messages to the bare asserts** in `gap_fill` / `seq_reset` (`assert msg.get("35") == "5"`) so a failure is diagnosable.

Small; one PR. Follows up #540.

PR #550, opened 2026-06-23
fix(fix-engine): session/transport hardening (#543)

### LINKED ISSUE #543: nexus-fix-engine: session/transport hardening + adversarial conformance

ChronoCoders:
Closes #543.

Four peer-facing hardening fixes from the 2026-06 audit, each with an adversarial test.

**1. Resend hang / u32 overflow** (`persist.rs`)
Clamp `end` to `next_outbound.saturating_sub(1)` in `resend()`. A peer sending `EndSeqNo=4000000000` no longer causes a ~4B-iteration loop.

**2. Inbound checksum validation** (`frame.rs`)
Call `validate_checksum` at the frame boundary in `try_parse()`. Frames with a wrong checksum are now surfaced as `FrameError::Garbage` instead of being silently accepted.

**3. Garbage frame counter** (`transport.rs`)
Add `garbage_frames: u64` field, incremented on every `FrameError::Garbage`. Exposed via `garbage_frame_count()`. Misbehaving peers are now observable without log scraping.

**4. SequenceReset Reset-mode guard** (`session/mod.rs`)
Reject `new_seq == 0 || new_seq < next_inbound` before applying a Reset-mode SequenceReset. Prevents peers from rewinding or zeroing the inbound sequence counter.

**Tests added:** unit tests in `frame.rs` and `persist.rs`; four adversarial integration tests in `tests/transport.rs` covering each fix.

### DIFF
--- a/nexus-fix-engine/src/frame.rs
+++ b/nexus-fix-engine/src/frame.rs
@@ -6,6 +6,7 @@
 //! tags `8=` (BeginString), `9=` (BodyLength), and `10=` (CheckSum) that
 //! are invariant across all FIX versions.
 
+use nexus_fix_codec::validate_checksum;
 use nexus_net::buf::{ReadBuf, WriteBuf};
 use nexus_net::wire::ParserSink;
 
@@ -280,6 +281,10 @@ impl FrameReader {
 
         if data.len() < message_end {
             return ParseResult::Incomplete;
+        }
+
+        if validate_checksum(&data[..message_end]).is_err() {
+            return ParseResult::Garbage;
         }
 
         ParseResult::Complete(message_end)
@@ -979,5 +984,30 @@ mod tests {
         // data() should start right at the message — no gap.
         assert!(writer.data().starts_with(b"8=FIX.4.4\x01"));
         assert_eq!(writer.data().len(), len);
+    }
+
+    // ---- checksum validation ----
+
+    #[test]
+    fn corrupt_checksum_rejected() {
+        let mut msg = heartbeat();
+        let n = msg.len();
+        msg[n - 2] ^= 1; // flip the last checksum digit
+
+        let mut reader = FrameReader::builder().build();
+        reader.read(&msg).unwrap();
+        let err = reader.next().unwrap_err();
+        assert!(
+            matches!(err, FrameError::Garbage { .. }),
+            "corrupt checksum must be treated as garbage"
+        );
+    }
+
+    #[test]
+    fn valid_checksum_accepted() {
+        let msg = heartbeat();
+        let mut reader = FrameReader::builder().build();
+        reader.read(&msg).unwrap();
+        assert!(reader.next().unwrap().is_some());
     }
 }
--- a/nexus-fix-engine/src/frame.rs
+++ b/nexus-fix-engine/src/frame.rs
@@ -90,7 +90,7 @@ impl std::error::Error for FrameError {}
 ///     .build();
 ///
 /// // A complete Heartbeat message.
-/// let msg = b"8=FIX.4.4\x019=5\x0135=0\x0110=162\x01";
+/// let msg = b"8=FIX.4.4\x019=5\x0135=0\x0110=163\x01";
 /// reader.read(msg).unwrap();
 ///
 /// let frame = reader.next().unwrap().unwrap();
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -50,7 +50,7 @@ impl<'a> Iterator for ResendIter<'a> {
                 });
             }
             let seq = self.seq;
-            self.seq += 1;
+            self.seq = self.seq.saturating_add(1);
             let is_gap = match self.journal.resend_one(seq) {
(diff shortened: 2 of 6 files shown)


Abso1ut3Zer0:
Nice — three of the four are solid. The checksum fix even caught a latent bug: the doc example had a wrong checksum (`10=162`) that the old code was silently accepting, now corrected to `10=163`. Garbage counter and the SequenceReset guard both look right.

One real issue in the resend clamp (Fix 1) before this is good to go:

```rust
let re = if re == 0 {
    self.state.next_outbound_seq().saturating_sub(1)   // ✓ last sent
} else {
    re.min(self.state.next_outbound_seq())             // ✗ missing the -1
};
```

The `re == 0` branch clamps to `next_outbound - 1` (last sent), but the bounded branch clamps to `next_outbound` — one past the last sent. So a peer requesting `end >= next_outbound` gets a resend that covers the **unsent** `next_outbound`, which gap-fills to `new_seq = next_outbound + 1` (the trailing gap emits `high + 1`). The engine then sends its next real message at `next_outbound`, which is now seqnum-too-low from the peer's side → disconnect. So a malicious large `EndSeqNo` still breaks the session, just a different way.

The test comment even says "clamp to next_outbound-1," but it only asserts the engine *responds* (no-hang), and the mock peer doesn't check sequence numbers, so the off-by-one slips through.

Fix is just the missing `-1`:
```rust
re.min(self.state.next_outbound_seq().saturating_sub(1))
```
Two suggestions while you're in there:
- Consider doing the clamp once inside `journal.resend()` (it already handles `end == 0`) so there's a single source of truth rather than clamping in both the transport and the journal.
- Tighten the test to assert the gap-fill's `new_seq` (should be `next_outbound`), not just that a frame came back — that's what would have caught this.

Non-blocking note on the SequenceReset guard: silently dropping the bad reset is safe (it recovers via the too-low → disconnect path), so good for now. The spec-strict response is a Reject (35=3) with `RefTagID=36` + reason — same "reject the bad input with its tag id" policy we flagged for the invalid-tag work; can come later.

ChronoCoders:
Good catch. Fixed both: clamped to `saturating_sub(1)` in both branches, and updated the test to request `begin=1,end=1` (within the engine's allocated range) so the iterator isn't empty. Also had to fix the existing `resend_request_triggers_gap_fill` test which was sending `begin=2,end=3` when the engine had only allocated seq=1 — the tighter clamp exposed that.

Abso1ut3Zer0 (approved):
Off-by-one is fixed correctly — both branches clamp to `next_outbound - 1` now — and the test change is the right call (the old `begin=2,end=3` was relying on the off-by-one to land in range; `begin=1,end=1` exercises a real in-range gap-fill). All four hardening fixes check out, CI green. Approving.

One non-blocking nice-to-have: the resend tests still assert `n > 0` rather than the gap-fill's `new_seq`, so they wouldn't actually catch the off-by-one if it came back (old code → `new_seq=3`, fixed → `new_seq=2`, both pass `n > 0`). Worth asserting `new_seq == 2` (the engine's `next_outbound`) in the huge-end test to lock the boundary in — can fold into the conformance cleanups (#547). Thanks for the quick turnaround.

ChronoCoders:
Done — added `assert_eq!(new_seq, Some(2))` on tag 36 of the gap-fill response. Would've caught the off-by-one.

PR #551, opened 2026-06-25
fix(fix-codec): harden FixDecimal against scale overflow

On0n0k1:
Related to #546 

This PR is associated with `Decimal encode overflow` subject in the issue.

**1. Private fields + validated constructor** (`types.rs`)
Make `mantissa` and `scale` private. Add `new()` enforcing `scale ≤ 19` — the ceiling where `10u64.pow(scale)` still fits in u64. Direct struct construction previously bypassed this, allowing `encode()` and `Display` to overflow silently in release or panic in debug.

**2. `debug_assert` in `encode()` and `Display`** (`types.rs`)
Add `debug_assert!(self.scale <= 19)` before the `10u64.pow(scale)` call in both. Catches any future regression in test builds at zero cost in release.

**3. `From<FixDecimal> for Decimal<i128, D>` → `TryFrom`** (`types.rs`)
The infallible `From` impl was unsound: `10_i128.pow(D - scale)` overflows at exponent ≥ 39 and the multiply can overflow too. Converted to `TryFrom` with `checked_pow` and `checked_mul`/`checked_div`.

**4. `checked_pow` in `TryFrom` for `Decimal<i64, D>` and `Decimal<i32, D>`** (`types.rs`)
`checked_mul` guarded the multiply but unchecked pow ran before it, making the guard ineffective for large exponents. Both branches now use `checked_pow`.

**Tests added:** updated `decimal_roundtrip` proptest in `property.rs`; added `decimal_roundtrip_reject_overflow` covering all mantissa values and all out-of-range scales (`20..=255`). Construction sites in `nexus-fix-codegen-tests` updated to `FixDecimal::new()`.

### DIFF
--- a/nexus-fix-codec/src/types.rs
+++ b/nexus-fix-codec/src/types.rs
@@ -1579,10 +1579,10 @@ mod decimal_conv {
 
             let scaled = if D >= d.scale {
                 d.mantissa
-                    .checked_mul(10_i64.pow((D - d.scale) as u32))
+                    .checked_mul(10_i64.checked_pow((D - d.scale) as u32).ok_or_else(err)?)
                     .ok_or_else(err)?
             } else {
-                d.mantissa / 10_i64.pow((d.scale - D) as u32)
+                d.mantissa / 10_i64.checked_pow((d.scale - D) as u32).ok_or_else(err)?
             };
             Ok(Self::from_raw(scaled))
         }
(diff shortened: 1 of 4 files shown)


On0n0k1:
`tests/cancel_alloc_count.rs` failed but it is unrelated to my changes. Running it locally with `cargo test -p nexus-async-rt --test cancel_alloc_count` didn't generate the same error. So please restart the Action.

Abso1ut3Zer0 (approved):
Approving — the overflow hardening is solid and well-tested: private fields + validated `new`, `debug_assert!(scale <= 19)` in `encode`/`Display`, `From<i128>` → `TryFrom`, `checked_pow` across all four conversion branches, and a thorough `decimal_roundtrip_reject_overflow` proptest (`scale in 20..=255`). CI green, no external breakage. Nice work.

A few non-blocking follow-ups (no need to touch this PR):
1. **Getters** — add `pub const fn mantissa(&self) -> i64` / `scale(&self) -> u8` now that the fields are private, matching `FixTime`'s const getters.
2. **Constructor ergonomics** — make `new` a `const fn -> Option<Self>` (mirrors `Decimal::from_scaled`, enables `const` construction), and add a `const fn new_unchecked` for hot-path sites where the scale is statically known good. The current `Result` `new` is fine to ship in the meantime.

Scope note: this is the decimal half of #546 — the `parse_tag` (E12) half is open again now that #549 is closed, so it'll want a fresh small PR (the saturating approach was the right one).

PR #552, opened 2026-06-25
fix(fix-codec): harden parse_tag against u32 overflow (#546)

ChronoCoders:
Closes part of #546.

- `reader.rs`: `tag * 10 + d` → `tag.saturating_mul(10).saturating_add(d)`; excess digits consumed so parsing continues past the bad field, u32::MAX sentinel is ignored by all valid tag lookups
- Added `parse_tag_overflow_saturates` unit test
- Added transport test: frame with overflowed tag 34 → `MissingMsgSeqNum`

### DIFF
--- a/nexus-fix-codec/src/reader.rs
+++ b/nexus-fix-codec/src/reader.rs
@@ -239,7 +239,7 @@ pub fn parse_tag(buf: &[u8]) -> (u32, usize) {
     let mut tag = 0u32;
     let mut i = 0;
     while i < buf.len() && buf[i] >= b'0' && buf[i] <= b'9' {
-        tag = tag * 10 + (buf[i] - b'0') as u32;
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
Approving — saturating `parse_tag` is the right approach (consume all digits, `u32::MAX` sentinel, parsing continues past the bad field — matching how QuickFIX/Artio handle it). Verified the callers bounds-check before indexing so the sentinel flows through as an unknown tag with no panic or halt. Both tests are solid: `parse_tag_overflow_saturates` proves saturate-and-consume-all (the anti-regression for the digit-cap), and the overflowed-tag-34 transport test confirms clean propagation to `MissingMsgSeqNum` with a valid checksum. Nice and tight.

PR #553, opened 2026-06-25
fix(fix-engine): conformance harness cleanups (#547)

### LINKED ISSUE #547: nexus-fix-engine: conformance harness cleanups (#540 follow-ups)

ChronoCoders:
Closes #547.

- `fix_peer.py/resend`: drop gap-fill escape for begin=end=2; require replay (43=Y)
- `fix_peer.py/gap_fill`, `seq_reset`: add failure messages to bare asserts
- `fix_conformance.rs`: `ChildGuard` kills child on drop so a Rust-side panic doesn't leave the Python peer running for 10s

### DIFF
--- a/nexus-fix-engine/tests/fix_conformance.rs
+++ b/nexus-fix-engine/tests/fix_conformance.rs
@@ -94,7 +94,21 @@ fn tmp_dir(suffix: &str) -> PathBuf {
     p
 }
 
-fn spawn_peer(scenario: &str) -> (std::process::Child, u16) {
+struct ChildGuard(std::process::Child);
+
+impl Drop for ChildGuard {
+    fn drop(&mut self) {
+        let _ = self.0.kill();
+    }
+}
+
+impl ChildGuard {
+    fn wait(&mut self) -> std::io::Result<std::process::ExitStatus> {
+        self.0.wait()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
Approving — clean close of #547. All three: the resend scenario now hard-asserts the PossDup replay (43=Y + 122) instead of accepting a gap-fill escape; `ChildGuard` kills the peer on drop so a Rust-side panic doesn't leave it running; and the bare asserts gained diagnostic messages. Tests-only, CI green. Thanks.

PR #554, opened 2026-06-25
refactor(fix-engine): fold recover() into open(); guard seqnum wrap (#545)

### LINKED ISSUE #545: nexus-fix-engine: recovery + sequence-ownership cleanup
Recovery and sequence-ownership cleanups from the 2026-06 audit.

- **`recover()` separate from `open()`** (`persist.rs`). After a crash, forgetting to call `recover()` leaves `next_outbound = 1` → resend from seq 1 → seqnum collision with the peer. Fold recovery into `open()`, or return a `MustRecover` wrapper that forces it before use. (The existing `recover_sets_next_outbound` test demonstrates the footgun: open → 1, then `recover()` → 8.)
- **Two sources of truth for `next_outbound`** — `SessionState::allocate_seq` and `FixJournal::store`/`recover`; nothing enforces they agree. Pick one owner (single-writer).
- **Unchecked `u32` seqnum `+= 1`** across session + journal — debug panic / release wrap-to-0 at 2³², which collides with reset semantics. Document the daily-reset assumption and/or guard it.

Part of #523.

ChronoCoders:
Closes #545 (two-sources-of-truth deferred).

- `persist.rs`: `open()` calls `recover_from_journal()` internally; public `recover()` removed to eliminate the forget-to-call footgun; `recover_sets_next_outbound` test updated to reflect the new contract
- `session/mod.rs`: all bare `next_outbound += 1` and `next_inbound += 1` → `wrapping_add(1)`, consistent with the journal; documents the daily-reset assumption

### DIFF
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -95,17 +95,19 @@ impl FixJournal {
         } else {
             conductor.session().open()?
         };
-        Ok(Self {
+        let mut this = Self {
             journal,
             _conductor: conductor,
             offsets: vec![None; window].into_boxed_slice(),
             window,
             next_outbound: 1,
             next_inbound: 1,
-        })
+        };
+        this.recover_from_journal();
+        Ok(this)
     }
 
-    pub fn recover(&mut self) {
+    fn recover_from_journal(&mut self) {
         let mut pos = self.journal.read_start();
         let mut last_seq: Option<u32> = None;
         while let Some(frame) = self.journal.read_next(&mut pos) {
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0:
Let's trim this to just the `recover()` → `open()` fold (E6) — that part's solid and we want it. Drop the `wrapping_add` seqnum change: we looked into how 24/7 venues handle this, and silently wrapping to seq 0 (an invalid FIX seqnum) isn't the right answer. The production pattern is to **error before the limit** and have the caller drive a sequence reset (`ResetSeqNumFlag=Y`), with the current seq exposed so they can reset well ahead of the boundary. I'm filing a separate issue for that.

Could you also change "Closes #545" → "Closes part of #545"? The sequence-ownership piece (single `next_outbound` owner, E8) stays in #545.

PR #556, opened 2026-06-26
fix(fix-engine): seqnum limit at i32::MAX + recover() folded into open() (#545)

ChronoCoders:
Closes part of #545.

**Recover fold**: `recover()` folded into `open()` — removed as a separate public call. Crash-restart no longer silently starts at `next_outbound=1`.

**Seqnum limit at `i32::MAX`**: outbound seqnums capped at `i32::MAX` for signed-int interoperability (Java/Python). `allocate_seq` returns `Result<u32, SessionError::SeqNumExhausted>` so the app layer can force a sequence reset or logout. Internal admin sends disconnect with `DisconnectReason::SeqNumExhausted` on breach via a private `bump_outbound` helper.

### DIFF
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -15,7 +15,7 @@ use nexus_fix_codec::{
 };
 use nexus_fix_engine::{
     AdminMsg, CompId, DisconnectReason, Event, FixJournal, FrameError, FrameReader, FrameWriter,
-    Out, ReplayItem, SessionConfig, SessionState, State,
+    Out, ReplayItem, SessionConfig, SessionError, SessionState, State,
 };
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio::time::timeout_at;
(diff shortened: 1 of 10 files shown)


Abso1ut3Zer0 (approved):
Approving — clean implementation. The recover-into-open fold is solid, and the seqnum limit is done right: `allocate_seq -> Result<_, SeqNumExhausted>` with a check-then-`+= 1` (no wrap, no overflow), `bump_outbound` for internal admin sends that disconnects gracefully on exhaustion, and `i32::MAX` as the cap for signed-int interop. Library callers handle the `Result` cleanly (no internal `.unwrap()`), and the boundary tests for both `allocate_seq` and `bump_outbound` are real. CI green.

Couple of tracking notes (non-blocking):
- Could you change "Closes #545" → "Closes part of #545"? The single-`next_outbound`-owner cleanup (E8) stays in #545.
- This covers the error-before-wrap half of #559 plus the `i32::MAX` cap and monitoring via `state().next_outbound_seq()`. #559 stays open for the rest: a configurable limit (a venue may reset at ~100M, well below `i32::MAX`), the optional soft-threshold warning, and `ResetSeqNumFlag` initiation — without that last one the recovery loop isn't end-to-end yet (after a `SeqNumExhausted` disconnect, reconnecting with `141=Y` isn't supported).

Minor: the `blocking_session.rs` example `.unwrap()`s `allocate_seq` — since this PR adds the graceful-error path, modeling the `Err(SeqNumExhausted)` handling there (rather than a `.unwrap()` a user would copy into production) would be a nice touch. Not blocking.

PR #557, opened 2026-06-26
feat(fix-codec): FixDecimal const getters, const new, new_unchecked (#555)

### LINKED ISSUE #555: nexus-fix-codec: FixDecimal getters + const constructors (new -> Option, new_unchecked)
Follow-up from #551, which privatized `FixDecimal`'s `mantissa`/`scale` fields to enforce the `scale <= 19` invariant. Two API-ergonomics items:

- **Getters:** add `pub const fn mantissa(&self) -> i64` and `pub const fn scale(&self) -> u8` (matching `FixTime`'s const getters) — restores read access now that the fields are private.
- **Constructors:** make `new` a `const fn -> Option<Self>` (mirrors `nexus-decimal`'s `Decimal::from_scaled`, enables `const` construction); add a `const fn new_unchecked(mantissa, scale) -> Self` for hot-path sites where the scale is statically known-good. Existing `.unwrap()` call sites are unchanged by the `Result -> Option` switch.

Related to #546.

ChronoCoders:
Closes #555.

- `pub const fn mantissa(&self) -> i64` and `pub const fn scale(&self) -> u8` — restores read access after #551 privatized the fields
- `new` → `const fn -> Option<Self>` (was `Result<Self, FixValueError>`; all `.unwrap()` call sites unchanged)
- `pub const fn new_unchecked(mantissa: i64, scale: u8) -> Self` for statically-valid hot-path construction

`property.rs`: updates `Err(FixValueError::Overflow)` match → `.is_none()` and drops the now-unused `FixValueError` import.

### DIFF
--- a/nexus-fix-codec/src/types.rs
+++ b/nexus-fix-codec/src/types.rs
@@ -23,12 +23,23 @@ pub struct FixDecimal {
 }
 
 impl FixDecimal {
-    pub fn new(mantissa: i64, scale: u8) -> Result<Self, FixValueError> {
+    pub const fn new(mantissa: i64, scale: u8) -> Option<Self> {
         if scale > 19 {
-            return Err(FixValueError::Overflow);
+            return None;
         }
+        Some(Self { mantissa, scale })
+    }
 
-        Ok(Self { mantissa, scale })
+    pub const fn new_unchecked(mantissa: i64, scale: u8) -> Self {
+        Self { mantissa, scale }
+    }
+
+    pub const fn mantissa(&self) -> i64 {
+        self.mantissa
+    }
+
+    pub const fn scale(&self) -> u8 {
+        self.scale
     }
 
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
Approving — faithful to #555: `const fn new -> Option`, `new_unchecked`, and `const fn` `mantissa()`/`scale()` getters matching `FixTime`. The `Option` migration is transparent at the `.unwrap()` call sites, and the proptest now uses `.is_none()`. CI green.

One non-blocking should-add: give `new_unchecked` a one-line doc stating its contract — caller must ensure `scale <= 19`, otherwise `encode`/`Display` produce incorrect output in release (the debug-asserts only fire in debug). A `debug_assert!(scale <= 19)` inside it would also catch misuse early at zero release cost. Thanks!

PR #558, opened 2026-06-26
fix(fix-codec): FixDecimal encode/Display use checked_pow (#546)

### LINKED ISSUE #546: nexus-fix-codec: parser/encoder overflow hardening (parse_tag, decimal encode)

ChronoCoders:
Closes #546 (decimal encode half; parse_tag half closed in #552).

Replaces bare `10u64.pow(self.scale as u32)` in `encode` and `Display` with `checked_pow(...).expect("scale ≤ 19 by constructor")`, dropping the `debug_assert!` guards. The scale invariant is enforced at construction by #551; `checked_pow` makes the dependency self-documenting and safe under any future refactor.

The existing `decimal_roundtrip` proptest covers encode ↔ parse over the full `(i64, 0..=19)` domain.

### DIFF
--- a/nexus-fix-codec/src/types.rs
+++ b/nexus-fix-codec/src/types.rs
@@ -138,17 +138,14 @@ impl FixDecimal {
 
         let abs = self.mantissa.unsigned_abs();
 
-        debug_assert!(
-            self.scale <= 19,
-            "FixDecimal::encode: scale over 19 ({}) will result in an integer overflow",
-            self.scale
-        );
         if self.scale == 0 {
             pos += encode_u64(abs, &mut buf[pos..]);
             return pos;
         }
 
-        let scale_pow = 10u64.pow(self.scale as u32);
+        let scale_pow = 10u64
+            .checked_pow(self.scale as u32)
+            .expect("scale ≤ 19 by constructor");
         let integer = abs / scale_pow;
         let frac = abs % scale_pow;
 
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0 (approved):
Approving — solid refinement. `checked_pow(...).expect("scale ≤ 19 by constructor")` turns the impossible `scale > 19` case into a loud, documented panic in both debug and release, instead of the old release-build silent `pow` wrap that would corrupt a serialized price. Since `new`/`parse` enforce the invariant (private fields, #551), the expect never fires in practice — and it's a nice safety net for the `new_unchecked` footgun too. Removing the now-redundant debug_asserts is right. CI green.

Optional future micro-opt (not this PR): a `const POW10: [u64; 20]` table would be faster than `pow` and drop the `checked_pow`/`expect` entirely (a 20-entry table can't overflow).

Issue #560, opened 2026-06-26
nexus-fix-engine/codec: review follow-up nits


Abso1ut3Zer0:
Small non-blocking follow-ups flagged while reviewing the FIX hardening/cleanup PRs (#550, #553, #556, #557, #558). All good-first-issue sized.

1. **Assert the gap-fill `new_seq` in the resend test** (`nexus-fix-engine/tests/transport.rs`). The resend-clamp tests only assert a frame came back (`n > 0`), not the content — so the off-by-one fix (#550) isn't regression-guarded. Assert `new_seq == next_outbound` in the huge-end-seq test.
2. **`blocking_session.rs` example: handle `Err(SeqNumExhausted)`** instead of `.unwrap()` on `allocate_seq` (#556 added the graceful-error path). Examples get copied into production; a copied `.unwrap()` panics at exhaustion (~9h for a 24/7 high-rate session). Model the reset path.
3. **`FixDecimal::new_unchecked` doc + optional `debug_assert`** (#557). The constructor skips the `scale <= 19` check; add a one-line doc stating the caller's contract (violating it produces incorrect `encode`/`Display` output), and optionally `debug_assert!(scale <= 19)` to catch misuse early.
4. **`const POW10: [u64; 20]` lookup table** for `FixDecimal::encode`/`Display` (#558). Faster than `pow` and removes the `checked_pow`/`expect` entirely (a 20-entry table can't overflow).

Part of #523.

On0n0k1:
Hello is this issue resolved? Or are there remaining changes missing?

ChronoCoders:
All four items are done. The huge-end-seq test now asserts new_seq == next_outbound (transport.rs:465), blocking_session.rs handles SeqNumExhausted (line 168), new_unchecked has its doc and debug_assert (types.rs:56-59), and the POW10 table is at types.rs:25. Nothing is pending here.

ChronoCoders:
All four items are done -- confirmed in the previous comment. Closing.

PR #563, opened 2026-06-26
fix: new_unchecked contract doc+assert; example SeqNumExhausted handling

ChronoCoders:
Addresses review feedback from #556 and #557.

- `FixDecimal::new_unchecked`: add one-line doc stating the `scale <= 19` invariant and a `debug_assert!` to catch misuse early
- `blocking_session.rs`: replace `.unwrap()` on `allocate_seq` with explicit `SeqNumExhausted` handling so the example models the error path callers should implement

Closes part of #560.

### DIFF
--- a/nexus-fix-codec/src/types.rs
+++ b/nexus-fix-codec/src/types.rs
@@ -30,7 +30,10 @@ impl FixDecimal {
         Some(Self { mantissa, scale })
     }
 
+    /// Construct without checking `scale <= 19`. Caller must ensure the
+    /// invariant; violating it causes incorrect output from `encode`/`Display`.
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
Approving — both nits done right. `new_unchecked` gets the contract doc + a `debug_assert!` (zero release cost, and a `const` use with a bad scale now fails at compile time), and the example replaces `.unwrap()` with explicit `SeqNumExhausted` handling that models the reset path. CI green.

One housekeeping fix: the body says "Closes part of #545," but #545 is already closed and unrelated — these nits are tracked in #560, and this addresses 2 of its 4 (the resend `new_seq` assertion and the POW10 table remain). Could you point it at #560 instead so the tracking stays straight? Thanks.

PR #565, opened 2026-06-28
fix(fix-engine): journal all outbound messages; add reset_seq_nums

### LINKED ISSUE #561: nexus-fix-engine: journal all outbound messages (incl. admin) — authoritative next_outbound + audit trail
Currently only application messages are journaled (`send_app` → `FixJournal::store`); admin messages (Logon, Heartbeat, TestRequest, ResendRequest, SequenceReset, Logout) are sent via `encode_admin`/`flush_out` and never stored. This was an optimization for the resend path — admin messages gap-fill on a ResendRequest, so their content isn't strictly needed — but it costs us two things:

1. **Incomplete audit/debug trail.** The journal is a partial record (orders only); you can't see when logons / heartbeats / resend-requests went out. Production FIX engines journal the full message stream for replay, recovery, and audit.
2. **The E8 seqnum drift.** `SessionState.next_outbound` counts every outbound; `FixJournal.next_outbound` only counts journaled (app) messages → they diverge, and crash recovery under-counts (admin seqnums are lost). This was the deferred E8 part of #545 (which auto-closed via #556's "Closes #545").

## Plan
- **Journal every outbound message**, admin included — admin sends go through `store` too, not just `send_app`.
- **Resend is unaffected.** `ResendIter` already gap-fills admin *by `MsgType`* (`is_admin_type`), so a journaled admin frame still gap-fills on replay — storing it changes nothing about resend behavior.
- **`journal.next_outbound` becomes the true last-sent**, so it can be the single authoritative seqnum source (fixes E8). Add the bridge so the session restores its outbound seqnum from the journal on restart — today there's no API to set `SessionState.next_outbound` from the recovered value.
- Cost is negligible: heartbeats are ~30s and tiny, logons rare — nothing next to order flow at 100k/s.

## Out of scope / follow-up
A full debug trail also wants the **inbound** side (what the peer sent us). The resend journal is inherently outbound; capturing inbound is a separate, larger item.

Supersedes the E8 part of #545. Part of #523.

ChronoCoders:
Fixes the E8 seqnum drift from #561.

- `transport.rs`: every outbound admin message (Logon, Logout, Heartbeat, TestRequest, ResendRequest, SequenceReset) now goes through `journal.store()` via a free function `store_admin`. Previously only app messages were journaled, so `journal.next_outbound` diverged from `state.next_outbound` after any admin activity. Resend behavior is unaffected — `ResendIter` already gap-fills admin by `MsgType`.
- `session/mod.rs`: adds `SessionState::reset_seq_nums(next_outbound, next_inbound)` so callers can restore sequence numbers from a recovered journal before reconnecting.
- `tests/transport.rs`: `journal_recovers_admin_seqnums` — runs a logon+logout session, reopens the journal, asserts `next_outbound == 3` (logon seq=1, logout-ack seq=2).

Closes #561.

### DIFF
--- a/nexus-fix-engine/src/session/mod.rs
+++ b/nexus-fix-engine/src/session/mod.rs
@@ -166,6 +166,13 @@ impl SessionState {
         Some(s)
     }
 
+    /// Restores outbound and inbound sequence numbers from a recovered journal.
+    /// Call before connecting to resume a prior session without gaps.
+    pub fn reset_seq_nums(&mut self, next_outbound: u32, next_inbound: u32) {
+        self.next_outbound = next_outbound;
+        self.next_inbound = next_inbound;
+    }
+
     /// Earliest instant at which [`on_timeout`](Self::on_timeout) has work.
     pub fn next_timeout(&self) -> Option<Instant> {
         match self.state {
--- a/nexus-fix-engine/src/transport.rs
+++ b/nexus-fix-engine/src/transport.rs
@@ -10,7 +10,7 @@ use nexus_fix_codec::{
 use crate::frame::{FrameError, FrameWriter};
 use crate::framework::{Message, MessageReader, MessageWriter, SessionConfig, SessionError};
 use crate::persist::{FixJournal, ReplayItem};
-use crate::session::{DisconnectReason, Event, SessionState, State};
+use crate::session::{AdminMsg, DisconnectReason, Event, SessionState, State};
 use crate::timestamp::UTC_TIMESTAMP_LEN;
 
 #[derive(Debug)]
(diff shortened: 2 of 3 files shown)


Abso1ut3Zer0 (approved):
Approving — journaling all outbound (admin via `store_admin`) is the right fix: the seqnum drift is gone, you get the full audit trail, and resend is unaffected since admin still gap-fills by `MsgType`. The admin-seqnum recovery test confirms it.

A few follow-ups (fold in here or a quick follow-up):
- **Auto-wire recovery in `from_parts`** — it has both the session and the recovered journal, so reconcile them there (`state.reset_seq_nums(journal.next_outbound(), …)`). Then restart-resume is automatic and nobody can forget it and resume at seq 1. Fresh journal → no-op; recovered → resumes. (Outbound only; inbound stays with the logon handshake.)
- **Doc `reset_seq_nums` as the local coordination primitive** — it sets *our* counters; a mid-session reset must be paired with a coordinated `ResetSeqNumFlag=Y` exchange or you desync the peer. (Deliberately *not* gating it by state — mid-session reset is valid for 24/7 markets, it just has to be coordinated.)
- Minor: the bridge itself isn't tested (the test checks the journal's recovered value, not that a resumed session sends at it), and `store_admin` encodes-then-stores so a rare `store` backpressure failure leaves encoded-but-unsent bytes + re-drifts that one seqnum.

Separately, the outbound side isn't a typed/dictionary-driven path yet (engine hand-rolls admin; `FixDictionary` is decode-only) — tracking in #568. Once that lands, the reset becomes a first-class typed op and callers won't touch `reset_seq_nums` directly.

Issue #562, opened 2026-06-26
nexus-fix-engine: session-level Reject (35=3) generation for invalid inbound (RefTagID + SessionRejectReason)


Abso1ut3Zer0:
FIX rejects a session-level problem in an inbound message with a **Reject (35=3)** that carries the offending details and keeps the session alive (the engine still advances its expected seqnum). Today the engine generates no Rejects — it either silently ignores or surfaces an app error:

- **Undefined/unknown tags** → silently ignored (`parse_tag` saturates to `u32::MAX`, so `find_tag` skips the field).
- **Out-of-range SequenceReset** (backward/zero `NewSeqNo` in Reset mode) → silently dropped (the #550 guard returns an empty `Out`).
- **Malformed / missing required field** → returned to the app as `Err(Protocol(...))`, but no Reject is sent to the peer.

None of these tell the counterparty *what* was wrong, which is what the spec mandates.

## Plan
Add a `Reject (35=3)` admin message carrying:
- `RefSeqNum (45)` — the rejected message's `MsgSeqNum`
- `RefTagID (371)` — the offending tag. **Make sure this is actually populated** — it was a known gap in other engines (per the Artio review).
- `RefMsgType (372)` — optional, the rejected `MsgType`
- `SessionRejectReason (373)` — 0 = invalid tag number, 1 = required tag missing, 2 = tag not defined for this message type, 5 = value out of range, 6 = incorrect data format
- `Text (58)` — optional human-readable reason

Wire it into the session for the cases above (invalid/undefined tag, out-of-range SequenceReset, malformed/missing required field), keeping the session alive and advancing the expected seqnum.

## Consider
Policy: which errors warrant a Reject vs a Logout+disconnect vs silent ignore. FIX permits unknown/user-defined fields, and some engines (QuickFIX) gate this behind config (`ValidateUserDefinedFields`) — we may want a similar toggle rather than rejecting every unknown tag.

Part of #523.

PR #566, opened 2026-06-28
fix(session): Reject (35=3) generation

ChronoCoders:
Closes part of #562.

**What**

- `AdminMsg::Reject` variant (`seq`, `ref_seq_num`, `ref_tag_id: Option<u32>`, `session_reject_reason: u8`)
- `SessionState::on_reject_inbound` — validates seq first (gap → ResendRequest, too-low → Logout/dup-ignore), then allocates outbound seq, sends Reject, advances `next_inbound` via `validate_seq`
- `on_sequence_reset`: backward Reset (new_seq < next_inbound or new_seq == 0) now sends Reject (RefTagID=36, reason=5) instead of silent no-op
- `encode_admin`: Reject encodes tags 45 (RefSeqNum), 371 (RefTagID, omitted if `None`), 373 (SessionRejectReason via `encode_fix_uint`)
- `transport::recv`: missing tag 35 → `on_reject_inbound(seq, poss_dup, Some(35), 1)` + flush; returns `Ok(None)` so session stays alive

**Tests**

- `sequence_reset_backward_sends_reject` — unit
- `reject_inbound_sends_reject_and_advances_seq` — unit
- `reject_inbound_gap_sends_resend_request` — unit

### DIFF
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -294,6 +294,7 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
             AdminMsg::TestRequest { .. } => b"1",
             AdminMsg::ResendRequest { .. } => b"2",
             AdminMsg::SequenceReset { .. } => b"4",
+            AdminMsg::Reject { .. } => b"3",
         };
 
         let seq = match admin {
@@ -302,7 +303,8 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
             | AdminMsg::Heartbeat { seq, .. }
             | AdminMsg::TestRequest { seq, .. }
             | AdminMsg::ResendRequest { seq, .. }
-            | AdminMsg::SequenceReset { seq, .. } => seq,
+            | AdminMsg::SequenceReset { seq, .. }
+            | AdminMsg::Reject { seq, .. } => seq,
         };
 
         let begin_string = self.begin_string;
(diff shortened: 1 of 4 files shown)


Abso1ut3Zer0 (changes requested):
Nice — the Reject mechanism is well-built: the new `AdminMsg::Reject`, the `encode_admin` field set (45 RefSeqNum / 371 RefTagID omitted-if-None / 373), keeping the session alive (`Ok(None)`), and the backward-SequenceReset → Reject (RefTagID=36, reason=5, correctly *not* advancing) are all right. Two things to fix before merge, since one reopens a hardening gap.

### 🔴 `on_reject_inbound` mis-sequences out-of-order messages

```rust
let Some(seq) = self.bump_outbound(now, &mut out) else { return out; };
out.push_admin(AdminMsg::Reject { seq, ref_seq_num: inbound_seq, … });
self.next_inbound = self.next_inbound.saturating_add(1);   // blind +1, no seq check
```

It advances `next_inbound` by 1 **without checking `inbound_seq == next_inbound`**, and the transport's missing-35 path calls it *before* any sequence validation. So if a malformed message arrives with a **gap** (expected 3, got 7):

- FIX requires the gap to take precedence → send a **ResendRequest**, not a Reject.
- Doing `next_inbound 3 → 4` while the message was seq 7 **desyncs the session**.

Per the spec, MsgSeqNum validation precedes field-level rejection (QuickFIX et al. do gap → ResendRequest first, then Reject). Suggest having `on_reject_inbound` run `validate_seq` first: gap → ResendRequest (no reject), too-low → dup/logout handling, in-sequence → Reject + advance (validate_seq already advances, so drop the manual `+1`). It'll also need `poss_dup` for the too-low branch. The current unit tests only cover `inbound_seq == next_inbound`, so a gap-input test would lock this in.

### SessionRejectReason encoded as a single digit

```rust
fmt.field(373, &[b'0' + session_reject_reason]);
```

Only works for reasons 0–9. Reasons ≥ 10 exist (10 = SendingTime accuracy, 11 = invalid MsgType, 16 = incorrect NumInGroup, …) and would encode to garbage (`b'0' + 11 = ';'`). Only 1 and 5 are used today so it's latent, but please use `encode_fix_uint` like the other numeric fields.

### Scope

This covers missing-35 + backward-SequenceReset, but not missing-34 (still returns `Err`), malformed fields, or undefined tags — so could you change "Closes #562" → "Closes part of #562" and leave it open for the rest?

Abso1ut3Zer0 (approved):
Approving — the sequencing fix is exactly right: `on_reject_inbound` now runs `validate_seq` first, so a gapped malformed message sends a ResendRequest instead of rejecting+mis-advancing, and only in-sequence messages get a Reject. The new `reject_inbound_gap_sends_resend_request` test locks that in (asserts ResendRequest + no advance). Reason encoding fixed in the sync engine, and scope re-marked to 'Closes part of #562'. Nice.

One non-blocking note to clean up: the async engine's copy of `encode_admin` (`nexus-async-fix-engine/src/lib.rs`) still encodes field 373 as `&[b'0' + session_reject_reason]` — the single-digit version that garbles `SessionRejectReason >= 10`. Only reasons 1 and 5 are used today so it's latent, but please mirror the `encode_fix_uint` fix there (or it folds in naturally when #568 de-duplicates the two encode_admin copies).

PR #567, opened 2026-06-28
feat(fix-engine): bounded sequence numbers (#559)

### LINKED ISSUE #559: nexus-fix-engine: bounded sequence numbers — error before wrap + configurable limit for proactive reset
FIX seqnums accumulate over a session's life. For 24/7 (crypto) and 24/5 (FX) markets there's no daily session boundary, so a long-lived high-rate session approaches the seqnum limit in hours — at ~132k/s, the signed-32 ceiling (2^31 ≈ 2.1B, which several engines enforce on `BeginSeqNo`) is ~4.5h away. Silently wrapping `next_outbound`/`next_inbound` to 0 yields an invalid seqnum (they start at 1) → peer rejects → ungraceful break.

Production engines reset *before* the boundary rather than wrap:
- **Bullish** caps at 100,000,000 → unsolicited Logout + a News warning 1,000 messages prior.
- **Coinbase** resets per-connection (server sends a SequenceReset on Logon).
- **Gemini** resets via `ResetSeqNumFlag`.

The FIX spec provides in-session `ResetSeqNumFlag(141)=Y` (NextNumIn/Out → 1) specifically for continuously-operating markets.

## Requirements

1. **Error before wrap** — `allocate_seq` (and the inbound advance) must not silently wrap past a configurable limit. When allocation would exceed it, return an error / signal "sequence reset required" so the caller drives the reset (Logout + reconnect, or in-session `ResetSeqNumFlag=Y`). The library sets the boundary; the caller owns the reset policy.
2. **Surface the current seqnums for monitoring** — `SessionState::next_outbound_seq()` / `next_inbound_seq()` already exist (reachable via `FixConnection::state()`); make them first-class / documented as the proximity-to-limit hook, and consider a direct `FixConnection` accessor. Some venues reset at a low threshold (e.g. 100M) far below the type limit, so the caller needs visibility to reset well beforehand.
3. **Configurable limit** (default well below 2^31; a venue may set 100M), with an optional soft-threshold warning/event ahead of the hard limit (the Bullish News-warning pattern).
4. **`ResetSeqNumFlag` support** for both initiating (initiator sends 141=Y on Logon) and the in-session reset, zeroing both counters to 1.

Supersedes the seqnum-overflow (E9) framing in #545, which assumed a daily reset that isn't universal. Part of #523.

Refs: [Bullish — 24/7 FIX sequence number overflow](https://medium.com/bullish-engineering/24-7-trading-addressing-fix-sequence-number-overflow-9d239799e0ba), [Gemini — when can sequence numbers reset](https://docs.gemini.com/fix/overview/session-level-messages/when-can-sequence-numbers-reset).

ChronoCoders:
Closes #559.

Adds in-session sequence reset (`reset_sequence`/`connect_reset`) and
removes the warning-based approach. The reset handshake uses two new
states: `AwaitingResetDrain` and `AwaitingResetAck`.

Changes:

- `session/event.rs`: add `AwaitingResetDrain`, `AwaitingResetAck` states;
  add `ResetTimeout` disconnect reason; drop `SeqNumWarning`
- `session/mod.rs`: full state machine rewrite
  - `connect(now)` - normal logon, no reset param
  - `connect_reset(now)` - logon with `ResetSeqNumFlag(141)=Y`
  - `reset_sequence(now)` - in-session reset: sends TestRequest to drain,
    then `Logon(141=Y)` on Heartbeat reply, transitions through
    `AwaitingResetDrain` -> `AwaitingResetAck` -> `Active`
  - `allocate_seq` returns `Err(ResetInProgress)` during reset window
  - peer-initiated reset handled in `on_logon` while `Active`
  - remove `seq_limit`, `warn_threshold`, `warn_fired` and related methods
- `framework.rs`: add `ResetInProgress` to `SessionError`; remove
  `SeqNumWarning` from `Message`
- `transport.rs`: update `connect` API; add `connect_reset`; remove
  `approaching_seq_limit`
- `nexus-async-fix-engine`: same connect API update

Pending buffer helpers for callers managing in-flight messages during a
reset are deferred to a follow-up issue (#574).

### DIFF
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -111,7 +111,7 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
         let out = self
             .state
             .connect_reset(Instant::now())
-            .map_err(|()| Error::Io(io::Error::other("connect_reset: session not disconnected")))?;
+            .map_err(|e| Error::Io(io::Error::other(e.to_string())))?;
         self.flush_out(out).await
     }
 
@@ -120,7 +120,7 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
         let out = self
             .state
             .reset_sequence(Instant::now())
-            .map_err(|()| Error::Io(io::Error::other("reset_sequence: session not active")))?;
+            .map_err(|e| Error::Io(io::Error::other(e.to_string())))?;
         self.flush_out(out).await
     }
 
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -10,8 +10,8 @@ use std::io;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use nexus_fix_codec::{
-    encode_fix_uint, find_tag, parse_fix_bool, parse_fix_seqnum, parse_fix_uint, FieldReader,
-    FrameFormatter,
+    FieldReader, FrameFormatter, encode_fix_uint, find_tag, parse_fix_bool, parse_fix_seqnum,
+    parse_fix_uint,
 };
 use nexus_fix_engine::{
     AdminMsg, CompId, DisconnectReason, Event, FixJournal, FrameError, FrameReader, FrameWriter,
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -346,16 +346,11 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
             fmt.field(52, &ts);
 
             match admin {
-                AdminMsg::Logon { heart_bt_int_s, .. } => {
+                AdminMsg::Logon { heart_bt_int_s, .. }
+                | AdminMsg::LogonReset { heart_bt_int_s, .. } => {
                     let mut buf = [0u8; 10];
                     let n = encode_fix_uint(heart_bt_int_s, &mut buf);
                     fmt.field(108, &buf[..n]);
-                }
-                AdminMsg::LogonReset { heart_bt_int_s, .. } => {
-                    let mut buf = [0u8; 10];
-                    let n = encode_fix_uint(heart_bt_int_s, &mut buf);
-                    fmt.field(108, &buf[..n]);
-                    fmt.field(141, b"Y");
                 }
                 AdminMsg::Logout { .. } | AdminMsg::Heartbeat { echo: None, .. } => {}
                 AdminMsg::Heartbeat {
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -101,8 +101,10 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
     }
 
     /// Initiates a session: encodes and sends the opening Logon.
-    pub async fn connect(&mut self) -> Result<(), Error> {
-        let out = self.state.connect(Instant::now());
+    ///
+    /// Set `reset = true` to send `ResetSeqNumFlag(141)=Y`.
+    pub async fn connect(&mut self, reset: bool) -> Result<(), Error> {
+        let out = self.state.connect(reset, Instant::now());
         self.flush_out(out).await
     }
 
@@ -323,10 +325,13 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
             fmt.field(52, &ts);
 
             match admin {
-                AdminMsg::Logon { heart_bt_int_s, .. } => {
+                AdminMsg::Logon { heart_bt_int_s, reset, .. } => {
                     let mut buf = [0u8; 10];
                     let n = encode_fix_uint(heart_bt_int_s, &mut buf);
                     fmt.field(108, &buf[..n]);
+                    if reset {
+                        fmt.field(141, b"Y");
+                    }
                 }
                 AdminMsg::Logout { .. } | AdminMsg::Heartbeat { echo: None, .. } => {}
                 AdminMsg::Heartbeat {
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -10,8 +10,8 @@ use std::io;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use nexus_fix_codec::{
-    FieldReader, FrameFormatter, encode_fix_uint, find_tag, parse_fix_bool, parse_fix_seqnum,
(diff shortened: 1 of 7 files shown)


Abso1ut3Zer0 (changes requested):
After digging into how Artio and QuickFIX handle this, let's **reduce this PR to first-class sequence reset support** and drop the warning machinery. Real engines don't have a built-in "approaching limit" warning — they expose the current seqnum and let the app/venue decide (Bullish's News-before-100M is app-level; Artio's actual direction is wider 64-bit seqnums). The only error we want is the genuine "ran out of seqnums" one, which you already have.

**Remove:**
- `Event::SeqNumWarning` + `Message::SeqNumWarning`, `warn_threshold`/`warn_fired`, the `on_timeout` warning check, `approaching_seq_limit()`, and `with_seq_limit`/`with_seq_warn_threshold`. (That `on_timeout` check is also misplaced — it only runs on a read-idle, so it never fires for a busy session anyway.)

**Keep:**
- `SeqNumExhausted` from `allocate_seq` at the hard ceiling (`i32::MAX`) — the valid "truly out of seqnums" error, no seqnum consumed.
- `next_outbound_seq()` so the app can monitor and run its own reset policy (how the real engines expose it).

**Rework the connect API (drop the bool):**
- `connect(now)` + `connect_reset(now)` instead of `connect(reset: bool, now)`. Keep the typed `AdminMsg::Logon { reset }` → `141=Y`.

**Add first-class in-session reset** — the main piece, matching Artio (`Session.tryResetSequenceNumbers()`) / QuickFIX (mid-session `141=Y`):
- A `reset_sequence()` op that works while **Active**: drain in-flight (TestRequest → await Heartbeat so there's no gap), then send `Logon 141=Y, seq=1` and reset `next_outbound` to 1; reset `next_inbound` on the counterparty's `141=Y` ack (confirm-first, per Artio — if they reject, keep the old value), with an await-ack timeout → disconnect.
- Responder side: `on_logon` currently rejects a Logon arriving while `Active` — it needs to accept a `141=Y` Logon mid-session, reset, and reply `141=Y`.
- A wrong-state reset should error / be a visible no-op, never silently nothing.

Tests: in-session reset round-trip (initiate while Active → counters to 1, `141=Y` sent), responder side (inbound `141=Y` while Active → reset + ack), and the await-ack timeout → disconnect.

This is the meatiest part of the session work — happy to talk through the state-machine shape if useful. #559 stays the tracking issue; its warning/threshold items are dropped per the above.

ChronoCoders:
Updated the PR description to match our discussion. Will rework the implementation accordingly.

Abso1ut3Zer0:
The rework is a faithful take on the agreed direction — warning machinery dropped, `connect`/`connect_reset` split (no bool), the `AwaitingReset*` states, `allocate_seq → ResetInProgress`, responder path, and await-ack timeout are all there. Tests pass, clippy clean. Two structural moves + four gaps before it's ready:

**Restage (encode moves out):**
Pull the `reset` flag and the `141=Y` encoding out of here — they move to #572/#573 as `encode_logon_reset` + the `LogonReset` variant. This PR becomes **state-machine-only**: emit `AdminMsg::Logon` vs `AdminMsg::LogonReset` at the right sites (`connect` → `Logon`; `connect_reset` / `reset_sequence` / reset-responder → `LogonReset`; acceptor reply mirrors what it received). No bool, no `FrameFormatter`, no `141=Y` literal in this PR.

**Gaps (the blockers):**

1. **`reset_sequence` isn't reachable through the transports.** It's on `SessionState`, but only `connect_reset` got a `FixConnection`/`AsyncFixConnection` wrapper — `reset_sequence` returns an `Out` the recv-loop never flushes, so the in-session reset never goes on the wire. Add the wrappers mirroring `connect_reset`.
2. **In-flight inbound is dropped during the reset window.** `on_app` / `on_test_request` gate to `Active | Resending | LogoutPending`, so a message arriving in `AwaitingResetDrain`/`AwaitingResetAck` is silently dropped. The inbound counter is still on the old numbering in those states, so let them through — `validate_seq` handles them. As-is this loses application messages mid-reset.
3. **The drain accepts any heartbeat.** The `AwaitingResetDrain` branch in `on_heartbeat` proceeds to the reset Logon on any heartbeat, before `validate_seq` and ignoring the TestReqID(112) echo — so it can complete prematurely or with an undetected gap, and the reset-to-1/1 then discards the unreceived message(s). Match the TestReqID echo and confirm inbound is caught up before sending the reset Logon.
4. **Wrong-state reset is a silent no-op.** `reset_sequence` returns `Out::EMPTY` when not `Active` (same for `connect_reset` when not `Disconnected`). Make it observable — a `Result`/`bool` or an event.

#1 and #2 are the hard ones (feature unreachable + silent data loss). With those in, plus a test for an in-flight app message during the window and a transport-level reset round-trip, this is in good shape.

ChronoCoders:
All four gaps addressed plus the restage. `reset_sequence` now has `FixConnection`/`AsyncFixConnection` wrappers so it actually goes on the wire. `on_app` and `on_test_request` guard includes `AwaitingResetDrain | AwaitingResetAck` so in-flight messages pass through to `validate_seq`. The drain heartbeat now checks the TestReqID echo matches `test_req_counter` and `seq == next_inbound` before sending the reset Logon. `connect_reset` and `reset_sequence` return `Result<Out, SessionError>` with `Err(InvalidState)` on wrong state. `141=Y` encoding is out of this PR, deferred to #572/#573; `LogonReset` encodes as `Logon` for now.

Abso1ut3Zer0 (approved):
Verified the rework — approving. All four gaps are fixed with targeted tests, and it builds clippy-clean (both crates) with all session tests passing:

- **`reset_sequence` wired** to `FixConnection` + `AsyncFixConnection`, tag 112 parsed into `echo_id` on both heartbeat paths.
- **In-flight inbound flows** — `on_app`/`on_test_request`/`on_heartbeat` admit `AwaitingReset*`; `in_flight_app_delivered_during_drain` proves it.
- **Drain is real** — requires `echo_id == test_req_counter` && `seq == next_inbound`, else falls through to `validate_seq`.
- **Observable** — `connect_reset`/`reset_sequence` → `Result<_, InvalidState>`, tested both ways.

Confirm-first inbound checks out end-to-end (drain consumes only the drain heartbeat; inbound flips to 2 on the ack; the `seq=1` reset Logon is handled before `validate_seq`). Responder path resets both counters and replies. Solid.

**One staging note for whoever merges:** `LogonReset` currently encodes as a plain `Logon` (the `141=Y` is deferred to #573 per the split), so the reset isn't functional on the wire until #573 lands — these two are a pair. The tests pass because they assert the state machine, not the bytes (the wire bytes get covered in #573). So land #567 with or just before #573, not as a standalone "reset works."

Two minor, non-blocking items — no need to touch them here, I'll fold them into a cleanup pass after this lands:
- `Event::Established` is re-emitted on reset completion; a distinct `ResetComplete` would avoid double-firing for consumers.
- A gapped drain heartbeat falls to ResendRequest and ends in the reset timeout rather than retrying the drain — safe (routes into reconnect-reset), just noting.

Nice work on the rework.

PR #570, opened 2026-06-29
feat(fix-engine): multi-session FixJournal under one directory

On0n0k1:
`FixJournal::open` hardcodes `sessions_on_disk().first()`, limiting one process to one session per directory. This PR wires up the `session_id` that `Conductor` already supports, allowing N independent sessions under a single directory.

**`FixJournal::open` now takes an explicit `session_id: u32`** (`persist.rs`)
Replaces the `.first()` heuristic with a caller-supplied ID. `conductor.session().session_id(id).open()` handles both recovery (manifest exists) and fresh creation (no manifest) in a single call, so the old `if/else` is removed. `recover_from_journal()` is unchanged — it already replays only the messages belonging to the opened session.

**All call sites updated to pass `0`** (`transport.rs`, `fix_conformance.rs`, `async_conformance.rs`, `blocking_session.rs`, internal tests in `persist.rs`)
Single-session callers each use their own directory, so `0` is the correct and conventional choice.

**Test: two concurrent sessions in one directory** (`persist.rs`)
Opens two `FixJournal` instances under the same directory with session IDs `0` and `1`, stores a different number of messages in each, confirms `next_outbound` is independent, drops both, reopens, and verifies each recovers its own sequence number without interference.

Related to #518

### DIFF
--- a/nexus-async-fix-engine/tests/async_conformance.rs
+++ b/nexus-async-fix-engine/tests/async_conformance.rs
@@ -47,7 +47,7 @@ async fn connect(port: u16, dir: &Path) -> AsyncFixConnection<TcpStream> {
             sender: CompId::new(b"ENGINE").unwrap(),
             target: CompId::new(b"PEER").unwrap(),
         },
-        FixJournal::open(dir, 256).unwrap(),
+        FixJournal::open(dir, 0, 256).unwrap(),
         BEGIN,
     )
 }
--- a/nexus-fix-engine/examples/blocking_session.rs
+++ b/nexus-fix-engine/examples/blocking_session.rs
@@ -112,7 +112,7 @@ fn run_acceptor(listener: &TcpListener, dir: &Path) {
             sender: CompId::new(b"ACCEPTOR").unwrap(),
             target: CompId::new(b"INITIATOR").unwrap(),
         },
-        FixJournal::open(dir, 256).unwrap(),
+        FixJournal::open(dir, 0, 256).unwrap(),
     );
 
     let mut n = 0usize;
@@ -141,7 +141,7 @@ fn run_initiator(addr: std::net::SocketAddr, dir: &Path) {
                 sender: CompId::new(b"INITIATOR").unwrap(),
                 target: CompId::new(b"ACCEPTOR").unwrap(),
             },
-            FixJournal::open(dir, 256).unwrap(),
+            FixJournal::open(dir, 0, 256).unwrap(),
         )
         .unwrap();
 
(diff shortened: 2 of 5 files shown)


ChronoCoders:
Three things to address.

persist.rs:93: wrong session_id silently opens a fresh journal. open_inner creates a new session for any ID with no manifest on disk. A reconnecting caller that passes the wrong session_id skips all unacknowledged message replay with no error returned.

persist.rs:99,174: next_inbound is not journaled and always resets to 1 on reopen. advance_inbound is memory-only; recover_from_journal never restores it. The new test only asserts next_outbound independence so this is invisible in CI.

persist.rs:87: doc comment is stale. Still reads 'one session per journal; if a session already exists the first one is reopened' -- both claims are wrong after this change.

Abso1ut3Zer0:
Thanks for the review here — @ChronoCoders' three points are all valid. Disposition so this PR can land focused on the multi-session (`session_id`) piece:

**Address in this PR:**
- **Stale doc** (Point 3) — `FixJournal::open`'s doc still reads "one session per journal; first one reopened," which is wrong after this change. Please update it, and have it state the **open-or-create** semantics explicitly (opens `session_id` if its manifest exists, else creates a fresh session) so the behavior is clear at the call site.

**Tracked separately (pre-existing / orthogonal — not blocking this PR):**
- **Wrong `session_id` silently opens a fresh journal** (Point 1) → #576. Real footgun now that ids are caller-supplied; wants an open-existing-only mode that errors when the manifest is absent. (`open_strict` only guards config mismatch, not existence.)
- **`next_inbound` not durable across restart** (Point 2) → #577. Genuine resume asymmetry — outbound recovers, inbound resets to 1; pre-existing and independent of multi-session.
- **One shared `Conductor` across journals** → #575. Orthogonal to the `session_id` work: this PR gives each `FixJournal` its own conductor (fine as a primitive), and collapsing N conductor threads to one per directory is a separate follow-up.

So: land the `session_id` support + the doc fix; the recovery-correctness items live in #576 / #577 and the shared-conductor optimization in #575.

On0n0k1:
Thank you for the feedback. I included the requested change on this PR so it can be merged.

On0n0k1:
The failing test is from `nexus-async-rt`. Unrelated to the changes from this PR.

Abso1ut3Zer0 (approved):
Doc fix looks good — the open-or-create semantics are clear now, and the multi-session (session_id) scope is complete. Approving.

If you're looking for what to pick up next on the multi-session front, **#575** is the natural follow-on — it's the orthogonal piece to this PR: running multiple journals through a *single* shared `Conductor` per directory (one background archival thread instead of one per session), rather than each `FixJournal` owning its own. nexus-journal already supports it at its layer, so it's mostly engine-side wiring. Yours if you want it.

Thanks for the quick turnaround on the doc.

Issue #568, opened 2026-06-29
nexus-fix-engine: bidirectional FixDictionary — typed outbound path (dictionary-driven admin encoding)


Abso1ut3Zer0:
`FixDictionary` is decode-only — every associated type is a decoder, and the engine reads inbound as typed `Message<'_, D>`. Outbound is asymmetric: `encode_admin` hand-rolls standard session fields via `FrameFormatter` (uses `D` only for `BEGIN_STRING`), and the codegen typestate encoders (`HeaderEncoder` / `<Msg>Encoder`) aren't part of the dictionary or used by the engine.

## Consequences
- **Venue-specific admin fields can't be sent** — most critically **Logon auth** (`Username 553`, `Password 554`, `RawData 96`, API-key/HMAC), which crypto venues require → can't actually log on to a real venue.
- **Session ops like the `ResetSeqNumFlag` reset must be hand-wired** — callers end up touching low-level primitives (`reset_seq_nums`) instead of a typed operation.

## Plan
Extend `FixDictionary` to be **bidirectional** — add encoder associated types/methods (backed by the codegen typestate encoders that already exist), and wire the engine's `encode_admin` + reframe path to use the dictionary's typed Logon/Heartbeat/… encoders.

Outcome: outbound symmetric with inbound, venue-specific admin fields expressible, and admin operations (incl. sequence reset) become first-class typed ops the user doesn't hand-roll.

Part of #523.

Abso1ut3Zer0:
### Scope correction, and the direction we landed on

**The body above is stale in one important way.** It says `encode_admin` "hand-rolls standard session fields via `FrameFormatter` (uses `D` only for `BEGIN_STRING`)". That was true when filed; **#572/#573 fixed it.** `FixDictionary` now has `encode_logon` / `encode_logon_reset` / `encode_logout` / etc., and `encode_admin` genuinely dispatches to them (`D::encode_logon(spare, mk_hdr(seq), heart_bt_int_s)`). Anyone scoping from the original text would redo work that has already landed.

**What actually remains is narrower.** Those encoders take no config and are static trait fns (no `&self`), so a venue has nowhere to source runtime credentials from. That, and only that, is what blocks logging onto a real venue.

### We are NOT generating a per-venue config from the XML

The obvious-looking fix (derive a typed `Config` from the Logon schema using `required="Y"`, wire it into the generated typestate encoders) **does not work**. We researched this rather than guessing.

[...]

PR #572, opened 2026-06-30
feat(fix-codec): add encoder methods to FixDictionary

ChronoCoders:
Extends FixDictionary with encoder associated types to mirror the existing decoder side.

Part of #568.

### DIFF
--- a/nexus-fix-codec/src/dict.rs
+++ b/nexus-fix-codec/src/dict.rs
@@ -13,6 +13,23 @@ pub trait FixAdminMsg<'buf>: Sized {
     fn decode(buf: &'buf [u8]) -> Result<Self, crate::DecodeError>;
 }
 
+pub struct AdminHeader<'a> {
+    pub seq: u32,
+    pub sender: &'a [u8],
+    pub target: &'a [u8],
+    pub ts: &'a [u8],
+}
+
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0:
Looking good — the typed encoder methods + `AdminHeader` are the right shape. One addition before it lands, tied to a design call on the reset work: **add `encode_logon_reset`** (emits `ResetSeqNumFlag(141)=Y`) alongside `encode_logon`.

We're dropping the `reset: bool` approach in favor of two distinct methods — a reset Logon and a plain Logon are different enough that a self-documenting method beats an opaque `true`/`false` at the callsite:

- `encode_logon` — plain Logon, no `141=Y`. Covers both the first Logon (seq 1) and a resume Logon (seq N); they differ only by seqnum, which is already in the header.
- `encode_logon_reset` — emits `141=Y`, seq 1.

That completes the dictionary's encode surface here, so #573 routes to both and #567 stays out of the encode path. Staging: #572 → #573 → #567.

One note: the new encoders aren't exercised by tests in this PR — they get their byte-parity coverage when #573 routes `encode_admin` through them, so this is best landed paired with #573 rather than on its own.

Abso1ut3Zer0 (approved):

PR #573, opened 2026-06-30
feat(fix-engine): wire encode_admin through FixDictionary encoder methods

ChronoCoders:
Closes part of #568.

`MessageWriter::encode_admin` now delegates to `D::encode_*` default methods added in #572 instead of hand-rolling `FrameFormatter` for each variant. The `AdminHeader` struct groups the four common session params (seq, sender, target, ts) passed to each call.

Dead `encode_u64` helper removed.

The async engine (`AsyncFixConnection`) is not generic over `D`, so its `encode_admin` is unchanged; that path stays correct from #571.

### DIFF
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -346,11 +346,16 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncFixConnection<S> {
             fmt.field(52, &ts);
 
             match admin {
-                AdminMsg::Logon { heart_bt_int_s, .. }
-                | AdminMsg::LogonReset { heart_bt_int_s, .. } => {
+                AdminMsg::Logon { heart_bt_int_s, .. } => {
                     let mut buf = [0u8; 10];
                     let n = encode_fix_uint(heart_bt_int_s, &mut buf);
                     fmt.field(108, &buf[..n]);
+                }
+                AdminMsg::LogonReset { heart_bt_int_s, .. } => {
+                    let mut buf = [0u8; 10];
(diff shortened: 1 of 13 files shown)


Abso1ut3Zer0:
Two things, both tied to the reset staging (#572 → #573 → #567):

1. **Carry the reset Logon here.** Split `AdminMsg::Logon` into `Logon` (plain) and `LogonReset` (`141=Y`), and add the match arm → `D::encode_logon_reset` (the method added in #572). No `reset: bool` — two variants, two methods. This makes #573 the home for *how a reset Logon renders to bytes*, so #567 stops touching `encode_admin` entirely.

2. **Keep the `encode_admin` enum match.** That match is the sans-IO seam — `SessionState` is dictionary-free, so it emits a closed set of `AdminMsg` and the D-aware writer dispatches to `D::encode_*`. That's correct; leave it as-is.

Heads-up: this PR and #567 both touch `framework.rs::encode_admin` and `AdminMsg::Logon`. With the split landing here, the `141=Y` encoding lives in #573 and #567 just emits the `LogonReset` variant — that avoids the collision and keeps the flag from being dropped in the dedup.

Abso1ut3Zer0:
Quick update now that #567 is approved: it **already adds the `AdminMsg::LogonReset` variant** (encoding it as a plain `Logon` for now). So this PR no longer needs to split the variant — just **rebase on #567 and route the existing `LogonReset` → `D::encode_logon_reset`** (the method to add in #572), replacing #567's plain-`Logon` placeholder so it emits `141=Y`. That's the piece that makes the in-session reset actually function on the wire.

Effective order is now #567 → #572 + #573.

Abso1ut3Zer0 (approved):

Issue #578, opened 2026-06-30
nexus-fix-engine: FIX conformance suite (Autobahn-analog) — QuickFIX interop + raw-bytes adversarial


Abso1ut3Zer0:
Build out `tools/fix-harness/` (seeded by #569) into a comprehensive FIX session-conformance suite — the FIX analog of the Autobahn suite nexus-web passes (517/517). On-demand for now; CI-gating is a later step once it's fleshed out.

## Two drivers under one Behave catalog

A conformance suite needs both, and they're complementary:

1. **QuickFIX driver (protocol-level, seeded by #569).** A real reference engine drives all *valid* behaviors + protocol edge cases: logon variants (incl. `ResetSeqNumFlag`), heartbeat/test-request timing, resend/gap-fill, sequence reset, large messages, logout races. Strong positive-interop signal a hand-rolled peer can't match.
2. **Raw-bytes driver (adversarial / fuzzing).** QuickFIX is spec-compliant and *cannot* emit malformed input, so the byte-level adversarial layer needs a raw socket driver (the existing `tests/fixtures/fix_peer.py` is the seed). Targets: bad checksum (E3), lying BodyLength (E3), garbage framing (E4), truncated/oversized/duplicated tags, malformed/zero/backward seqnums (E5), unbounded ResendRequest end-seq (E2). These double as regression tests for the E2–E5 hardening.

Design the step/driver layer for this split early so adversarial scenarios drop in without a retrofit.

## Build-out tasks (grown over time)

- [ ] Dedicated harness-server binary — don't overload the `blocking_session` example. Current harness mode is an infinite accept loop that leaks an `acceptor_{n}` tmp dir per connection; replace with a server that reuses/cleans dirs.
- [ ] Real step logic — current `given`/`when` are no-ops; `before_scenario` does the work.
- [ ] QuickFIX positive scenarios: heartbeat, test-request, resend/gap-fill, sequence reset, logout, reconnect-with-reset.
- [ ] Raw-bytes adversarial scenarios for E2–E5 (paired with the hardening fixes).
- [ ] Eventually CI-gate (cache the QuickFIX C++ image so the build isn't paid per run).

Seeded by #569. Related: #540 (hand-rolled adversarial peer), #543 (adversarial conformance), #523 (umbrella).

Abso1ut3Zer0:
Clarification after looking closer at how FIX engines actually handle malformed input, so future adversarial scenarios assert the right behavior:

[...]

PR #579, opened 2026-07-01
feat(fix-harness): dedicated harness-server binary

ChronoCoders:
Adds nexus-fix-harness as a standalone binary crate. Replaces the harness mode in blocking_session (which leaked an acceptor_{n} tmp dir per connection) with a server that resets a single journal dir between sessions.

- New nexus-fix-harness workspace crate with accept loop
- Journal dir wiped and recreated per connection so each Behave scenario starts with fresh sequence numbers
- run.sh updated to cargo run -p nexus-fix-harness

Part of #578.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ members = [
 	"nexus-fix-codegen-tests",
 	"nexus-fix-engine",
 	"nexus-async-fix-engine",
+	"nexus-fix-harness",
 	"nexus-platform",
 ]
 
--- a/nexus-fix-engine/examples/blocking_session.rs
+++ b/nexus-fix-engine/examples/blocking_session.rs
@@ -87,28 +87,15 @@ impl<'buf> FixHeader<'buf> for Fix44Header<'buf> {
 // ── main ─────────────────────────────────────────────────────────────────────
 
 fn main() {
-    let port: u16 = std::env::var("FIX_PORT")
-        .ok()
-        .and_then(|v| v.parse().ok())
-        .unwrap_or(0);
-    let listener = TcpListener::bind(("127.0.0.1", port)).unwrap();
+    let listener = TcpListener::bind(("127.0.0.1", 0)).unwrap();
     let addr = listener.local_addr().unwrap();
     println!("listening on {addr}");
 
-    if port != 0 {
-        let mut n = 0u64;
-        loop {
-            let dir = tmp_dir(&format!("acceptor_{n}"));
-            n += 1;
-            run_acceptor(&listener, &dir);
-        }
-    } else {
-        let acceptor_dir = tmp_dir("acceptor");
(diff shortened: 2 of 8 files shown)


Abso1ut3Zer0:
Nice build-out — this hits several #578 items cleanly: dedicated harness server, real step logic (the `@when` steps drive the initiator now instead of `before_scenario` doing everything), and three genuine QuickFIX scenarios (logout, TestReqID-echo heartbeat, reset-on-reconnect). The per-connection journal reset is the right fix for the tmp leak, and the crate builds clippy-clean.

One thing to finish before merge: the description says this "replaces the harness mode in `blocking_session`," but `blocking_session.rs` isn't in the diff — the leaky `FIX_PORT` branch and the `loop { … acceptor_{n} … }` are still there. Since `nexus-fix-harness` now owns the harness role, please **revert `blocking_session.rs` to its clean demonstrate-usage form** (spawn acceptor + run initiator + join, drop the `FIX_PORT` branch). That's what actually removes the leak the PR set out to kill — right now the old code just lingers, dead.

Minor / optional: `#![cfg(unix)]` at the crate root means the bin compiles to nothing on Windows → "main function not found" → `cargo build --workspace` fails there. macOS is unix so it's fine, and Windows is experimental so this is low priority — but a small `#[cfg(not(unix))] fn main() {}` stub would keep the workspace building everywhere.

Otherwise this is in good shape — solid step toward the Autobahn-style suite.

ChronoCoders:
Both done in 8a06b58. `blocking_session.rs` is back to the spawn+join demo, and the harness crate compiles on Windows -- switched from crate-level `#![cfg(unix)]` to per-item gates with a `#[cfg(not(unix))] fn main() {}` stub.

On0n0k1:
Good changes. I assume the following are needed to be implemented next?

- [ ] Raw-bytes adversarial scenarios for E2–E5 (paired with the hardening fixes).
- [ ] Eventually CI-gate (cache the QuickFIX C++ image so the build isn't paid per run).

There are no tests for Windows needed at the moment, correct?

Abso1ut3Zer0 (approved):
Verified both fixes — `blocking_session.rs` is back to the clean spawn-acceptor + run-initiator + join demo (the `acceptor_{n}` leak is gone), and the `#[cfg(not(unix))] fn main() {}` stub compiles everywhere instead of "main not found." Clippy `-D warnings` clean on harness + engine, and the example builds. Approving.

@On0n0k1 on your question — yes, that's the direction. We'll keep growing the conformance suite (#578), and a big part of that is **additional testing, including the adversarial layer**: the raw-bytes scenarios for E2–E5 (bad checksum, lying BodyLength, garbage framing, malformed/backward seqnums). QuickFIX can't drive those since it only emits spec-valid frames, so that layer needs a raw-frame driver alongside these positive QuickFIX scenarios. So yes — that's next on the roadmap.

PR #582, opened 2026-07-04
feat(fix-harness): raw-bytes adversarial scenarios (E2-E5)

ChronoCoders:
Adds the raw-bytes driver layer to the Behave catalog alongside the existing QuickFIX driver. A `RawPeer` class wraps a raw TCP socket and drives the FIX framing manually, covering input that a spec-compliant engine like QuickFIX can never emit.

Six scenarios across three files:
- `adversarial_steps.py` -- `RawPeer`, builder with `bad_checksum`/`body_len_override`, per-scenario `given`/`when`/`then` steps
- `adversarial.feature` -- E2 (unbounded ResendRequest clamped), E3 (bad checksum, lying BodyLength), E4 (garbage framing), E5 (zero/duplicate seqnum)
- `environment.py` -- adds `raw_peer` lifecycle (init + close in after_scenario)

Part of #578.

### DIFF
--- a/tools/fix-harness/features/adversarial.feature
+++ b/tools/fix-harness/features/adversarial.feature
@@ -1,0 +1,36 @@
+Feature: FIX session adversarial input
+
+  Scenario: bad checksum is rejected (E3)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer performs a Logon handshake
+    And the peer sends a Heartbeat with a bad checksum
+    Then the engine closes the connection
+
+  Scenario: lying BodyLength causes disconnect (E3)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer performs a Logon handshake
+    And the peer sends a Heartbeat with BodyLength 5
+    Then the engine closes the connection
+
+  Scenario: garbage framing is rejected (E4)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer sends garbage bytes
+    Then the engine closes the connection
+
+  Scenario: zero seqnum is rejected (E5)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer performs a Logon handshake
+    And the peer sends a Heartbeat with seqnum 0
+    Then the engine closes the connection
+
+  Scenario: duplicate seqnum is rejected (E5)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer performs a Logon handshake
+    And the peer sends a Heartbeat with seqnum 1
+    Then the engine closes the connection
+
+  Scenario: unbounded ResendRequest is clamped, not crashed (E2)
+    Given a raw FIX 4.4 peer connects to the harness
+    When the peer performs a Logon handshake
+    And the peer sends a ResendRequest with EndSeqNo 0
+    Then the engine replies with SequenceReset
--- a/tools/fix-harness/features/adversarial.feature
+++ b/tools/fix-harness/features/adversarial.feature
@@ -1,21 +1,25 @@
 Feature: FIX session adversarial input
 
-  Scenario: bad checksum is rejected (E3)
+  Scenario: bad checksum is ignored, session survives (E3)
     Given a raw FIX 4.4 peer connects to the harness
     When the peer performs a Logon handshake
     And the peer sends a Heartbeat with a bad checksum
-    Then the engine closes the connection
+    And the peer sends a TestRequest
+    Then the engine replies with Heartbeat
 
-  Scenario: lying BodyLength causes disconnect (E3)
+  Scenario: lying BodyLength is ignored, session survives (E3)
     Given a raw FIX 4.4 peer connects to the harness
     When the peer performs a Logon handshake
     And the peer sends a Heartbeat with BodyLength 5
-    Then the engine closes the connection
+    And the peer sends a TestRequest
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0:
Solid raw-bytes driver — the `RawPeer` (framing, `sum % 256` checksum, BodyLength override) is well-built, and **E2** (unbounded ResendRequest → resend clamped → `SequenceReset`) plus both **E5** scenarios (zero / duplicate seqnum → `SeqNumTooLow` → close) are correct and pass.

The three malformed scenarios need their expectation flipped, though. **E3 bad checksum, E3 lying BodyLength, and E4 garbage all assert the engine disconnects — but per the FIX spec + QuickFIX default, malformed messages are *ignored*, not disconnected** (dropped, inbound seqnum not incremented, wait for the next message; FIX is optimistic). nexus already does this correctly: bad checksum, bad BodyLength, and garbage all resolve to `ParseResult::Garbage` → the transport counts it and resyncs to the next `8=`. Disconnecting there would make nexus stricter than the reference engines.

Rework those three to assert **survival-and-resync** instead of closure:
- send the malformed frame → engine ignores it (no disconnect, seqnum unchanged)
- then send a valid frame (e.g. a TestRequest) → engine still replies, proving it stayed alive and in sync.

That's observable from the raw peer (the internal garbage counter isn't). E2 + E5 are good as-is.

Separately, making the caller able to *see* these malformed frames (they're swallowed today) is tracked in #583 — engine-side and orthogonal to this PR.

ChronoCoders:
Good catch. E3/E4 flipped to survival-and-resync: send the malformed frame, then a TestRequest, assert Heartbeat reply. Also fixed a seqnum bug in the step code -- the malformed sends were auto-incrementing _seq, so the follow-up would arrive one ahead of what the engine expected. They now pass seq explicitly to stay in sync with the engine's view.

Abso1ut3Zer0 (approved):
Ran the full harness against the current engine. The raw-bytes driver is solid: `RawPeer` framing, the `sum % 256` checksum, and the BodyLength override are all correct, and this is the reusable foundation the rest of the adversarial suite builds on.

Three scenarios (E5 zero seqnum, E5 duplicate seqnum, E3 lying BodyLength) are red, but they are correct as written. They fail because of two engine-side bugs, not the harness:

- #584: the transport drops `Event::Disconnected` for the Heartbeat/TestRequest/ResendRequest/SequenceReset/Reject arms, so a too-low seqnum in a Heartbeat never tears the session down. That is why the two E5 scenarios do not see the connection close.
- #585: `validate_checksum` false-accepts a frame with no CheckSum(10), so a lying BodyLength becomes a fatal `MissingMsgSeqNum` instead of a garbled-ignore. That is the E3 lying-BodyLength failure.

Both are maintainer-side and small; the three scenarios will go green when those land, with no change needed here. E2, E3 bad-checksum, and E4 garbage pass and confirm the engine resyncs correctly.

Merging so the driver lands and the suite can grow. Nice catch surfacing these.

Abso1ut3Zer0:
Merged, thanks. To be clear on next steps:

**Do not touch the three red scenarios.** E5 zero seqnum, E5 duplicate seqnum, and E3 lying BodyLength are correct as written. They are red because of two engine-side bugs, #584 (transport drops the disconnect for Heartbeat/TestRequest/etc.) and #585 (`validate_checksum` false-accepts a frame with no CheckSum). Both are maintainer-side and we are picking them up. When they land, those three go green with no change to the harness. So please leave the assertions asserting survive-and-resync / close as they are.

**What to build next:** keep extending the adversarial suite on the `RawPeer` foundation you just landed, tracked under #578. Good candidates the driver can already drive: repeated malformed frames in a row followed by a valid message (proves resync holds under repetition), a partial frame followed by a complete one, and interleaved garbage between two valid messages. Pull the roadmap from #578 rather than expanding scope on your own; ping if you want a specific slice to pick up.

**Heads-up, separate infra issue:** the pre-existing `session.feature` suite currently fails in the container with `MemoryStoreFactory.__init__() takes 1 positional argument but 2 were given`, a quickfix binding API mismatch (`quickfix==1.15.1`). That is unrelated to your adversarial work. If you hit red there when running the full suite, that is why. We can track it separately; no action needed from you on it right now.

Abso1ut3Zer0:
@ChronoCoders here are three the harness surfaced that are ready for you to pick up while I'm out, in suggested order. Each is assigned to you and has an "Implementation notes" comment with exact file:line, a code sketch, scope boundaries, and how to verify.

1. **#586** (harness infra, ~1 line) — `session.feature` currently fails on a quickfix API mismatch; drop the `settings` arg to `MemoryStoreFactory`. Quick warm-up that gets the session suite green again.
2. **#585** (nexus-fix-codec / frame, ~4 lines + a test) — a lying BodyLength should be garbled-and-ignored, not a fatal error. Positional `10=` guard in `try_parse`. Flips the E3 lying-BodyLength scenario green.
3. **#584** (nexus-fix-engine / transport, ~15 lines + a test) — the five admin dispatch arms need to surface `Event::Disconnected` like the Logon/Logout/App arms already do. Flips both E5 scenarios green.

#585 and #584 turn the three red adversarial scenarios from this PR green, so landing them gives the harness its first full bugs-caught-and-fixed loop. Ping on the issue if any of the notes are unclear or you want a different slice.

PR #591, opened 2026-07-12
fix(fix-engine): persist next_inbound across restart

### LINKED ISSUE #577: nexus-fix-engine: persist next_inbound across restart (resume asymmetry)
`FixJournal` journals **outbound only**. On reopen, `recover_from_journal` (`persist.rs:110-124`) restores `next_outbound` by replaying tag 34 from the journal — but `next_inbound` is never persisted: `advance_inbound` / `set_next_inbound` are memory-only and `store` only records outbound messages, so `next_inbound` resets to **1** on every restart.

This is a **resume asymmetry**: after a crash the engine can correctly *answer* a peer's ResendRequest (outbound is durable), but it **cannot correctly request resends itself** — with `next_inbound` back at 1 it would either demand a full resend from seq 1 or mis-detect gaps. For a journal whose purpose is durable resume, inbound and outbound should both survive a restart.

Pre-existing and orthogonal to multi-session; surfaced by #570, whose multi-session test asserts only `next_outbound` independence (so the gap is invisible in CI). Sibling to the E8 "two sources of truth for `next_outbound`" recovery/ownership work in #545.

## Proposed

- Persist `next_inbound` so it recovers across restart (journal an inbound checkpoint, or include inbound state in the recovery scan).
- Add a recovery test asserting `next_inbound` survives reopen.

Refs #570, #545, #518.

ChronoCoders:
Closes #577.

`FixJournal` journaled outbound only. `next_inbound` reset to 1 on every restart because neither `advance_inbound` nor `set_next_inbound` wrote anything to disk, and `recover_from_journal` never read it back.

Two changes:

`persist.rs` -- `set_next_inbound` now appends a compact `NI={seq}` checkpoint record to the journal (no-op when the value is unchanged). `advance_inbound` delegates through it. `recover_from_journal` scans for these records alongside the existing outbound-seqnum scan, taking the last value found. The sentinel prefix `NI=` is unambiguous -- all real FIX messages start with `8=`.

`transport.rs` -- `recv` syncs `journal.set_next_inbound(state.next_inbound_seq())` at its top, before the read loop. This is deferred-by-one: the previous dispatch's advancement is checkpointed at the start of the next call. After a crash the engine is at most one message behind; the peer resends on reconnect, which FIX already handles.

Test: `open_recovers_next_inbound` in `persist::tests` -- advances inbound, reopens the journal, asserts the value survives.

### DIFF
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -3,6 +3,29 @@ use std::path::Path;
 use nexus_fix_codec::{find_tag, parse_fix_seqnum};
 use nexus_journal::{Conductor, Frame, LogOffset, OpenError, RotatingJournal, WriteError};
 
+const INBOUND_MARKER: &[u8] = b"NI=";
+
+fn encode_inbound_checkpoint(seq: u32, buf: &mut [u8; 13]) -> usize {
+    buf[0] = b'N';
+    buf[1] = b'I';
+    buf[2] = b'=';
+    if seq == 0 {
+        buf[3] = b'0';
+        return 4;
+    }
+    let mut tmp = [0u8; 10];
+    let mut len = 0;
+    let mut n = seq;
+    while n > 0 {
+        tmp[len] = b'0' + (n % 10) as u8;
+        len += 1;
+        n /= 10;
+    }
+    tmp[..len].reverse();
+    buf[3..3 + len].copy_from_slice(&tmp[..len]);
+    3 + len
+}
+
 enum ResendPlan<'a> {
     Replay(Frame<'a>),
     GapFill,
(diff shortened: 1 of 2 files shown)


Abso1ut3Zer0 (approved):
Fixes #577 — `next_inbound` now survives restart (it was resetting to 1 every time, which broke session recovery). Verified the resend path can't ship an `NI=` checkpoint to the counterparty: checkpoints go through `journal.append` (not `store`), so they never get a seqnum->offset entry, and `resend_one` reads only via that table with a `find_tag(34) == seq` guard. Strict improvement, merging. We'll refine the checkpoint mechanism (dedicated crash-exact store instead of journal sentinels) in the upcoming cleanup + perf round.

PR #593, opened 2026-07-12
feat(journal): open_existing -- error instead of silently creating a fresh session

### LINKED ISSUE #576: nexus-fix-engine: open-existing-only journal mode (error instead of silently creating a fresh session)
`FixJournal::open(dir, session_id, window)` (and the journal layer's `conductor.session().session_id(id).open()`) is **open-or-create**: if `session_id` has no manifest on disk, it silently creates a fresh session and returns it. A reconnecting caller that passes the wrong `session_id` therefore starts a brand-new journal and **skips all unacknowledged-message replay, with no error**.

Pre-existing behavior (single-session `open()` created fresh too), but #570 sharpens it by making the id caller-supplied — a fat-fingered or off-by-one id now silently loses replay state.

`open_strict` does **not** address this — verified it only errors on a structural *config* mismatch (segment size) when a manifest already exists (`open_inner`, `rotating/mod.rs:129`); an absent manifest still creates fresh either way.

## Proposed

- An **open-existing-only** path that errors (e.g. `OpenError::SessionNotFound { session_id }`) when the session's manifest is absent — for reconnect callers that must recover, not create.
- Keep open-or-create for callers that genuinely want "open or start fresh."
- Until then, document the open-or-create semantics on `FixJournal::open` so the footgun is visible (the #570 doc fix should say this).

Refs #570, #518.

ChronoCoders:
Closes #576.

`RotatingJournalBuilder::open` (and `FixJournal::open`) is open-or-create: a wrong session_id silently starts a fresh journal and drops all replay state. `open_strict` doesn't help -- it only guards against a structural config mismatch when a manifest already exists; an absent manifest still creates fresh either way.

New `open_existing` method on `RotatingJournalBuilder` checks whether the manifest exists before creating directories or acquiring the session lock, and returns `OpenError::SessionNotFound { session_id }` if it doesn't. The happy path falls through to the same `open_inner(false)` as `open`. `FixJournal::open_existing` mirrors it.

Tests:
- typo id on an empty directory: `SessionNotFound`
- wrong id when another session exists: `SessionNotFound`
- correct id: recovers and reads back previously written data
- all three at both the `nexus-journal` and `nexus-fix-engine` layers

### DIFF
--- a/nexus-fix-engine/src/persist.rs
+++ b/nexus-fix-engine/src/persist.rs
@@ -105,6 +105,33 @@ impl FixJournal {
         Ok(this)
     }
 
+    /// Open an existing session journal, returning [`OpenError::SessionNotFound`]
+    /// when no manifest exists for `session_id`.
+    ///
+    /// Unlike [`open`](Self::open) this never creates a new session. Use for
+    /// reconnect callers that must recover state — a wrong `session_id` is an
+    /// error rather than a silent fresh start.
+    pub fn open_existing(
+        dir: impl AsRef<Path>,
+        session_id: u32,
+        window: usize,
+    ) -> Result<Self, OpenError> {
+        assert!(window.is_power_of_two());
+        let mut conductor = Conductor::open(dir)?;
+        let journal: RotatingJournal =
+            conductor.session().session_id(session_id).open_existing()?;
+        let mut this = Self {
+            journal,
+            _conductor: conductor,
+            offsets: vec![None; window].into_boxed_slice(),
+            window,
+            next_outbound: 1,
+            next_inbound: 1,
+        };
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0 (approved):
Good failure-mode fix — a wrong `session_id` silently started a fresh journal and dropped replay state; now it returns `SessionNotFound` after checking the manifest before any dir/lock side effects. Well-tested at both the nexus-journal and FixJournal layers. Merging. Non-blocking follow-up we'll take in an upcoming cleanup round: `open`/`open_strict`/`open_existing` is an incomplete 2x2 (create-vs-error x check-vs-nocheck); we'll consolidate to explicit flags.

PR #600, opened 2026-07-17
Venue auth: SessionCustomizer hook for outbound admin messages (#568)

Abso1ut3Zer0:
## Summary

Closes the venue-auth gap in #568: a `SessionCustomizer` hook that lets a venue inject Logon credentials (`Username 553`, `Password 554`, `RawData 96`) into outbound admin messages, **including HMAC/Ed25519 signatures computed over the engine-stamped header**. This is the last thing standing between the engine and logging onto a real crypto venue.

**No code generation.** Plain FIX is untouched: `FixConnection<TcpStream, Fix44>` and every existing call site compile and behave identically, and `NoCustomizer` is a ZST that monomorphizes away.

## Why a hook, and explicitly not codegen

The obvious-looking design (derive a typed `Config` from the Logon schema via `required="Y"`, wire it into the generated typestate encoders) **does not work**. Venue auth is a *computed function of engine-owned per-message state*, not a static string:

- **Coinbase**: HMAC-SHA256 over SOH-joined `52 | 35 | 34 | 49 | 56 | 554` produces `RawData(96)`
- **Binance**: **Ed25519** (HMAC-SHA256 and RSA explicitly unsupported) over `35|49|56|34|52`
- **Deribit**: `RawData(96)` is `timestamp.base64(nonce)` with a >=32-byte CSPRNG nonce that fails on reuse; `Password(554)` is `base64(sha256(RawData ++ secret))`

Every one is uncomputable before the engine assigns 34/52, so a generated config could hold the static secret but never the credential itself. Codegen would add a mechanism **without removing one**.

The decisive evidence is that QuickFIX already ran this experiment: it ships a declarative `LogonTag` config (plain settings strings, no codegen), and `setLogonTags` runs *before* `initializeHeader`, so 34 and 52 do not yet exist and a signature over them is literally uncomputable there. Full reasoning and sources in the [#568 comment](https://github.com/Abso1ut3Zer0/nexus/issues/568).

## Design

The seam is invariant across QuickFIX/J, QuickFIX/n, Artio and Philadelphia: **stamp header, run hook, frame, persist, send.**

### 1. Split the frame seam

`FixDictionary::encode_*` went from one-shot (`FrameFormatter::new`, fields, `finish()`) to writing only its standard fields into a **caller-owned** frame:

```rust
fn encode_logon(fmt: &mut FrameFormatter<'_>, hdr: &AdminHeader<'_>, heart_bt_int_s: u32)
```

The engine now owns the frame lifecycle and brackets the dictionary:

```rust
let mut fmt = FrameFormatter::new(spare, D::BEGIN_STRING, b"A");          // 8, 35
D::encode_logon(&mut fmt, &hdr, heart_bt_int_s);                          // 34/49/56/52, 108
customizer.configure_logon(&mut AdminMsgOut::new(&mut fmt, &hdr, b"A"));  // hook
fmt.finish().ok()                                                          // 9, 10 over everything
```

This is what keeps the dictionary customizer-agnostic: no `Config` associated type, no unit config, no engine-owned-tag classification problem. Nothing overrode `encode_*` (codegen emits only associated types, `BEGIN_STRING` and `is_admin`), so the change is contained.

### 2. `SessionCustomizer`

```rust
pub trait SessionCustomizer {
    /// Fires after the engine stamps 8/35/34/49/56/52, before 9/10 are framed.
    fn configure_logon(&self, m: &mut AdminMsgOut<'_, '_>) { let _ = m; }
    // one per admin message type, all defaulted no-ops
}

pub struct NoCustomizer;
impl SessionCustomizer for NoCustomizer {}
```

**Per-message-type methods** (Artio's shape), not one undifferentiated `toAdmin`. QuickFIX fires its single hook for all seven admin types *including PossDup GapFills*, so every app must branch on MsgType itself or leak credentials into Heartbeats. Defaulted no-ops make that miswiring unrepresentable.

### 3. `AdminMsgOut`

Not append-only: the hook must **read** the stamped header in order to sign over it.

```rust
pub fn seq_num(&self) -> u32;          // 34
pub fn sending_time(&self) -> &[u8];   // 52
pub fn sender(&self) -> &[u8];         // 49
pub fn target(&self) -> &[u8];         // 56
pub fn msg_type(&self) -> &[u8];       // 35
pub fn field(&mut self, tag: u32, value: &[u8]);   // debug_assert on engine-owned tags
```

`debug_assert`, **not** enforcement. QuickFIX has zero enforcement here and two decades of production venue connectivity; the customizer is expert-authored once per venue; and the failure is immediate and loud (a duplicate 34 means the Logon is rejected on first connect, never reaching a resend). `assert!` would be wrong: it fires in release for a shouldn't-happen check.

### Division of labor

| | Source | Scope |
|---|---|---|
| Dictionary (decoders, MsgType, BEGIN_STRING, is_admin) | generated from FIX XML | per-**dictionary** |
| `SessionCustomizer` (auth logic) | hand-written, ~20 lines | per-**venue** |

This is what Artio validated: it code-generates `LogonEncoder` from XML yet hand-writes its entire customisation surface. Codegen can see that tag 96 is `required="Y"`; it cannot write an Ed25519 signature. Auth is code, not a schema shape.

## Testing

21 new tests (17 engine, 4 codec): byte-identity oracles for all eight admin types on the `NoCustomizer` path, injected 553/554 on the wire, 9/10 correctness over injected fields, accessor-driven signing, per-message dispatch, and a `#[should_panic]` on `field(34, ..)`.

Every load-bearing test was **fault-injected to prove it has teeth** (each fault reverted after):

| Fault | Caught by | Symptom |
|---|---|---|
| `body_len - 25` in `finish()` | 9/10 correctness | BodyLength 61 vs 86 |
| `seq_num() + 1` | accessor signing | digest mismatch |
| 108 written before the header | byte-identity | full byte diff |
| SequenceReset arm wired to `configure_reject` | per-message dispatch | "got reject, want sequence_reset" |

**No existing test needed editing**, which is the byte-identity guarantee holding across the frame-seam split.

## Deliberate decision: credentials are journaled

`store_admin` journals the encoded bytes, so injected credentials (including Coinbase's plaintext `Password(554)`) land in the outbound journal. **Accepted on purpose**: the journal is local to the box, and an archive that matches the wire byte-for-byte is worth more than a redaction step that would make it lie. QuickFIX/J has the same property.

Exposure is limited to the original Logon: resend gap-fills reframe directly and never run the hook, so PossDup traffic cannot carry credentials.

## Follow-ups

- `AdminMsg::Heartbeat`'s `echo` is `Option<([u8; 64], u8)>` while `TEST_REQ_ID_CAP` is private, so an external customizer must hardcode `64`. Pre-existing; worth exporting the constant.
- **#544** (nexus-net wire seam + TLS, plus the async backpressure yield) is the last structural piece after this.

### DIFF
--- a/nexus-async-fix-engine/CHANGELOG.md
+++ b/nexus-async-fix-engine/CHANGELOG.md
@@ -10,6 +10,18 @@ contained.
 
 ## [Unreleased]
 
+### Added
+
+- Venue Logon auth, mirroring the sync engine. `FixConnection` and
+  `FixConnectionBuilder` take a per-venue `SessionCustomizer` (from
+  `nexus-fix-codec`) as a trailing type parameter defaulting to `NoCustomizer`,
+  so existing call sites — `FixConnection<TcpStream, Fix44>`,
+  `FixConnection::from_parts(...)`, `FixConnection::builder()` — are unchanged.
+
+  Attach one with `FixConnectionBuilder::customizer(c)` or
+  `FixConnection::from_parts_with_customizer(...)`. All hook behavior lives in
+  the shared `FixSession` core; this crate only threads the type parameter.
+
 ### Internal
 
 - Test scratch directories are now removed on drop. The `tmp_dir` helpers in
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -21,7 +21,7 @@ use std::io;
 use std::marker::PhantomData;
 use std::time::{Duration, Instant};
 
-use nexus_fix_codec::FixDictionary;
+use nexus_fix_codec::{FixDictionary, NoCustomizer, SessionCustomizer};
 use nexus_fix_engine::{
     DisconnectReason, FixJournal, FixSession, FrameReader, FrameWriter, Message, MessageReader,
     MessageWriter, PollOutcome, SessionConfig, SessionError, SessionState, State,
@@ -43,21 +43,27 @@ pub use nexus_fix_engine::TransportError as Error;
 /// else — framing, the [`SessionState`] machine, journaling, resend — is in the
 /// shared core. This is the async twin of
 /// [`nexus_fix_engine::FixConnection`].
-pub struct FixConnection<S, D: FixDictionary> {
+///
+/// `C` is the per-venue [`SessionCustomizer`] applied to outbound admin
+/// messages, defaulting to [`NoCustomizer`] — plain-FIX callers write
+/// `FixConnection<TcpStream, Fix44>` and never name it. Attach one with
+/// [`FixConnectionBuilder::customizer`].
+pub struct FixConnection<S, D: FixDictionary, C = NoCustomizer> {
     stream: S,
-    session: FixSession<D>,
+    session: FixSession<D, C>,
 }
 
 /// Builder for [`FixConnection`], mirroring the sync
 /// [`FixConnectionBuilder`](nexus_fix_engine::FixConnectionBuilder).
-pub struct FixConnectionBuilder<D: FixDictionary> {
+pub struct FixConnectionBuilder<D: FixDictionary, C = NoCustomizer> {
     reader_cap: usize,
     writer_cap: usize,
     nodelay: bool,
+    customizer: C,
     _dict: PhantomData<fn() -> D>,
 }
 
-impl<D: FixDictionary> FixConnectionBuilder<D> {
+impl<D: FixDictionary, C> FixConnectionBuilder<D, C> {
     pub fn reader_capacity(mut self, n: usize) -> Self {
         self.reader_cap = n;
         self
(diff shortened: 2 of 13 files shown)


Abso1ut3Zer0:
### Follow-up planned: typed admin-emit seam

This PR ships the venue-auth hook and cleaned up the encode path (removed `AdminKind`; owned-tag lists now live on the dictionary next to each encoder, guarded by a drift test; `customize_*` naming). The `AdminMsg` enum and the `encode_admin` match stay as-is here and work correctly.

A separate follow-up will replace that seam with typed dispatch: each admin message becomes a struct implementing an `AdminEncode` trait, the state machine emits through an `AdminSink` (one generic method, which is the only reason it needs to be a trait rather than a closure), and journaling stays an injected closure on the emitter. The point is reducing state entropy: the concrete message type is never widened into a sum and matched back apart, and cross-message mis-wiring (an arm calling the wrong `customize_*` or encoder) becomes unrepresentable instead of test-guarded. `Control` stays an enum, since it is a genuine multi-outcome verdict; only the dispatch-vocabulary enum goes.

Carries forward unchanged: the `customize_*` names, the dictionary-sourced owned lists, and the drift test. No codegen. It reopens the emit boundary and touches every handler, so it is its own PR, likely paired with #544 (the nexus-net wire seam), which reworks the same driver boundary.

PR #610, opened 2026-07-18
nexus-fix-log: read-only FIX journal log viewer

ChronoCoders:
New binary crate. Attaches to a journal directory in read-only mode (no conductor, no session lock) and prints committed FIX messages to stdout.

- Discovers sessions by scanning subdirs for `journal.manifest`
- Reads segment files directly; respects the commit_len sentinel (zero = uncommitted, safe to stop)
- Skips `NI=` checkpoint records written by the fix-engine
- Decodes each message with `FieldReader` and prints `tag=value` pairs
- Args: `<journal-dir> [--session-id <N>]`

Output:

```
[session=0 +0]       8=FIX.4.2 | 35=A | 34=1 | 49=NEXUS | 56=CLIENT | 52=20240101-12:00:00 | 98=0 | 108=30 | 10=123
[session=0 +128]     8=FIX.4.2 | 35=0 | 34=2 | 49=NEXUS | 56=CLIENT | 52=20240101-12:00:30 | 112=TESTID | 10=045
```

No live-tail in this cut; static snapshot at invocation time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ members = [
 	"nexus-inference",
 	"nexus-fix-codec",
 	"nexus-fix-codegen",
+	"nexus-fix-log",
 	"nexus-fix-codegen-tests",
 	"nexus-fix-engine",
 	"nexus-async-fix-engine",
--- a/nexus-fix-log/Cargo.toml
+++ b/nexus-fix-log/Cargo.toml
@@ -6,6 +6,7 @@ rust-version.workspace = true
 license.workspace = true
 description = "Read-only FIX journal log viewer"
 repository.workspace = true
+publish = false
 
 [lints]
 workspace = true
--- a/nexus-fix-log/Cargo.toml
+++ b/nexus-fix-log/Cargo.toml
@@ -13,3 +13,6 @@ workspace = true
 
 [dependencies]
 nexus-fix-codec = { path = "../nexus-fix-codec" }
+
+[dev-dependencies]
+nexus-journal = { path = "../nexus-journal" }
--- a/nexus-fix-log/Cargo.toml
+++ b/nexus-fix-log/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "nexus-fix-log"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+description = "Read-only FIX journal log viewer"
+repository.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+nexus-fix-codec = { path = "../nexus-fix-codec" }
--- a/nexus-fix-log/src/main.rs
+++ b/nexus-fix-log/src/main.rs
@@ -1,15 +1,31 @@
 use std::env;
 use std::fs;
+use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::process;
 
 use nexus_fix_codec::reader::FieldReader;
 
 const MAGIC: u32 = u32::from_le_bytes(*b"NXLG");
+const VERSION: u16 = 1;
+// 96-byte original layout + 8-byte meta slot appended in the same struct.
+const MANIFEST_MIN_LEN: usize = 104;
 const FRAME_HDR: usize = 8;
 const ALIGN: usize = 8;
 // Each stored frame's payload is [ts:8 LE UNIX-nanos][FIX wire message].
 const TS_LEN: usize = 8;
+
+// FIX length/data field tag pairs: (length-tag, data-tag).
+// A length tag's value is the exact byte count of the paired data field,
+// which may legally contain embedded SOH and '=' characters.
+const DATA_PAIRS: &[(u32, u32)] = &[
+    (90, 91),   // SecureDataLen / SecureData
+    (95, 96),   // RawDataLength / RawData
+    (212, 213), // XmlDataLen / XmlData
+    (348, 349), // EncodedIssuerLen / EncodedIssuer
+    (358, 359), // EncodedSecurityDescLen / EncodedSecurityDesc
+    (360, 361), // EncodedTextLen / EncodedText
+];
 
 struct SessionMeta {
     // Conductor-level session id as stored in the manifest.
@@ -30,11 +46,15 @@ impl SessionMeta {
 
 fn read_manifest(path: &Path) -> Option<SessionMeta> {
     let data = fs::read(path).ok()?;
-    if data.len() < 96 {
+    if data.len() < MANIFEST_MIN_LEN {
         return None;
     }
     let magic = u32::from_le_bytes(data[80..84].try_into().ok()?);
     if magic != MAGIC {
+        return None;
+    }
+    let version = u16::from_le_bytes(data[88..90].try_into().ok()?);
+    if version != VERSION {
         return None;
     }
     let segment_size = u64::from_le_bytes(data[64..72].try_into().ok()?) as usize;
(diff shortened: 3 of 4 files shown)


Abso1ut3Zer0 (changes requested):
Good call on the approach — reading the segments directly is the right move. I checked and
an active session holds `session.lock`, so the obvious route (open a `RotatingJournal` and
use `read_start`/`read_next`) can't work for a viewer by construction. Bypassing the API is
justified here.

I built it and ran it against real journals written through `FixJournal` rather than reading
the format off. The decoding is right where it counts:

- manifest offsets and the `[ts:8 LE][wire]` payload match `manifest.rs` / `persist.rs:281`
- the frame walk (`commit_len = len + 1`, 8-byte header, 8-align) matches `rotating/frame.rs`
- `conductor_id / 2` + even/odd → session/direction matches the two-journal split
- **rotation verified at `epoch=2`** — the slot-wraparound case. Printed exactly prev+current,
  strictly contiguous seqnums, nothing leaked from the recycled standby slot. I probed the
  segments independently and it agreed frame-for-frame.
- reads a live journal while a writer holds the lock, which is the whole point

So the hard part is done. This isn't going to release yet, so I'm not going to gate it on
polish — but there's one theme I do want closed: **there are three places where the tool
shows less than what's on disk without telling you.** A viewer that quietly under-reports is
worse than no viewer, because you'll trust it during an incident.

## Must fix

**Messages containing DATA fields get silently truncated.** `print_message` drives the bare
`FieldReader` iterator, which is a pure SOH scanner. With a legal embedded SOH in `96=RawData`:

```
on disk:  8=FIX.4.2|35=n|34=1|95=5|96=AB<SOH>CD|55=BTCUSD|10=999
rendered: 8=FIX.4.2 | 35=n | 34=1 | 95=5 | 96=AB |
```

`96` is cut at the embedded SOH, `next_field` then hits `CD\x01` with no `=`, returns `None`,
and the iterator ends. `55` and `10` are gone with no marker and exit 0 — indistinguishable
from a genuinely short message.

We already solved this: `FieldReader::next_data_field(len)` shipped in #501 (the Option B
escape hatch from #429). `next_data_field_embedded_soh` at `reader.rs:914` is exactly this
case, and `next_field()` resumes correctly at the following tag afterward. Switch the `for`
loop to a `while` over the pair — roughly:

```rust
// Length/Data pairs: the length gives the exact byte count of the value
// that follows, which may legally contain SOH and '='.
const DATA_PAIRS: &[(u32, u32)] = &[
    (90, 91), (95, 96), (212, 213),          // SecureData, RawData, XmlData
    (348, 349), (358, 359), (360, 361),      // Encoded* family
];

let mut r = FieldReader::new(fix_msg, 0);
let mut pending: Option<usize> = None;
loop {
    let field = match pending.take() {
        Some(n) => match r.next_data_field(n) { Ok(f) => f, Err(_) => break },
        None    => match r.next_field()       { Some(f) => f, None => break },
    };
    // ... render field ...
    if DATA_PAIRS.iter().any(|(l, _)| *l == field.tag) {
        pending = std::str::from_utf8(field.value.slice(fix_msg))
            .ok().and_then(|s| s.parse().ok());
    }
}
```

Keep a backstop for the cases the table won't cover (a venue's non-standard data field, a
malformed length): `r.pos()` tells you where the reader stopped, so if it's short of the
buffer end, mark the line rather than ending it silently. The rule I care about is that a
cut-off render never passes as a complete message.

**A bad `--session-id` silently prints every session.** `args.get(i).and_then(|s| s.parse().ok())`
swallows both a non-numeric value and a missing one:

```
$ nexus-fix-log <dir> --session-id notanumber   # printed all sessions, exit 0
$ nexus-fix-log <dir> --session-id              # same
```

On a multi-session journal you read another session's traffic believing it's the one you
named. Both cases should be an error to stderr and a non-zero exit.

**An unreadable segment reports an empty session and exits 0.** `let Ok(data) = fs::read(path)
else { return }` — with `chmod 000` on `0/seg0.dat` the outbound header prints with nothing
under it, exit 0. "No outbound messages" and "I couldn't open the file" render identically.
During an incident that's the difference between *we never sent it* and *you can't see what
we sent*. Report the path and the io error; exit non-zero if any segment failed.

## Worth doing while you're in here

**Broken pipe panics.** `nexus-fix-log <dir> | head -3` → `failed printing to stdout: Broken
pipe`, exit 101. `| head` / `| grep` / `| less` is how anyone uses a log viewer. Writing
through `BufWriter::new(io::stdout().lock())` and treating `BrokenPipe` as a clean exit fixes
this and the per-field `print!` cost together (~121 ms for 120k messages, release, pinned —
essentially all stdout-lock plus a flush per record).

**A round-trip test.** `read_manifest` hardcodes byte offsets against `ManifestHeader`, which
is `pub(crate)` in nexus-journal — no compile-time link between them. If that struct gets
reordered this tool won't fail to build and won't error; it'll misparse into plausible-looking
output and CI won't notice. Write N messages via `FixJournal`, assert the viewer reads back
exactly those seqnums, include a rotation. ~40 lines, and it turns silent drift into a red run.

Related and nearly free: `read_manifest` checks `magic` but ignores `version` at offset 88.
`Manifest::open` rejects `version != 1` with `UnsupportedLayout` — I flipped a manifest to
version 2 and the viewer parsed it happily. That check is what makes a future layout change
safe. (Also `data.len() < 96` is stale — the header is 104 bytes since the `meta` slot.)

**Set `publish = false`.** Workspace `release.toml` has `publish = true` and this isn't going
to crates.io — worth making that explicit rather than relying on nobody running it.

## Non-blocking

- A second positional arg silently wins: `nexus-fix-log good-dir bad-dir` reports on `bad-dir`.
- Payloads ≤ 8 bytes are silently skipped (`payload.len() > TS_LEN`). Not reachable via
  `FixJournal`, but it's a corrupt-frame path that currently vanishes — worth counting.
- Non-UTF8 values collapse to a single `?`, discarding the value. Hex-escaping would preserve
  it, and SecureData/RawData are exactly the fields most likely to be binary *and* matter.
- Session headers go to stderr while records go to stdout, so piping stdout drops the
  boundaries and terminal interleaving is nondeterministic.

## Two notes

Worth rebasing on main — #596 (the FIX journal rework) landed after this branch and replaced
the `NI=` checkpoint sentinels with the manifest meta-slot, so the "skips `NI=` checkpoint
records" line in the description no longer applies. Your code already handles the post-#596
two-journal layout correctly, so I think it's just the description that's stale, but a rebase
would confirm nothing else drifted underneath you.

Longer term I'd rather this crate not own a hand-copied binary layout at all — the better
shape is a read-only open in nexus-journal that reuses the existing `read_start`/`read_next`
walk, which today only refuses because it demands an exclusive session. Filed that as #620 —
not this PR's job, but once it lands this crate's hand-copied layout can go away.

Solid foundation — the parts that would have been expensive to get wrong are right.

ChronoCoders:
Addressed all the must-fix items and everything in the "worth doing" section. DATA field pairs now go through next_data_field with a backstop marker when the reader stops short of the buffer end. Bad or missing --session-id values exit non-zero. Unreadable segments print the path and error to stderr and set a non-zero exit. Stdout is now a single locked BufWriter with BrokenPipe treated as a clean exit. Manifest minimum size updated to 104 bytes and the version field is now checked. publish = false added.

Also added an integration test in tests/round_trip.rs that writes 4 outbound and 1 inbound message through a small-segment journal to force a rotation at epoch=1, then runs the binary against the journal directory and asserts every seqnum appears in the output. Any drift in the manifest byte offsets will turn into a red run.

Left the non-blocking items open for a follow-up: the second positional arg, the payloads <= 8 bytes counter, and the UTF-8 hex-escape. Will rebase on main once you confirm the rest looks good.

ChronoCoders:
All items addressed -- ready for re-review.

Abso1ut3Zer0 (approved):

Issue #613, opened 2026-07-19
nexus-fix-engine: recv should block; add try_recv / recv_timeout (drop the Option-from-recv hybrid)


Abso1ut3Zer0:
## Problem

`recv()` returns `Result<Option<Message>, Error>` — named like a blocking receive
but returning `Option` like a non-blocking one. That's the hybrid that's neither:
`Ok(None)` means "nothing ready right now," which is exactly what a `try_recv`
should say, not a `recv`.

The channel convention (std::sync::mpsc, crossbeam, tokio) is the target: `recv`
blocks (or awaits) and yields the item or an error; `try_recv` is the non-blocking
variant that reports "empty."

## Proposal

**Sync**
- `recv() -> Result<Message, Error>` — **blocks** until a frame (or error/EOF).
  A naked blocking read; it does **not** service the heartbeat clock (no crate
  does — tungstenite doesn't). Never returns `None`.
- `try_recv() -> Result<Option<Message>, Error>` — non-blocking; `None` = nothing
  ready.
- `recv_timeout(deadline) -> Result<Option<Message>, Error>` — block up to a
  deadline. This is the wake-and-tick primitive: a caller that must service
  heartbeats/TestRequests loops `recv_timeout` and runs the protocol clock on each
  wake, rather than the library blocking on the socket forever.

**Async**
- `recv().await -> Result<Message, Error>` — awaits; `Poll::Pending` is the
  "nothing yet" signal, so no `None` in the type.
- `try_recv() -> Result<Option<Message>, Error>` — non-blocking drain of what's
  already buffered.

Heartbeat servicing is decoupled from `recv` — the caller's event loop (or a
separate `tick(now)`) runs the protocol clock. This fits the sans-IO model: the
core is poll-driven, and blocking is a thin convenience on top for callers who
want it.

## Relationship

Third of a three-step recv-contract cleanup: (1) #583 malformed → recoverable
error [done], (2) abnormal disconnect → error / clean logout stays a Message,
(3) this. Best done last, once the `Message`/`Error` taxonomy from (1) and (2) is
settled.

PR #617, opened 2026-07-19
FIX engine: split recv into pure receive + explicit clock servicing

Abso1ut3Zer0:
Split the recv contract into **pure receiving** and **explicit clock servicing**, so
no method conflates the two. Receiving a message and enforcing session timers are
separate concerns; folding a heartbeat timer into `recv` made the control flow
implicit and the cancellation story murky. The caller now drives the clock on its own
schedule — a `select!` timer branch in async, a bounded `try_recv` loop (or a separate
thread) in sync.

## New surface

| Method | Behavior |
|---|---|
| `recv(now) -> Result<Message, Error>` | **Blocks** until a message. Never returns `None` (was `Result<Option<Message>>`). |
| `try_recv(now) -> Result<Option<Message>, Error>` | Non-blocking. `None` when nothing is ready. |
| `tick(now) -> Result<(), Error>` | Sends any due Heartbeat/TestRequest and enforces the logon/logout/test-request/reset deadlines. `Err(UnexpectedDisconnect)` when one blows. |
| `next_timeout() -> Option<Instant>` | When `tick` next has work — sizes a `select!` sleep or a bounded `try_recv` wait. |

**No `recv*` method touches the clock.** `recv_timeout` is intentionally gone: a
bounded wait is `try_recv` on a read-timeout socket (`None` = the timeout elapsed), or
`tokio::time::timeout(dur, conn.recv(now))` in async.

## Which blocking behavior you get is the caller's socket mode

`recv`/`try_recv` don't hard-code blocking vs non-blocking — the socket does, the same
contract as tungstenite:

- **Blocking socket** → `recv` parks in the read syscall until a message. A read
  timeout (`SO_RCVTIMEO`) is transparent to `recv` (a no-data wake is retried), so one
  read-timeout socket backs both `recv` (blocks through) and `try_recv` (returns on the
  timeout).
- **Non-blocking socket** → use `try_recv`. Handing one to `recv` is a documented
  misuse (the retry has nothing to wait on and busy-spins).

## Async: no more internal timer on the receive path

The async `recv` loses its internal `tokio::time::timeout_at` — the receive path is now
driven only by the executor + the `WireStream`. That removes the last `tokio::time`
coupling from receiving, so the core stays runtime-agnostic over any `WireStream`
(kernel-bypass transports included); timers become a `select!` branch the caller owns.

**Cancellation caveat (documented on `recv`):** a dropped `recv` future is safe at the
socket read and for any message that carries no reply. A message that emits an outbound
reply (a TestRequest ack, a resend, an acceptor Logon, the Logout ack) flushes that
reply before returning, so a future dropped *during that write* — only reachable under
write back-pressure — can drop the just-processed message. Draining the reply *after*
the poll is required so a terminal ack reaches the wire before the caller's loop ends;
draining before the poll deadlocks the logout. Application messages are always
cancel-safe.

## Tests

The conformance `drive` helpers are rewritten as reactor loops — `tokio::select!`
(`recv` vs `sleep_until(next_timeout)` → `tick`) in async, `try_recv` + `tick` against a
50 ms read-timeout socket in sync. Besides re-establishing the timeout scenarios that
used to lean on `recv`'s internal timer, they double as the reference caller loop for
the new API.

## Verification

`cargo fmt --check`, `clippy --all-features --all-targets -D warnings`, full test suites
for both engines (0 failures), `cargo doc`, and the behave adversarial harness (9/9) all
green.

### DIFF
--- a/nexus-async-fix-engine/CHANGELOG.md
+++ b/nexus-async-fix-engine/CHANGELOG.md
@@ -43,6 +43,13 @@ contained.
   terminal outcome is `Err(TransportError::Closed)`. `Message::Disconnected` is
   removed. (#612)
 
+- `recv`/`try_recv`/`tick`/`next_timeout` mirror the sync engine. `recv(now).await`
+  now yields `Message` directly (was `Option<Message>`) and no longer services
+  heartbeats internally — drive `tick` from a `select!` timer branch. This removes
+  the crate's internal `tokio::time` timer from the receive path, so the core is
+  driven only by the executor + the `WireStream`. There is no async `recv_timeout`;
+  compose `tokio::time::timeout(dur, conn.recv(now))`. (#613)
+
 ### Internal
 
 - Test scratch directories are now removed on drop. The `tmp_dir` helpers in
--- a/nexus-async-fix-engine/src/lib.rs
+++ b/nexus-async-fix-engine/src/lib.rs
@@ -345,13 +345,14 @@ impl<S: WireStream + Unpin, D: FixDictionary, C: SessionCustomizer> FixConnectio
     /// After any terminal outcome further calls return `Err(Closed)`. A malformed
     /// frame is a recoverable `Err(Malformed)` — the session stays live.
     ///
-    /// `now` is your duty-cycle clock reading. The core never reads the wall clock
-    /// itself — that is what keeps the session deterministically replayable — so read
-    /// `now` once per iteration, keep it monotonic, and pass the same value to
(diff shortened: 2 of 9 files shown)


Abso1ut3Zer0:
Superseded by #619.

This branch scoped the recv-naming split for #613. That work is now just phase 1 (clock-free session) of the broader framework reposition tracked in #619, which deletes the clock-servicing API this PR was refining (`now` on recv, `tick`, `next_timeout`). Reopening as phased work off `main` — see #619 for the settled contract. Closing.

PR #618, opened 2026-07-19
nexus-notify: partial-drain cursor for LocalNotify::poll_limit

### LINKED ISSUE #176: nexus-notify: partial-drain cursor for poll_limit
## Summary

`LocalNotify::poll_limit` currently uses `Vec::drain` which is O(remaining) — physical shift of the tail on every partial drain. At 4096 backlog with limit 32, costs 500-1000cy per partial drain after L2/L3 traffic and memmove prologue.

## Fix

Add `dispatch_head: usize` cursor to `LocalNotify`:
- Partial drain advances cursor (no shift)
- Full drain resets cursor and clears list
- Accessors subtract cursor

~15 lines, no contract change.

## Correction to Phase 1

Sync `Poller::poll_limit` in `event_queue.rs` does **NOT** need the same fix — already uses a proper ring-buffer consumer with internal head counter. No physical shift.

## Invariants verified

- `mark`/`poll` mutually exclusive via `&mut self`
- Cursor bounded
- FIFO preserved
- Stale entries are inert (dedup bits cleared, re-marks re-push to tail)

## References

See `.claude/scratch/deep-dive-notify-cursor.md` for full design.
See `.claude/perf-audit-2026-04-20.md` § "Bundle 2".

On0n0k1:
Closes #176

Summary
`LocalNotify::poll_limit` used `Vec::drain` on partial drains, which physically shifts all remaining elements left in memory — O(remaining). At 4096 backlog with a limit of 32, this costs 500–1000 cycles per call due to the memmove prologue.

Changes
nexus-notify/src/local.rs

- **Added `dispatch_head:** usize` cursor field to LocalNotify
- **poll_limit:** partial drain advances the cursor instead of calling `Vec::drain`; full drain resets cursor and clears the list as before
- **has_notified:** checks dispatch_head < dispatch_list.len() instead of is_empty()
- **notified_count:** returns dispatch_list.len() - dispatch_head
- No callers outside local.rs required changes — the public API contract is unchanged.

### DIFF
--- a/nexus-notify/src/local.rs
+++ b/nexus-notify/src/local.rs
@@ -87,7 +87,7 @@ impl LocalNotify {
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
             bits: vec![0u64; capacity.div_ceil(64)],
-            dispatch_list: Vec::with_capacity(capacity),
+            dispatch_list: Vec::with_capacity(capacity * 2),
             dispatch_head: 0,
             num_tokens: 0,
         }
@@ -147,8 +147,23 @@ impl LocalNotify {
         // token — register() grows it on allocation.
         if self.bits[word] & bit == 0 {
             self.bits[word] |= bit;
+            if self.dispatch_list.len() == self.dispatch_list.capacity() {
+                self.compact();
+            }
             self.dispatch_list.push(idx);
         }
+    }
+
+    /// Reclaim the stale prefix by moving live entries to the front.
+    ///
+    /// Called when the tail is out of room. Amortized: fires once per
+    /// `capacity / limit` ticks under sustained partial drain.
+    #[cold]
+    fn compact(&mut self) {
+        let live = self.dispatch_list.len() - self.dispatch_head;
+        self.dispatch_list.copy_within(self.dispatch_head.., 0);
+        self.dispatch_list.truncate(live);
+        self.dispatch_head = 0;
     }
 
     /// Drain all marked tokens into the events buffer.
@@ -443,5 +458,34 @@ mod tests {
         notify.register();
         notify.register();
         assert_eq!(notify.capacity(), 2);
+    }
+
+    #[test]
+    fn poll_limit_partial_sustained_physical_bound() {
+        let n = 128usize;
+        let limit = 32usize;
+        let mut notify = LocalNotify::with_capacity(n);
+        let mut events = Events::with_capacity(n);
+
+        let mut tokens = Vec::new();
+        for _ in 0..n {
+            tokens.push(notify.register());
+        }
+
+        for _ in 0..2000 {
+            for &t in &tokens {
+                notify.mark(t);
+            }
+            notify.poll_limit(&mut events, limit);
+        }
+
+        // Physical length must stay bounded regardless of tick count.
+        // With 2x headroom and compaction on exhaustion, the cap is ~2*n.
+        assert!(
+            notify.dispatch_list.len() <= n * 2,
+            "dispatch_list grew without bound: len={} cap={}",
+            notify.dispatch_list.len(),
+            notify.dispatch_list.capacity(),
+        );
     }
 }
--- a/nexus-notify/src/local.rs
+++ b/nexus-notify/src/local.rs
@@ -74,6 +74,10 @@ pub struct LocalNotify {
     /// Token indices queued for dispatch this cycle (deduped via `bits`).
     dispatch_list: Vec<usize>,
 
+    /// Index of the next token to dispatch. Advances on partial drain
+    /// instead of shifting the Vec; reset to 0 on full drain.
+    dispatch_head: usize,
+
     /// High-water mark for token allocation.
     num_tokens: usize,
 }
@@ -84,6 +88,7 @@ impl LocalNotify {
         Self {
             bits: vec![0u64; capacity.div_ceil(64)],
             dispatch_list: Vec::with_capacity(capacity),
+            dispatch_head: 0,
             num_tokens: 0,
         }
     }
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0 (changes requested):
The motivation is right and the numbers back it. 4096 tokens / limit 32 under sustained
backlog, pinned to core 0, best of 3:

```
main (Vec::drain)     7015 ns/tick mean
this PR (cursor)      3499 ns/tick mean     2x
```

`Vec::drain` really is the dominant cost and the cursor really does remove it. The diff is
clean and the accessors are updated correctly. But it can't land as-is.

## `dispatch_list` grows without bound

The dedup bits guarantee at-most-once among *live* entries, but the stale prefix
`[0..dispatch_head]` has its bits cleared — so, as your description notes, re-marks re-push
to the tail. Nothing reclaims that prefix except a full drain, and under sustained backlog a
full drain never comes. That's the workload `poll_limit` exists for.

128 tokens, limit 32, every token hot each tick:

```
tick     0: len=  128  head=   32  live=96  cap=  128
tick   800: len=25728  head=25632  live=96  cap=32768
tick  1600: len=51328  head=51232  live=96  cap=65536
```

The same probe on main stays at `len=96, cap=128` indefinitely. Live count is correctly
bounded at 96; physical length grows ~`limit` per tick, linearly, forever. Two consequences:

1. **`mark()` reallocates** — 512x past the preallocation in 2000 ticks. That's allocation on
   the hot path, which the crate's premise rules out.
2. **The memory is never returned.** `clear()` doesn't shrink capacity, so even an occasional
   full drain resets the length while the allocation stays for the life of the process.

## The fix — model it on `nexus-net`'s `ReadBuf`

`ReadBuf` solves exactly this shape: a moving head over a buffer that must not grow. Three
parts, in `nexus-net/src/buf/read_buf.rs`:

- **Auto-reset when drained to empty** (`advance`, :104) — `head == tail` snaps both to the
  front. Free, no memmove. You already have this in the full-drain path.
- **`compact()`** (:212) — `copy_within` moves live data to the front. `#[cold]`, called only
  when the tail is out of room.
- **`consumed()`** (:198) — exposes head drift so the caller can decide when to compact.

Mapped over, that's a compaction inside `mark()` when the tail is out of room, rather than a
shift on every partial drain:

```rust
if self.bits[word] & bit == 0 {
    self.bits[word] |= bit;
    if self.dispatch_list.len() == self.dispatch_list.capacity() {
        self.compact();   // #[cold] — drain(..head), head = 0
    }
    self.dispatch_list.push(idx);
}
```

**One adaptation matters.** `ReadBuf`'s live set is usually a small partial frame, so
compacting at exact exhaustion is cheap. Here live is nearly the whole set — at 4096 tokens
live sits at ~4064, so headroom is only 32 and it compacts *every* tick, moving 4064 entries.
I measured that and it's worse than the unbounded cursor. It needs slack so the memmove
amortizes. Two ways, equivalent in effect — pick whichever reads better:

- `Vec::with_capacity(capacity * 2)` and compact on exhaustion, or
- keep the current allocation and compact when `dispatch_head >= num_tokens`

Either way the tail is reclaimed once per `num_tokens / limit` ticks instead of every tick.

```
main (Vec::drain)                      7015 ns/tick    cap 128
this PR (unbounded)                    3499 ns/tick    cap 65,536 and climbing
+ compact on exhaustion, no headroom   4584 ns/tick    cap 128
+ compact on exhaustion, 2x headroom   3551 ns/tick    cap 256, flat at 100k ticks
```

The fix costs ~1.5%, inside the run-to-run spread of the PR itself (3499–3668). Physical size
goes flat: after 100,000 ticks, `len=256 head=160 live=96 cap=256` — no reallocation after
construction, which is what the crate promises. Full suite passes, clippy clean.

I'd avoid worst-case claims either way; single-sample maxima over 20k iterations were too
noisy to separate the variants once pinned.

## Test gap

`poll_limit_partial` drains partially and then *fully*, which resets the cursor — so nothing
in the suite covers partial → re-mark → partial, the only case the cursor changes. That's the
test that should have caught this, and it's what keeps it caught. Worth asserting the bound
directly (physical length stays within a multiple of `num_tokens` after N ticks), not just the
observable behavior.

## This originates in the design, not your code

Invariant #3 in #176 claims `dispatch_list.len() <= num_tokens`; invariant #4 immediately
describes the mechanism that breaks it (re-marks push to tail). The VecDeque section then says
physical size only matters "under permanent backlog — handled here by natural full-drain
resets," which assumes away the exact condition `poll_limit` exists to serve. You implemented
the spec faithfully; the spec was wrong. I'm correcting #176 so it doesn't mislead the next
person.

Good measurement work getting here — the direction is right, it just needs the prefix
reclaimed.

Abso1ut3Zer0 (approved):
Re-measured on a quiet machine (pinned, best-of-7) with an assembly pass — posting the
accurate picture, which is stronger than my earlier rough numbers.

**The change is O(remaining) → O(1) on the partial-drain path**, and the disassembly
confirms it structurally: `poll_limit`'s partial branch drops the `memmove` entirely and
becomes `dispatch_head += n`; `mark` gains only a `#[cold]`, `len == capacity`-guarded
`compact` call — predictable and off the hot path. Full drain is unchanged (neither
version memmoves there), so this is a pure win on the partial-drain case with no cost to
the common path.

Standing backlog, re-marking only the drained tokens (isolates `poll_limit`):

| n | main (`Vec::drain`) | this PR |
|---|---|---|
| 256 | 176 ns | 131 ns |
| 1024 | 170 ns | 139 ns |
| 4096 | 290 ns | 139 ns |
| 16384 | 2367 ns | 137 ns |

Main grows with the backlog; the PR is flat at ~137 ns regardless of N — the memmove vs
`dispatch_head += n` difference in wall-clock. Faster or equal at every point I measured,
across both the isolate-`poll_limit` and the mark-all-then-poll workloads — no regression.
And the compaction keeps physical size bounded (your new test covers that), so the O(1)
comes with no unbounded-memory cost.

One follow-up, not for this PR: nothing in-tree benches sustained partial drain — the
committed `bench_poll_limit` exercises the sync `Poller`, and `perf_local_notify`
full-drains each iteration, so neither engages the cursor. A bench in the standing-backlog
shape would lock in the O(1) property against future regressions.

Approving — correct, tested, and the win holds up under measurement.

PR #622, opened 2026-07-22
docs: add SAFETY justifications to nexus-shm unsafe blocks

ChronoCoders:
Adds inline // SAFETY: comments to every unsafe block and unsafe impl in nexus-shm. No behavior change.

One open item flagged as UNVERIFIED in docs/unsafe-audit-open.md: ShmSlotWriter::create and ShmRingWriter::create write elem_size/capacity to the payload after Segment::create_file returns, which means after the Release store of status=ALIVE. A reader that Acquire-loads status in Segment::attach is not formally guaranteed to see those writes under the C11 memory model. Works on x86 TSO in practice; fix is to move those writes before the Release.

### DIFF
--- a/docs/unsafe-audit-open.md
+++ b/docs/unsafe-audit-open.md
@@ -1,0 +1,34 @@
+# Unsafe audit open items
+
+Items that could not be fully verified during the SAFETY comment pass.
+Each entry has a crate, file, approximate line, and the specific question.
+
+---
+
+## nexus-shm
+
+### 1. Post-Release payload writes in `ShmSlotWriter::create` and `ShmRingWriter::create`
+
+**File:** `nexus-shm/src/slot.rs` (`ShmSlotReader::attach`, line ~135)
+**File:** `nexus-shm/src/ring.rs` (`read_capacity` / `read_elem_size`, lines ~26-31)
+
+**Issue:** `ShmSlotWriter::create` and `ShmRingWriter::create` both write
+`elem_size` (and `capacity` in the ring case) to the payload *after*
+`Segment::create_file` returns. `Segment::create_file` stores `status=ALIVE`
+with `Ordering::Release` inside `ControlBlock::write_header`. Because the
+payload writes happen after that Release, a reader that Acquire-loads
+`status=ALIVE` in `Segment::attach` is not formally guaranteed to observe
+those payload writes under the C11 memory model.
+
+**In practice:** On x86 (TSO), stores are globally visible in program order.
+On ARM/POWER, the page-cache coherency mechanism and the fact that the reader
+typically opens the file only after the writer process has finished `create()`
+provide de-facto ordering. However, this relies on OS-level happens-before
+(filesystem open/close), not the Rust/C11 memory model.
+
+**Fix:** Move the `elem_size`/`capacity` writes to before the Release store
+of `status=ALIVE`, i.e., initialize the payload header inside `Segment::create`
+(or via a callback) before `write_header` is called. This would make the
+Acquire/Release pair formally sufficient.
+
+**Risk level:** Low in practice (x86 Linux), but not formally sound.
--- a/docs/unsafe-audit-open.md
+++ b/docs/unsafe-audit-open.md
@@ -32,3 +32,29 @@ of `status=ALIVE`, i.e., initialize the payload header inside `Segment::create`
 Acquire/Release pair formally sufficient.
 
 **Risk level:** Low in practice (x86 Linux), but not formally sound.
+
+**Tracking:** #625
+
+---
+
+### 2. Ring slot alignment in `slot_ptr` (`ShmRingWriter`/`ShmRingReader`)
+
+**File:** `nexus-shm/src/ring.rs` (`slot_ptr`, line ~53)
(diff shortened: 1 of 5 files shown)


Abso1ut3Zer0 (changes requested):
This is the right way to do a SAFETY-comment pass. Coverage is complete — every
`unsafe` block and `unsafe impl` in the crate now carries a justification — and the
comments are, with the exceptions below, accurate against the code. Most importantly:
when you hit something you couldn't prove sound, you flagged it `UNVERIFIED` with a
tracking doc instead of writing a confident justification you didn't believe. That's
exactly the discipline this pass exists to enforce. A SAFETY comment that quietly
overclaims is worse than no comment, because the next person trusts it.

Writing the invariants down did what it's supposed to do: it turned up two real
soundness gaps that were latent before, neither introduced by this PR. I've filed both
so they get a proper fix with tests — **#624** (ring slot alignment) and **#625**
(post-Release payload ordering). Let's keep this PR a pure no-behavior-change docs pass
and land the actual fixes there. What that means for the diff here:

## 1. `slot_ptr`'s comment overclaims — soften it to match #624

`slot_ptr`'s SAFETY comment says:

> `align_of::<T>() <= DATA_OFFSET` is asserted in `ShmRingWriter::create`, so the cast
> to `*mut T` is alignment-sound.

That assert doesn't imply it. `DATA_OFFSET = 192` isn't a power of two (192 = 64·3), so
a slot at offset 192 is only ever 64-aligned — `192 ≡ 64 (mod 128)` no matter how
aligned `data()` is. A `Pod` type with `align_of == 128` passes the assert (128 ≤ 192)
but lands every slot on a misaligned address → UB in the `copy_nonoverlapping`/`read`.
`Pod` is a public `unsafe trait`, so it's reachable via a user `#[repr(align(128))]`
type. Nothing in-tree hits it (std `Pod` maxes at align 16), which is why it's latent —
details and the one-line fix are in #624.

What makes it worth calling out here: the *same* comment on the slot side is correct,
because `PAYLOAD_OFFSET = 64` is a power of two (`align <= 64` ⇒ `align | 64` ⇒ aligned).
Identical "align ≤ OFFSET" reasoning, sound for the slot, unsound for the ring, purely
because 192 isn't a power of two. Exactly the thing that only shows up once you write the
invariant out.

For this PR: change `slot_ptr`'s comment so it stops asserting soundness it doesn't have,
and point at the gap the way `read_capacity`/`attach` already point at theirs — i.e. add
the alignment gap to `docs/unsafe-audit-open.md` (it only lists the ordering one right
now) and reference it (+ #624). The assert change itself lives in #624.

## 2. The `UNVERIFIED` ordering comments — correct, tracked in #625

I worked through it and I agree with your reading. `write_header` Release-stores
`status = ALIVE`; the `elem_size`/`capacity` writes run after `create_file` returns, so
they're sequenced after that Release. The reader's Acquire on `status` orders the header
fields written *before* the store (magic, data_len) but establishes no happens-before for
bytes written after it — non-atomic write and read, unordered → data race → UB in the
model, benign on x86/ARM in practice exactly as your doc says. Your proposed fix (init
the payload header before the Release, via an init step threaded through
`Segment::create`) is the right one; that's #625.

No change to these comments in this PR — they already point at `docs/unsafe-audit-open.md`
and read honestly. Adding the issue number (#625) to that doc entry is worth doing so the
tracking is linked.

## 3. Two comments don't match their code — fix these here

Small, but this is the one PR where comment accuracy *is* the deliverable, and both are
pure doc corrections (no behavior change), so they belong in this pass:

- **`slot.rs`, `Price` Pod impl:** the comment ends "…no implicit padding beyond the
  declared `_pad` field." `Price` has no `_pad` field — it's `bid: f64, ask: f64, seq:
  u64`, three 8-byte fields with genuinely zero padding. The clause is copy-pasted from
  the `Order` struct in `ring.rs` (which does have `_pad`). Drop it.

- **`ring.rs`, `ShmRingWriter` Send impl:** "`!Sync` prevents shared `&self` writes that
  would corrupt the seqlock invariant." The ring has no seqlock — that's the slot's
  protocol; the ring is an SPSC head/tail buffer. And `try_push` already takes `&mut
  self`, so there's no `&self` write path to prevent. The real reason is the ordinary
  one: `Send` moves the single writer across threads, `Sync` isn't derived because
  sharing `&ShmRingWriter` across threads has no meaning for an SPSC writer. The reader's
  Send comment right below is fine — mirror that.

## Not blocking

The recurring "`data()` is 64-byte aligned (mmap)" shorthand is fine as-is — just worth
knowing the real reason is `data() = mmap_base + HEADER` and `HEADER =
size_of::<ControlBlock>()` is cache-line sized (`CachePadded`), so the page alignment of
the base carries through. Conclusion's correct; no change needed.

Genuinely good pass — thorough, and the honesty on the parts you couldn't close is the
whole point. For this PR: fix the two comments in §3, and register the alignment gap in
`docs/unsafe-audit-open.md` (with #624) so `slot_ptr` stops overclaiming. The real
soundness fixes are #624 and #625.

Abso1ut3Zer0 (approved):

Issue #626, opened 2026-07-24
nexus-notify: Events has no draining iterator — unprocessed tokens are silently dropped on the next poll


Abso1ut3Zer0:
## Summary

`Events` (`nexus-notify/src/event_queue.rs:230`) exposes only borrowing accessors —
`as_slice()`, `iter()`, and `IntoIterator for &'a Events`. There is no draining iterator
and no by-value `IntoIterator`. Consumption happens through a shared reference, so nothing
in the type expresses "these tokens have been taken."

## Why it matters

Every producer clears the buffer before filling it:

- `Poller::poll` / `poll_limit` (`event_queue.rs:163,178`)
- `LocalNotify::poll` / `poll_limit` (`local.rs:157,168`)
- `EventChannel::recv` / `recv_limit` / `recv_timeout` / `try_recv` (`event_channel.rs:95-131`)

all begin with `events.clear()`. And by the time a token is sitting in `Events`, the
notifier has already given up its claim on it — the dedup bit is cleared (`local.rs`) or
the MPSC slot popped and the flag reset (`event_queue.rs`).

So a token the caller does not process before the next poll is **silently discarded**.
There is no error, no counter, no debug assertion — the notification is simply gone. For
an edge-triggered one-shot notifier that can wedge a consumer indefinitely: the wakeup
that would have told it to make progress was dropped on the floor.

With only `iter()`, "I consumed these" is a convention rather than something the type
enforces. Secondary: `iter()` can be called twice, double-processing the same tokens.

## What to add

- `pub fn drain(&mut self) -> Drain<'_>` yielding owned `Token`, removing as it goes.
- Possibly `IntoIterator for Events` (by value) — though for a pre-allocated reusable
  buffer that is probably the wrong shape. Decide explicitly rather than by omission.

## Design questions

1. **Early-exit semantics.** `Vec::drain` removes its whole range on drop even if you stop
   iterating early — which would re-create exactly the silent-loss bug. We want "stop early
   ⇒ untaken tokens stay in the buffer." That is a cursor, not `Vec::drain`. Same shape as
   the `LocalNotify` partial-drain cursor in #618 / #176.

2. **Should `poll` keep calling `clear()`?** If `drain` becomes the consumption path, a
   `poll` that silently clears a non-empty buffer is the same bug in miniature. Options:
   keep `clear()`, `debug_assert!(events.is_empty())`, or append instead. Needs a decision.

[...]

PR #628, opened 2026-07-29
nexus-notify: Events::drain() + poll_limit loss guard

On0n0k1:
- Add `Events::drain()` — a cursor-based draining iterator that removes tokens as they're consumed, leaving any untaken remainder in the buffer if the caller stops early (instead of `Vec::drain`'s memmove-the-remainder-on-drop behavior)
- The cursor (`head`) lives on `Events` itself, not on the `Drain` iterator, so partial drain is a pointer bump rather than an O(remaining) shift — same approach as `LocalNotify`'s `dispatch_head` cursor (#618)
- `Poller::poll_limit` and `LocalNotify::poll_limit` now `debug_assert!` that the previous batch was fully drained before clearing and refilling, turning what was previously a silent, unrecoverable token loss into a loud panic in debug/test builds
- Document the deliberate divergence from mio's `Events` (borrow-only accessors) directly on the type, per the design discussion on #626

Part of #626. Follow-up (not in this PR): update `nexus-rt`'s two dispatch loops to use `drain()`, add dedicated `Drain`-specific test coverage, fix the same read-without-drain pattern in `benches/event_queue.rs`.

## Test plan
- [x] `cargo test -p nexus-notify` — lib, integration, and doctests all pass
- [x] `RUSTFLAGS="--cfg loom" cargo test -p nexus-notify --test loom_event_queue` passes
- [x] `cargo clippy -p nexus-notify --all-targets` clean
- [x] Existing tests/doctests that polled repeatedly without draining updated to call `.drain()` between polls, rather than relaxing the new assert

### DIFF
--- a/nexus-notify/CHANGELOG.md
+++ b/nexus-notify/CHANGELOG.md
@@ -10,6 +10,30 @@ contained.
 
 ## [Unreleased]
 
+### Added
+
+- `Events::drain()`, a cursor-based draining iterator that removes
+  tokens as they're consumed. Unlike `Vec::drain`, stopping early
+  leaves the untaken remainder in the buffer instead of discarding it.
+
+### Breaking
+
+- **`Poller::poll_limit` / `LocalNotify::poll_limit` now `debug_assert!`
+  the previous batch was fully drained before clearing and refilling.**
+  Code that polled again without draining (or explicitly clearing) the
+  prior `Events` batch — previously legal, silently discarding the
+  leftover tokens — now panics in debug/test builds. Release builds
+  are unaffected (`debug_assertions` off).
+
+### Migration notes
+
+Most consumers are unaffected — `cargo update -p nexus-notify` is
+sufficient for any poll loop that already fully consumes `Events` each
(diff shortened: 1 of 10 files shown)


Abso1ut3Zer0 (approved):
Faithful build of the #626 design — the cursor lives on `Events`, `drain()` bumps `head` with no `Drop`/memmove, `clear()` resets it, all accessors offset by `head`, and the `poll_limit` guard uses the cursor-aware `is_drained()` rather than `tokens.is_empty()`. Migrated nexus-rt's two dispatch loops to `drain()` and updated the tests to drain rather than weakening the assert — right call. Verified: nexus-notify + nexus-rt suites, loom (5), and miri (9) all green, clippy `-D warnings` clean. Checked the edge cases too (partial drain leaves the remainder in `as_slice()`, multiple `drain()` calls continue from the cursor). The `Breaking` changelog wording is worth a second look at release-time since release behavior is unchanged (the assert is debug-only), but that's not blocking. Approving.

PR #627, opened 2026-07-27
docs(nexus-collections): add SAFETY comments to all unsafe blocks

ChronoCoders:
Covers `undocumented_unsafe_blocks` for nexus-collections under `--all-targets`: all `unsafe {}` blocks in src, tests, and benches now have `// SAFETY:` comments.

19 files, 244 lines added. Zero clippy warnings under `-W clippy::undocumented_unsafe_blocks`.

Part of the workspace-wide SAFETY audit (follows #622 nexus-shm, #623 nexus-slab).

PR #623, opened 2026-07-23
docs(nexus-slab): add SAFETY justifications to all unsafe blocks

ChronoCoders:
Adds \`// SAFETY:\` comments to every \`unsafe {}\` block in nexus-slab (production code, all test files, and all example files) and \`# Safety\` sections to \`unsafe fn\` doc comments.

Part of the workspace-wide unsafe audit. nexus-shm was covered in #622.

Covers:
- \`src/bounded.rs\`, \`src/unbounded.rs\`
- \`src/byte/bounded.rs\`, \`src/byte/unbounded.rs\`
- \`src/rc/bounded.rs\`, \`src/rc/unbounded.rs\`
- \`tests/miri_tests.rs\`, \`tests/allocator_tests.rs\`, \`tests/random_tests.rs\`
- \`examples/\` (all six benchmark examples)

Zero warnings under \`-W clippy::undocumented_unsafe_blocks --all-targets\`.

PR #629, opened 2026-07-29
docs(nexus-smartptr): add SAFETY comments to all unsafe blocks

ChronoCoders:
Adds `// SAFETY:` comments to all 33 undocumented `unsafe` blocks in nexus-smartptr, making the crate clean under `clippy::undocumented_unsafe_blocks`.

Files touched: `build.rs`, `src/flat.rs`, `src/flex.rs`, `src/meta.rs` (4 files, 33 insertions).

Part of the workspace-wide `undocumented_unsafe_blocks` audit (#622, #623, #627).

### DIFF
--- a/nexus-smartptr/build.rs
+++ b/nexus-smartptr/build.rs
@@ -37,6 +37,7 @@ fn main() {
     let data_ptr: *const u8 = &raw const val as *const u8;
     let trait_ptr: *const dyn Probe = &val as &dyn Probe;
 
+    // SAFETY: trait_ptr is a valid local reference; ptr::read copies its bytes to inspect fat pointer layout.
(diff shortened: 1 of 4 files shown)


Abso1ut3Zer0 (approved):
Verified the load-bearing claims rather than eyeballing: `extract_metadata`'s "verified fat pointer above" (real `is_fat_ptr` early-return before the 2nd-word read), flex's `VALUE_OFFSET` / flat's `META_SIZE` "const assert" (both `new_raw`s invoke `const { assert_*_buffer }`, so the offsets are in-bounds by construction), and the fat-pointer reconstruction comments (offset-0 metadata, Sized-needs-no-metadata, same-type vtable reuse). `undocumented_unsafe_blocks --all-targets` clean, clippy `-D warnings` clean, 67 tests green. Accurate pass on the gnarliest crate in the sweep — approving.

PR #630, opened 2026-07-30
FIX: reposition nexus-fix-engine as a user-driven, sans-IO framework

### LINKED ISSUE #619: FIX: reposition nexus-fix-engine as a framework — clock-free session, transport-per-call, user-driven protocol actions
## Positioning

`nexus-fix-engine` ships a FIX session **framework**, not a FIX engine. We provide *mechanism*; the user owns *policy*. That's the same stance as the rest of the workspace — a framework ecosystem, not an application ecosystem (the FIX journal tool being the one application we ship).

| Ours (mechanism) | Yours (policy) |
|---|---|
| Framing, checksum, BodyLength | The loop, and every timer |
| Sequence assignment + gap detection | Comply with vs refuse a resend |
| The journal | Accept vs reject a Logon (auth) |
| Resend replay encoding (PossDup, gap-fill) | When to disconnect, and why |
| Inbound validation, state tracking | What a Reject means to your app |

Anyone wanting the batteries-included engine wraps this and exposes a callback API. We're the layer underneath.

## Binding design rules

1. **One variant = one required response.** If a situation needs a different response, it's a different variant. Nothing may require probing the engine or digging through a payload to discover which of N actions applies.
2. **The variant carries what its response needs** — `TestRequest` carries the TestReqID so the answer is one call, never "go find tag 112."
3. **No hidden unbounded work.** Anything that can run long or block (resend) is user-pumped, never driven to completion inside `recv`.
4. **Clock-free.** The state machine holds no timer state; `now` appears only where a SendingTime(52) is stamped — i.e. only on sends.
5. **State machine ≠ I/O.** The session never owns the transport, mirroring `nexus-web` (`WsReader::recv(&mut self, conn: &mut S)`).

## Motivating defects in the current design

- **A resend blocks the session thread.** An inbound ResendRequest drives the *entire* replay inside `recv` with blocking writes, and surfaces `Message::ResendRequest` only *after* it completes. A peer demanding 10k (or 1M) messages parks the thread with no chance to pace, bound, or refuse — a backpressure and DoS hazard.
- **Session lifetime ≠ connection lifetime.** A FIX session must survive reconnects with its sequence numbers and journal intact. `FixConnection<S, …>` owning the socket fights that.
- **Variants that need probing.** A plain Logon and a reset Logon (141=Y) need different responses but arrive as one variant. `Heartbeat`'s doc literally says "no reply required *unless* it carries a TestReqID" — i.e. go probe.
- **Docs that describe a manual mode that doesn't exist.** Several variant docs say the caller replies ("Send your own Logon back", "Echo the TestReqID") while the engine already auto-replies.

## Surface

**Receive** — no timestamp; receiving stamps nothing and emits nothing:
```rust
session.recv(&mut transport)      -> Result<Message<'_, D>, Error>
session.try_recv(&mut transport)  -> Result<Option<Message<'_, D>>, Error>
```

**One action per variant**, each spelled out in its doc:

| Variant | Required response |
|---|---|
| `Application { header }` | route/decode |
| `TestRequest { id }` | `heartbeat(t, now, Some(id))` |
| `Heartbeat` / `Reject` / `SequenceReset` | none |
| `GapDetected { begin }` | `resend_request(t, now, begin)` |
| `ResendRequest { cursor }` | pump the cursor (drop = refuse) |
| `LogonRequest(d)` / `LogonResetRequest(d)` | `d.accept(..)` or `d.reject(..)` |
| `LogonAcknowledged` / `LoggedOut` | none |
| `LogoutRequest { msg }` | `logout(t, now, reason)` |

**Send helpers** — one per protocol action, each stamping SendingTime from `now`, assigning the sequence number, journaling where required: `heartbeat`, `test_request`, `resend_request`, `logon` / `logon_reset`, `logout`, `send_app`, `allocate_seq`.

**Resend cursor** — lazy and sans-IO; yields bytes, the caller writes and paces:
```rust
Message::ResendRequest { cursor } => {
    let mut c = cursor;
    while let Some(bytes) = c.next(&mut session)? {
        transport.write_all(bytes)?;   // sync
        // .await + yield_now() every N  // async
    }
}
```
Per batch it decides replay-as-PossDup vs `SequenceReset-GapFill` (admin messages, rotated-off segments, anything below the journal's retained low-water, clamped at `next_outbound`), so a request entirely outside our window pumps one gap-fill.

**Timers — yours.** The session exposes only `heartbeat_interval()` (negotiated at Logon). Three timers, each self-identifying so nothing needs interpreting when it fires:
1. heartbeat ticker → `heartbeat(t, now, None)`
2. peer liveness — reset on *any* inbound; first fire → `test_request`, restart shorter; second fire → peer dead, you disconnect
3. handshake/logout deadline → you give up

tokio (`interval` + `timeout` in `select!`) and sync (`Instant` + socket read timeout) recipes ship as documented examples, and are a first-class deliverable: forget timer 2 and your session never notices a dead peer.

## Removed

- State: `last_sent`, `last_received`, `state_entered`, `test_request_sent`, `inbound_grace`
- API: `on_timeout`, `tick`, `next_timeout`
- Behavior: every auto-emit inside the state machine — auto Heartbeat answering a TestRequest, auto Logon reply, auto ResendRequest on a gap, auto-driven resend
- Timeout `DisconnectReason`s become reasons *you* pass to `logout()`

## Phasing

1. Clock-free session — drop timer state, `on_timeout`, `next_timeout`; add `heartbeat_interval()`
2. Transport-per-call — session stops owning `S`
3. Unwind auto-emit — send helpers, variant splits, obligation docs
4. Resend cursor
5. Conformance rewrite + timer recipes + docs
6. Optional: bundled `FixConnection` layer

## Open questions

1. `Transport` trait shape — async reuses `nexus_net::WireStream`; sync is `Read + Write`. One trait, or two surfaces?
2. Does the bundled `FixConnection` ship in this cut or follow?
3. Does `FixJournal` expose its retained low-water today? The cursor's gap-fill decisions need it.
4. `SessionCustomizer` — with sends becoming user-called helpers, confirm the hook still lands after the session header and before checksum.

## Supersedes

Replaces the recv-contract work on #613 / PR #617, which re-cut the same surface before this positioning was settled. That PR should be closed.

Abso1ut3Zer0:
## Summary

Repositions `nexus-fix-engine` from a FIX **engine** (owns the loop, the timers, and every auto-reply) to a FIX **framework**: the crate provides *mechanism* (framing, checksum/BodyLength, seqnum assignment + gap detection, the journal, resend replay encoding, inbound validation, state tracking); the user owns *policy* (the loop, every timer, every reply, comply-vs-refuse a resend, accept-vs-reject a Logon). Same wire behavior, a far more honest and composable API.

Closes #619. Supersedes the recv-naming work in #617 (closed).

## What this gives you

- **Three-object, sans-IO core.** `FixSession` (state machine + journal — no buffers, no socket) + `MessageReader` + `MessageWriter`, held as `FixParts`. `session.recv(&mut reader, &mut writer, &mut conn, now)` returns a `Message<'r>` that borrows **only the reader**, so `&mut session`/`&mut writer` stay free to reply with **no copy-out**. The full byte seam (`poll`/`message`/`spare`/`filled`/`data`/`advance`) is public — a kernel-bypass / io_uring transport drives the state machine directly, with no `recv`/`WireStream` needed.
- **Deterministic.** Every `SendingTime(52)` comes from a caller-supplied `now: i128` (UTC unix-nanos) — on `recv` and on every send. No internal `SystemTime::now()` remains in the emit path; the core is a pure function of `(bytes, now)`, so a test or a historical replay feeds a fixed clock and **every outbound byte is reproducible**.
- **User-driven (policy → user, mechanism → engine).** The engine no longer auto-replies. Unsolicited inbound admin surfaces as a typed `Message` carrying its single required response; the engine keeps only *mechanism* it initiated (the in-session reset handshake, protocol-error disconnects). See the table below.
- **Non-blocking resend.** An inbound ResendRequest hands back an owned `Copy` `ResendCursor` you pump (`next` / `next_batch(max)`) and pace yourself — replacing the old blocking in-`recv` replay (a backpressure/DoS hazard). A request the journal can't fully cover surfaces as `ResendOutOfRange` (both water-marks carried) to gap-fill / reset / logout as you choose.
- **Batteries, optional.** `FixConnectionBuilder` does the socket/TLS setup and hands back the raw `(FixParts, S)` (**primary**, zero-copy, run the ordinary loop); an owns-everything `FixConnection` (**secondary**, `open`/`from_parts`) bundles it into one object at the cost of a small copy-out. Mirrors nexus-web's `WsStreamBuilder` → parts / `WsStream` split.
- **Timer recipes ship as worked examples** (sync + tokio) — the three user-owned timers: heartbeat ticker, two-phase peer-liveness (probe → dead, reset on any inbound), handshake deadline. A user who omits the liveness timer never notices a dead peer, so these are a first-class deliverable.

## The `Message` → response contract

| Variant | Your required response |
|---|---|
| `Application { header }` | route / decode |
| `TestRequest { id: &AsciiTextStr }` | `heartbeat(w, conn, now, Some(id))` |
| `GapDetected { begin }` | `resend_request(w, conn, now, begin)` |
| `LogonRequest` / `LogonResetRequest(decision)` | `decision.accept(..)` / `.reject(.., reason)` (or `session.accept_logon`/`reject_logon`) |
| `LogoutRequest { msg }` | `logout(w, conn, now, reason)` |
| `ResendRequest { cursor }` | pump the cursor (drop = refuse) |
| `ResendOutOfRange { begin, end, low_water, high_water }` | `sequence_reset` / `gap_fill`, or logout |
| `Heartbeat` / `Reject` / `SequenceReset` / `LogonAcknowledged` | none |
| `LoggedOut` | none — terminal |

Reply-carrying payloads stay borrowed (`id` is a `&AsciiTextStr` into the frame; `Application` decodes zero-copy); adversarial/malformed admin surfaces as a `Malformed` error rather than being swallowed.

## Phases (reviewable commit-by-commit)

1. **Clock-free session** — delete the monotonic-`Instant` timer machinery (`on_timeout`/`tick`/`next_timeout`).
2. **Un-nest the buffers** — three-object surface; transport passed per call; the session stops owning the socket.
3a. **Send surface** — `encode_<verb>` (sans-IO) + combined `<verb>` (encode + flush) for every action, both crates.
3b. **User-driven unwind** — delete auto-emit, split the `Message` variants, add `LogonDecision`, thread the caller-supplied `now` through everything, rewrite both conformance suites.
4. **Resend cursor** — user-pumped replay + up-front providability check + `ResendOutOfRange`; `sequence_reset` / `gap_fill` helpers.
5. **Docs + timer recipes + Text(58)** — `reject`/`logout` reasons reach the wire via `Option<&AsciiTextStr>`.
6. **`FixConnection` batteries** — builder → raw parts (primary) + owns-everything bundle (secondary).
- **housekeeping** — a consistency pass: dead-code removal, doc-drift fixes, sync↔async naming parity.

## Migration (breaking)

The old `Session::new` / `handle_message` / `poll_event` / `handle_timeout` surface is gone. The new loop:

```rust
let FixParts { mut session, mut reader, mut writer } =
    FixSession::builder().customizer(MyVenue).build(state, config, journal);
let mut conn = TcpStream::connect(addr)?;
session.connect(&mut writer, &mut conn, now)?;          // send the opening Logon

loop {
    match session.recv(&mut reader, &mut writer, &mut conn, now)? {
        Some(Message::TestRequest { id })     => session.heartbeat(&mut writer, &mut conn, now, Some(id))?,
        Some(Message::Application { header })  => route(header),
        Some(Message::LoggedOut { .. })        => break,
        _ => {}
    }
    // your heartbeat / liveness / handshake timers — see examples/timer_recipes.rs
}
```

## Testing

- Both conformance suites (sync + async) rewritten to the user-driven model and driven against the Python peer — **same wire behavior**, `SendingTime(52)` asserted exactly against a fixed clock.
- Coverage for: reconnect (`FixParts` survives a fresh socket), the sans-IO seam (drives the machine with no `recv`/`WireStream`), the `recv` borrow guarantee, `ResendCursor` byte-parity vs the old walk, `ResendOutOfRange` bounds, `Text(58)` present/absent, and feature-gated TLS `connect_tls`.
- Guardrails green across `nexus-fix-codec` / `nexus-fix-engine` / `nexus-async-fix-engine`: `build --workspace`, `clippy --all-features --all-targets -D warnings`, `fmt --check`, `cargo doc -D warnings`, full test suite.

## Known limits / follow-ups (non-blocking)

- Acceptor-side `LogonDecision` isn't a bundle method (it borrows the reader, conflicting with `&mut self`) — bundle users drop to `into_parts` for auth. Documented.
- `TransportError` is also reachable via the `fix_session::Error` module path — a cosmetic alias; a rename is a possible small follow-up (kept as-is to preserve the module's public docs).
- Required-field-presence validation beyond `TestReqID(112)`, and an `AsciiTextStr` literal macro, are unscheduled.


Abso1ut3Zer0:
## How to read this PR

Repositions `nexus-fix-engine` from a FIX **engine** (owns the loop, the timers, and every auto-reply) to a FIX **framework**: the crate provides *mechanism* (framing, checksum/BodyLength, seqnum assignment + gap detection, the journal, resend encoding, state tracking); the caller owns *policy* (the loop, every timer, every reply). Same wire behavior, a more honest and composable API. It's large — so here's the mental model first, then a layered reading order.

### Mental model — four ideas explain most of the diff

- **Three objects, not one.** `FixSession` (state machine + journal — no buffers, no socket) + `MessageReader` + `MessageWriter`, held as `FixParts`. Reconnect is "same session, new socket" — seqnums and the journal survive. Mirrors `nexus-web`'s `WsReader`/`WsWriter`/`conn`.
- **`recv` borrows the reader only.** `session.recv(&mut reader, &mut writer, &mut conn, now)` returns a `Message<'r>` tied to `reader`, so `&mut session`/`&mut writer` stay free to send the reply while the payload is still borrowed — no copy-out. This one constraint explains a lot of the shape: `poll()` returns a *non-borrowing* outcome, and a separate `message()` re-borrows the frame in place.
- **Deterministic clock.** The core reads no clock of its own; every `SendingTime(52)` stamps from a caller-supplied `now: i128` (UTC unix-nanos). The core is a pure function of `(bytes, now)` — replayable for tests and historical replay.
- **Mechanism, not policy.** The engine no longer auto-replies. Unsolicited inbound surfaces as a typed `Message` carrying its one required response (see the table in `lib.rs`); the engine keeps only mechanism it *initiated* — the reset handshake and protocol-error disconnects.

### Suggested review order

Review the **net diff** ("Files changed"), top-down by layer — **not** commit-by-commit. The history has ~9 unrelated `main` commits interleaved from a merge, and the phases refactor each other (phase 2 rewrites phase 1's buffers), so commit-order makes you read the same code three times. For narrative, follow the phase commits only (`3b42ba62` → `b46423b1`).

**0 · Thesis.** `src/lib.rs` module docs + this PR body. This *is* the contract: the three-object trio, the `Message`→response table, the deterministic clock, "timers are yours." If you disagree with the thesis, stop here — that's the first conversation to have.

**1 · Public API (highest leverage, hardest to change later).** `framework.rs` (`Message`, `MessageReader/Writer`, `ResendCursor`, `LogonDecision`), the `recv` signature + the `encode_X`/`X` send-verb pairs in `fix_session.rs`, and `connection.rs` (`FixConnection`). Watch for: misuse-resistance, consistency across verbs/constructors, and whether the reader-only borrow actually holds.

**2 · Hot path.** `fix_session.rs`: `recv` → `poll` → `process_frame` → `message` + the classify dispatch; `session/mod.rs`: the `on_*` handlers + seqnum/gap/resend transitions. Watch for: allocation-free steady path, single-writer seqnum, `now` threaded everywhere (no hidden `SystemTime::now`). Read the deleted `transport.rs` once — it's the old engine-owned loop the three-object recv replaced, so it shows what changed.

**3 · Behavior parity (the proof).** `tests/fix_conformance.rs` + `fixtures/fix_peer.py` (the Python counterparty), the `transport`/`session`/`connection` tests, async conformance. These assert the *same wire bytes* (SendingTime pinned to a fixed clock) and cover the new surfaces: the `recv` borrow guarantee, the resend cursor, out-of-range, reconnect, and the raw sans-IO seam.

**4 · Skim.** `examples/*timer_recipes.rs` (the three timers the user now owns — heartbeat, peer-liveness probe, handshake deadline), README/CHANGELOG, and the async `lib.rs` (mostly a newtype mirroring core — parity-skim, don't re-derive).

### Navigating `fix_session.rs` (the +2.5k file)
Don't read it top-to-bottom. Jump to the anchors — `recv`, `poll`, `message`, `process_frame`, the send verbs, the builder — and use the per-file "Viewed" checkboxes to burn down the 30 files.

Abso1ut3Zer0:
**Merging — the reposition is complete.** `nexus-fix-engine` is now a user-driven, sans-IO FIX **framework**: mechanism (framing, checksum/BodyLength, seqnum + gap detection, the journal, resend replay, state tracking) is ours; policy (the loop, every timer, every reply, comply-vs-refuse a resend) is yours. Deterministic caller-supplied clock, the three-object `FixParts` trio, a user-pumped resend cursor, and `FixConnection` batteries. Reviewed by Altug — thanks.

## What's next

**Policy toolkit + cookbook (#665)** — the immediate follow-up direction. Mechanism-not-policy means we don't *impose* policy, not that we ship none: a curated menu of opt-in, sans-IO policy pieces you compose (same house pattern as `nexus-rate` / `nexus-stats`). First up is `PeerLiveness` — the two-phase peer-silence watchdog that today lives in `examples/timer_recipes.rs` as copy-this-correctly code; it's the one timer that's a state machine rather than a ticker, and forgetting it leaves a half-open connection alive with nothing watching. Variants (two-phase probe / idle-timeout / strict echo-match), then reconnect/backoff. Everything gets cookbook docs; the recipes stay as the compose-it-yourself reference.

**Discovery investigations** (perf/cleanup, none blocking):

- **#635** — single-pass inbound field extraction (kill redundant `find_tag` scans + the TestReqID re-scan)
- **#636** — single decode on the admin inbound path
- **#637** — outbound per-message left-align memmove
- **#639** — compile-time session-role typestate to shrink the state space

nexus-fix-engine/tests/fix_conformance.rs
```diff
@@ -210,7 +325,7 @@ fn conformance_heartbeat() {
     let dir = tmp_dir("heartbeat");
     let (mut child, port) = spawn_peer("heartbeat");
     let mut conn = connect(port, dir.path());
-    conn.connect(Instant::now()).unwrap();
+    conn.connect().unwrap();
     drive(&mut conn);
     assert!(child.wait().unwrap().success());
```
ChronoCoders, line 330: `heartbeat_fires_on_outbound_idle` is present at the merge base and absent here. The `conformance_heartbeat` scenario exercises the user-driven TestRequest echo path (the test rig calls `heartbeat(now, Some(id))` in its step handler), not the autonomous outbound-idle timer. The harness clock is fixed and never advances, so the idle-tick path cannot fire in this harness.

nexus-fix-engine/src/framework.rs
```diff
@@ -86,63 +83,256 @@ impl core::fmt::Display for SessionError {
...
+/// mutually exclusive — it has no destructor: **dropping it without pumping simply
+/// refuses the resend** (nothing is sent). Pace, bound, or abort the replay in your
+/// own loop; that is how you refuse an abusive multi-million-message demand instead
+/// of parking your session thread.
+///
+/// The replay pump (`next` / `next_batch`) is defined on the unix-only
+/// [`FixSession`](crate::FixSession) side, since it drives the journal; the state
+/// here is platform-independent.
+#[derive(Clone, Copy, Debug)]
+pub struct ResendCursor {
```
ChronoCoders, line 220: `LogonDecision` has `#[must_use = "a Logon must be accepted or rejected"]`, but `ResendCursor` does not. The doc says drop = refuse, which is intentional, but an accidental drop also silently refuses the resend and leaves the peer with an unanswered gap, with no compiler diagnostic. Since `LogonDecision` is `#[must_use]` precisely because an accidental drop has a silent consequence, the asymmetry seems worth a deliberate call here.

ChronoCoders:
Four `DisconnectReason` variants are removed at this head: `LogonTimeout`, `LogoutTimeout`, `TestRequestTimeout`, and `ResetTimeout`. The CHANGELOG entry for `DisconnectReason` covers `Logout` -> `PeerClosed` but does not name these four, so someone planning a migration from the CHANGELOG will not see them. Worth adding them to the entry.

PR #638, opened 2026-08-06
nexus-notify: dedicated Drain tests + bench fix (#626)

On0n0k1:
## Summary
Closes out the two minor follow-ups noted on #626 after #628 landed:

- Dedicated `Drain` test coverage — `drain_partial_leaves_remainder`,
  `drain_full_empties_buffer`, and `poll_limit_panics_on_undrained_buffer`
  (both `Poller` and `LocalNotify`). The cursor mechanics and the
  panic guard were previously only exercised indirectly, through
  existing tests that happened to call `drain()` between polls.
- `benches/event_queue.rs` — the `poll`/`poll_limit`/`recv` closures
  left tokens undrained across `b.iter()` calls, tripping the
  `poll_limit` assert under `debug_assertions` (latent, not currently
  panicking under plain `cargo bench` since the bench profile has
  assertions off by default). Added `drain()` after each poll/recv
  call.

## Test plan
- [x] `cargo test -p nexus-notify` — 52 lib tests pass (up from 48)
- [x] `cargo clippy -p nexus-notify --all-targets` clean
- [x] `RUSTFLAGS="-C debug-assertions=on" cargo bench -p nexus-notify --bench event_queue` — every group ran its full sample count (up to ~800M iterations) with no panics

### DIFF
--- a/nexus-notify/benches/event_queue.rs
+++ b/nexus-notify/benches/event_queue.rs
@@ -15,6 +15,7 @@ fn bench_notify(c: &mut Criterion) {
         let mut events = Events::with_capacity(4096);
         b.iter(|| {
             poller.poll(&mut events);
+            events.drain().for_each(drop);
             notifier.notify(token).ok();
         });
     });
@@ -69,6 +70,7 @@ fn bench_poll(c: &mut Criterion) {
                     notifier.notify(*t).ok();
                 }
                 poller.poll(&mut events);
+                events.drain().for_each(drop);
             });
         });
     }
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0 (approved):
Closes out the two #626 follow-ups cleanly. The `Drain` tests hit the right property — `drain_partial_leaves_remainder` exercises the early-exit-keeps-remainder semantics that distinguish the cursor from `Vec::drain`, and the `poll_limit` panic guards are correctly `#[cfg(debug_assertions)]`-gated so they don't fail in release. Verified: 52 lib tests pass, clippy clean, and the benches run to completion under `-C debug-assertions=on` with no panics — the actual thing the fix addresses.

Non-blocking nit: `bench_poll` / `bench_poll_limit` could use `events.clear()` instead of `drain().for_each(drop)` for the reset — O(1), and it keeps those benches measuring the poll path in isolation rather than adding an O(N) consume pass to the timed region. Approving.

On0n0k1:
I included the O(1) changes as discussed to the last PR.

PR #642, opened 2026-08-07
docs(nexus-ascii): add SAFETY comments to all unsafe blocks

ChronoCoders:
Add `// SAFETY:` comments to every `unsafe` block in nexus-ascii so the crate passes `cargo clippy -p nexus-ascii --all-targets -W clippy::undocumented_unsafe_blocks` with zero warnings.

Covered: format.rs macro bodies, SIMD scalar/SSE2 intrinsics, str_ref/string/text/text_ref split_once helpers, flat_string test helpers, and all example bench files.


Abso1ut3Zer0 (approved):
valid-ASCII tags match their literals; SIMD len>=8 tail comments backed by real early-exits. Approving.

PR #645, opened 2026-08-08
docs(nexus-platform): add SAFETY comment to Sync impl for Mapping

ChronoCoders:
Add `// SAFETY:` comment to the `unsafe impl Sync for Mapping {}` at `nexus-platform/src/mapping/mod.rs`. The preceding `Send` impl already had a comment; `Sync` was missing one.

### DIFF
--- a/nexus-platform/src/mapping/mod.rs
+++ b/nexus-platform/src/mapping/mod.rs
@@ -272,6 +272,8 @@ impl Drop for Mapping {
 // state. Concurrent access to the mapped contents must be synchronized
 // by the caller — the handle itself is safe to move across threads.
 unsafe impl Send for Mapping {}
+// SAFETY: same as Send; the Mapping handle has no interior mutability,
+// so shared references can cross thread boundaries safely.
 unsafe impl Sync for Mapping {}
 
 // ── POSIX shared memory (forwarded to platform backend) ──────────


Abso1ut3Zer0 (changes requested):
The comment reads fine as a **Send** justification ("the handle itself is safe to move across threads"), but the `Sync` impl it sits on is a pre-existing soundness gap, and documenting it is the moment to surface that rather than bless it.

`Mapping` is a public type with a **safe** `pub fn write_at(&self, ...)` that mutates the mapped bytes through a shared reference (`copy_nonoverlapping` on the stored pointer). With `unsafe impl Sync for Mapping`, safe downstream code can `Arc<Mapping>`, share `&Mapping` across threads, and call `write_at` from two threads concurrently — a data race on the mapped memory, with no `unsafe` anywhere. The comment's "concurrent access must be synchronized by the caller" states a contract the safe API doesn't (and can't) enforce.

This isn't yours to have introduced — the `Sync` impl predates this PR. But the fix is at the API level, not the comment: make `write_at` `unsafe` (documenting the single-writer requirement), or take `&mut self`, or drop the `Sync` impl, or write the bytes atomically. Until one of those, the `Sync` justification can't be made truthful.

Suggest holding this comment PR and tying it to that decision, or landing a comment that names the gap rather than asserting synchronization the type doesn't require. Happy to file a separate soundness issue to track the fix.

ChronoCoders:
Superseded by #682, which removed the Sync impl entirely to fix the soundness gap. Applying this would reintroduce it.

PR #648, opened 2026-08-08
docs(nexus-async-web): add SAFETY comments to bench unsafe blocks

ChronoCoders:
Add `// SAFETY:` comments to three undocumented `unsafe` blocks in `nexus-async-web/benches/perf_ws_cycles_tokio.rs`: two rdtsc intrinsic blocks and one `Waker::from_raw` call.

### DIFF
--- a/nexus-async-web/benches/perf_ws_cycles_tokio.rs
+++ b/nexus-async-web/benches/perf_ws_cycles_tokio.rs
@@ -23,6 +23,7 @@ use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
 #[inline(always)]
 fn rdtsc_start() -> u64 {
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0:
This overlaps #660 — both add near-identical SAFETY comments to the same bench file (`perf_ws_cycles_tokio.rs`). #660 has the marginally more precise noop-waker wording (`clone` forwards the sentinel pointer but never dereferences it, so "none of them dereference it" is more accurate than "all ignore it"). Suggest closing this in favor of #660 to avoid the conflicting duplicate — both are otherwise correct.

ChronoCoders:
Closing in favor of #660, which merged with the more precise wording.

PR #650, opened 2026-08-08
docs(nexus-shm): add SAFETY comments to unsafe Pod impls

ChronoCoders:
Add `// SAFETY:` comments to the twelve `unsafe impl Pod for ...` blocks in `nexus-shm/src/pod.rs`. The first impl already had a comment; the remaining eleven (u16 through f64, and isize) each needed their own comment on the immediately preceding line.


Abso1ut3Zer0 (approved):
scalar-primitive Pod impls; same-as-u8 holds for all including f32/f64/usize/isize. Approving.

PR #651, opened 2026-08-08
docs(nexus-slot): add SAFETY comments to undocumented unsafe blocks

ChronoCoders:
Add `// SAFETY:` comments to eleven undocumented `unsafe` blocks across five files:

- `src/spmc.rs`: `Sync for Inner` impl + two `Pod` impls for test structs
- `src/spsc.rs`: `Sync for Inner` impl + three `Pod` impls for test structs
- `benches/perf_slot_latency.rs`: `Pod` impl for `Message` + rdtsc intrinsic block
- `benches/perf_slot.rs`: `Pod` impl for `Quote`
- `benches/profile_slot.rs`: rdtsc intrinsic block

### DIFF
--- a/nexus-slot/benches/perf_slot.rs
+++ b/nexus-slot/benches/perf_slot.rs
@@ -22,6 +22,7 @@ struct Quote {
     _padding: [u8; 216],
 }
 
+// SAFETY: Quote is repr(C) with only Copy fields, no heap pointers, no Drop, all bit patterns valid.
 unsafe impl Pod for Quote {}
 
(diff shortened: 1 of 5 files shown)


Abso1ut3Zer0 (approved):
all annotated Pod structs are repr(C), padding-free, integer/float fields; the Sync note defers to the seqlock Send reasoning. Approving. Note: the crate-wide blanket Pod for T: Copy has a separate soundness problem (tracked separately), but these specific comments are accurate.

PR #652, opened 2026-08-08
docs(nexus-stats-detection): add SAFETY comments to undocumented unsafe blocks

ChronoCoders:
Add `// SAFETY:` comments to four undocumented `unsafe` blocks:

- `nexus-stats-detection/src/bocpd.rs`: two blocks where the existing SAFETY comment was separated from the `unsafe` keyword by intervening statements; moved the comments to immediately precede each block
- `nexus-stats-smoothing/src/windowed_median.rs`: two paired `from_raw_parts_mut` calls where the SAFETY comment covered only the first; added `// SAFETY: same as above` before each second call (appears twice in the file)


Abso1ut3Zer0 (approved):
relocates an existing invariant comment to sit above its unsafe block; same-invariant-as-Pass-1 verified. Approving.

PR #657, opened 2026-08-08
docs(nexus-id): add SAFETY comments to unsafe blocks

ChronoCoders:
Adds // SAFETY: comments to 10 unsafe blocks across src, benches, and tests. Passes -W clippy::undocumented_unsafe_blocks.


Abso1ut3Zer0 (approved):
unwrap_unchecked justified by construction-time hex validation; SIMD casts in-bounds; decode reorder sound. Approving.

PR #658, opened 2026-08-08
docs(nexus-fix-codec): add SAFETY comments to all unsafe blocks

ChronoCoders:
Adds missing `// SAFETY:` comments to all `unsafe` blocks in nexus-fix-codec, satisfying `clippy::undocumented_unsafe_blocks` with zero warnings.


Abso1ut3Zer0 (approved):
get_unchecked(scale) bounded by scale<=19 via new/parse. Approving. Aside: new_unchecked is a safe const fn that can seed scale>19 in release; pre-existing, worth a separate look, not this PR.

PR #660, opened 2026-08-08
docs(nexus-async-web): add SAFETY comments to all unsafe blocks

ChronoCoders:
Adds missing `// SAFETY:` comments to all `unsafe` blocks in nexus-async-web, satisfying `clippy::undocumented_unsafe_blocks` with zero warnings.

PR #655, opened 2026-08-08
docs(nexus-queue): add SAFETY comments to unsafe blocks

ChronoCoders:
Adds // SAFETY: comments to 6 unsafe blocks in bench files (all rdtsc intrinsic usages). Passes -W clippy::undocumented_unsafe_blocks.

PR #643, opened 2026-08-08
docs(nexus-net): add SAFETY comments to rdtsc intrinsic unsafe blocks

ChronoCoders:
Add `// SAFETY:` comments to the two undocumented `unsafe` blocks in `nexus-net/examples/perf_buf.rs` (rdtsc intrinsics). Fixes `clippy::undocumented_unsafe_blocks` for this crate.

PR #641, opened 2026-08-07
docs(nexus-async-rt): add SAFETY comments to all unsafe blocks

ChronoCoders:
Adds `// SAFETY:` comments to all 195 undocumented `unsafe` blocks in nexus-async-rt, making the crate clean under `clippy::undocumented_unsafe_blocks`.

25 files touched (src, tests, benches). Main patterns covered: raw waker vtable fns (`Arc::from_raw`/`into_raw`), UnsafeCell state access, lock-free queue raw pointer ops, `Pin::new_unchecked`, custom global allocator delegation, and x86_64 rdtsc intrinsics in benchmarks.

Part of the workspace-wide `undocumented_unsafe_blocks` audit (#622, #623, #627, #629).

PR #661, opened 2026-08-08
chore: enable clippy::undocumented_unsafe_blocks workspace-wide

ChronoCoders:
Adds `undocumented_unsafe_blocks = "warn"` to `[workspace.lints.clippy]` in the root `Cargo.toml`.

All workspace crates have been audited and pass this lint. This gates future unsafe blocks behind a required `// SAFETY:` comment.


Abso1ut3Zer0 (changes requested):
The comment content is fine — the problem is structural, and it's blocking.

**Merge ordering.** This flips `undocumented_unsafe_blocks` on workspace-wide (the one-line add to `[workspace.lints.clippy]` in the root `Cargo.toml`; all members already inherit via `[lints] workspace = true`), but the SAFETY comments for ~15 crates still live only in the open sibling PRs (#643–#660, none merged). Since CI runs `clippy --all-features --all-targets -- -D warnings`, merging this first turns the new warn into a hard error across every crate whose unsafe blocks aren't documented yet — e.g. nexus-id still has undocumented `unsafe` on `main` today (closed by #657). So this can't land until those siblings do.

**It also isn't fully self-consistent even for the crates it touches.** It documents nexus-queue's `Sync` impls but not the crate's bench files (undocumented `rdtsc` blocks in `bench_spsc.rs` etc.), which `--all-targets` lints — #655 is what covers those. And its bundled commits duplicate #641/#642/#659 wholesale and #655 partially, so it'll conflict with them.

**Suggested shape:** reduce this PR to just the single-line `Cargo.toml` change — that one line *is* the capstone — drop the duplicated per-crate comments, and merge it **last**, once #641–#660 have all landed and the tree is already clean under the lint.

Minor, separate: in `nexus-async-rt/src/channel/mpsc.rs` / `mpsc_bytes.rs`, a couple of blocks are justified with "allocated by `Box::new`; not yet reclaimed," but every other comment in those files grounds node liveness in the **Arc refcount bumped in `push()`**. The block is sound for that same refcount reason as the line directly above it — just align the wording.

ChronoCoders:
This PR is now the single Cargo.toml line: `undocumented_unsafe_blocks = "warn"` added to the workspace `[lints.clippy]` block.

Backlog as of 2026-08-18 on upstream/main, rustc 1.94.0, `--all-targets`, Linux (WSL):

| crate | total | src | examples |
|---|---|---|---|
| nexus-journal | 16 | 16 | 0 |
| nexus-rt | 8 | 6 | 2 |
| nexus-slot | 13 | 13 | 0 |
| **total** | **37** | **35** | **2** |

Most of these are not 37 distinct problems. The nexus-journal and nexus-slot sites are runs of primitive `unsafe impl Pod for T {}` under a single block SAFETY comment; clippy scopes a comment to the single impl that immediately follows it, so each impl after the first fires. The mechanical fix is one comment per impl or a macro. The nexus-rt sites are in `view.rs` (6) and one example (2).

nexus-rt depends on nexus-platform, which is `compile_error!` on non-Linux targets. Measuring on Windows gives 29, not 37.

This PR is blocked until the count reaches zero. New PRs move the count in both directions, so a re-measure is needed immediately before merge.

ChronoCoders:
The sibling per-crate PRs (#641 through #660) have all merged. The merge-ordering concern is resolved.

Commit c36698d8 adds the 37 remaining SAFETY comments: `nexus-journal/src/pod.rs` (16 impls), `nexus-slot/src/lib.rs` (13 impls), `nexus-rt/src/view.rs` (6 impls), and `nexus-rt/examples/projection_spike.rs` (2 impls). The branch now closes the backlog itself rather than depending on anything else landing first.

Verified with clippy at 1.94.0, the toolchain CI pins, across the workspace with `--all-targets`: zero `undocumented_unsafe_blocks` warnings.

On the wording point: the "allocated by `Box::new`; not yet reclaimed" comments are in `nexus-async-rt/src/channel/mpsc.rs` (lines 128, 198) and `mpsc_bytes.rs` (lines 129, 197), present on `main` today, and not touched by this PR.

Abso1ut3Zer0 (approved):
Re-reviewed on `c36698d8`. Slimmed to the capstone shape I asked for — one-line `[workspace.lints.clippy]` enable plus four accurate stragglers (journal/slot Pod, rt view lifetime-erasure, rt example); the duplication is gone. Verified CI's exact command (`cargo clippy --workspace --all-targets -- -D warnings`) passes clean at head, so this goes green. Approving.

One gap, tracked as follow-up in #694 rather than blocking: under `--all-features` the lint isn't actually workspace-clean — `nexus-rt/src/reactor.rs` (behind the non-default `reactors` feature) still has ~11–12 undocumented unsafe blocks, invisible to CI because CI omits `--all-features`. The real fix is adding `--all-features` to the clippy CI job so the lint enforces what this PR turns on; #694 covers both.

PR #682, opened 2026-08-16
fix(nexus-platform): remove Sync from Mapping

### LINKED ISSUE #664: nexus-platform: Mapping is Sync but safe write_at(&self) mutates the mapping — data race in safe code
## Summary

`nexus-platform::Mapping` carries `unsafe impl Sync for Mapping`, but exposes a **safe**
`pub fn write_at(&self, data, offset)` that mutates the mapped bytes through a shared
reference (`copy_nonoverlapping` onto the stored raw pointer). With `Sync`, two threads can
hold `&Mapping` (e.g. via `Arc<Mapping>`) and call `write_at` concurrently — a data race on
the mapped memory, reachable entirely from safe code with no `unsafe` at the call site.

## Where

`nexus-platform/src/mapping/mod.rs`: `unsafe impl Sync for Mapping` (~L275),
`pub fn write_at(&self, ...)` (~L214).

## Why it's unsound

`Sync` asserts `&Mapping` is safe to share across threads. A safe `&self` method that writes
to the mapping breaks that: concurrent `write_at` from two `&Mapping` is a data race → UB,
with no `unsafe` required of the caller. The type's own SAFETY comment says "concurrent
access must be synchronized by the caller," but nothing in the API enforces it, because
`write_at` is safe.

## Options (pick one)

1. **Make `write_at` `unsafe`** and document the single-writer / external-synchronization
   requirement — puts the obligation on the caller, which is what `Sync` needs.
2. **Take `&mut self`** on `write_at` — then `Sync` is sound (no shared mutation), at the
   cost of requiring exclusive access to write.
3. **Drop `unsafe impl Sync`** — if `Mapping` doesn't need to be shared across threads,
   `Send`-only suffices.
4. **Atomic writes** — if concurrent writers are genuinely intended, the write path must go
   through atomics (as nexus-shm's slot/ring do via their seqlock), not a plain
   `copy_nonoverlapping`.

Which fits depends on how `Mapping` is actually consumed — nexus-shm builds `ShmSlot`/
`ShmRing` on top and enforces single-writer via the OFD lock, so at the platform level
option 1 or 2 likely matches the real usage.

## Context

Surfaced while reviewing #645 (which adds a SAFETY comment to the `Sync` impl) — the comment
can't be made truthful until this is resolved. Same shape as #662 (nexus-slot): a low-level
crate asserting a safety property (`Sync` / `Pod`) that safe callers can violate.

ChronoCoders:
## Problem

`Mapping` implements both `Send` and `Sync`. With `Sync`, safe code can put a
`Mapping` inside an `Arc`, share it across threads, and cause data races:

- `write_at(&self, ...)` writes through a raw pointer into the mapped bytes
- `read_at(&self, ...)` reads from the same region
- `as_slice(&self)` hands out `&[u8]` over memory another thread can write

Any two of these (or two `write_at` calls) running concurrently on the same
`Mapping` via `Arc<Mapping>` is a data race and undefined behavior. No `unsafe`
block is required from the caller.

## Fix

Remove `unsafe impl Sync for Mapping {}`. Keep `unsafe impl Send for Mapping {}`.

`Send` is correct: moving ownership of a `Mapping` to another thread is safe.
The mapped bytes live in kernel-managed memory, not thread-local storage.

`Sync` is not correct: sharing `&Mapping` across threads allows concurrent
mutation through safe API methods, which is a data race.

## Call-site impact: none

Every production call site either:
- owns the `Mapping` outright (rotating/mod.rs construction paths, shm
  construction paths), or
- holds `&mut self` exclusivity on the outer struct (Writer, Reader), or
- is already `unsafe fn` with its own documented exclusivity invariant
  (conductor.rs `prefault`, which is gated on SWAP_PENDING state and does
  not rely on `Mapping: Sync` for its correctness).

`Segment` in nexus-shm becomes `!Sync` automatically (it had no explicit impl).
`SegmentSwap` is unaffected (it has `unsafe impl Sync for SegmentSwap {}` backed
by its own `UnsafeCell` state machine). `ReadRange` in nexus-journal becomes
`!Send` automatically, which is correct (it is a single-threaded borrowing
iterator).

## Regression guard

A `_assert_mapping_is_send` check is added in `#[cfg(test)]` in `mapping/mod.rs`
to keep the `Send` impl explicit. A `!Sync` negative assertion requires trybuild;
the crate has no trybuild setup and no new dev-dependency was added.

## Follow-on (out of scope)

`as_ptr(&self) -> *mut u8` returns a mutable pointer from a shared reference.
Splitting it into `as_ptr(&self) -> *const u8` and `as_mut_ptr(&mut self) ->
*mut u8` would make the surface stricter, but `conductor.rs::prefault` holds
`&Mapping` (not `&mut`), so that change requires its own design decision.
Tracked in #664.

Closes #664.

## Clippy note

`cargo clippy --all-targets` is clean on 1.94.0, which is what CI pins. On 1.96.0 a `manual_checked_ops` error fires in `nexus-journal/examples/rotating_tail.rs:210` on clean upstream main as well, so it is unrelated to this branch.

### DIFF
--- a/nexus-platform/src/mapping/mod.rs
+++ b/nexus-platform/src/mapping/mod.rs
@@ -267,12 +267,22 @@ impl Drop for Mapping {
     }
 }
 
-// SAFETY: the mapping is a raw pointer plus its backing fd. The bytes
-// live in kernel-managed memory (file-backed or shared), not thread-local
-// state. Concurrent access to the mapped contents must be synchronized
-// by the caller — the handle itself is safe to move across threads.
+// SAFETY: a Mapping is a raw pointer plus its backing fd. Moving the handle to
+// another thread is safe because the mapped bytes live in kernel-managed memory,
+// not thread-local storage.
+//
+// Sync is NOT implemented. as_ptr, as_slice, read_at, and write_at all take
+// &self and touch the mapped bytes directly. Shared cross-thread access through
+// an Arc<Mapping> would therefore be a data race in safe code. Callers that
+// genuinely need to read or write the mapping from a second thread (e.g.
+// conductor.rs prefault) are already unsafe fn with their own exclusivity
+// invariants; they do not rely on Mapping: Sync.
+//
+// Note: as_ptr returns *mut u8 from &self. Splitting that into a const getter
+// (&self -> *const u8) and a mut getter (&mut self -> *mut u8) would make the
+// surface stricter, but conductor.rs holds &Mapping (not &mut), so that change
+// needs its own design decision. Tracked in #664.
 unsafe impl Send for Mapping {}
-unsafe impl Sync for Mapping {}
 
 // ── POSIX shared memory (forwarded to platform backend) ──────────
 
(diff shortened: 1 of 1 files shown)


Abso1ut3Zer0 (approved):

ChronoCoders:
The as_ptr follow-on note in mapping/mod.rs:284 points at #664, which this PR closed. The remaining overlap case between as_slice and write_at is now tracked in #703.

Issue #223, opened 2026-05-09
Add proptests for invariant-heavy crates (currently only 1 in workspace)


Abso1ut3Zer0:
## Context

\`nexus-timer/src/wheel.rs\` is the only proptest user in the workspace. Several crates have headline invariants that example-based tests can't fully exercise.

## Invariant-heavy crates without property tests

- **nexus-queue** — FIFO order, no loss/duplicates under contention
- **nexus-logbuf** — record-length stamping, wraparound, occupancy
- **nexus-pool** — object accounting, no leaks, no double-pop
- **nexus-collections** — BTree balance, RbTree color invariants
- **nexus-decimal** — arithmetic identity laws, precision bounds
- **nexus-stats** — numerical stability, monotonicity

## Action

Even one proptest per crate (covering the headline invariant) catches a class of bugs that example-based tests miss. ~2 hours per crate to add a single property test.

Could be split into per-crate sub-PRs as time allows.

On0n0k1:
Hello. I included proptests for nexus-pool related to this issue.

Please let me know if some changes need to be made.

On0n0k1:
Hello. Added the other proptests that were missing. Please let me know if more should be added.

Abso1ut3Zer0:
Completed by #714. Proptests landed for `nexus-logbuf`, `nexus-collections`, and `nexus-stats-core`; `nexus-decimal` already had coverage predating this issue; `nexus-queue` and `nexus-pool` turned out already covered. That's every crate on the list — closing.

PR #690, opened 2026-08-20
test(nexus-pool): add proptests for object accounting invariants (#223)

On0n0k1:
## Summary
- Adds property-based tests to `nexus-pool/src/local.rs` for both single-threaded pool variants, using the model established in `nexus-timer/src/wheel.rs` (random op-sequence fuzzing against an invariant).
- `Pool`: fuzzes `take`/`try_take`/`acquire`/`put`/drop-guard interleavings, asserting `available() + outstanding == total created` after every op.
- `BoundedPool`: fuzzes `try_acquire`/drop-guard interleavings, asserting `available() + outstanding == capacity` after every op.
- Adds `proptest = "1.4"` to `nexus-pool`'s dev-dependencies (already used elsewhere in the workspace, e.g. `nexus-decimal`, `nexus-slab`, `nexus-timer`).

Together these catch leaks, double-returns, and over-acquisition/phantom creation — the invariant class called out for `nexus-pool` in [Abso1ut3Zer0/nexus#223](https://github.com/Abso1ut3Zer0/nexus/issues/223).

## Scope / follow-up
`sync.rs`'s thread-safe `Pool` is not covered here — its accounting invariant is the same, but the interesting bugs are concurrent races that sequential proptest replay can't exercise. That would want loom (already a dev-dependency) rather than plain proptest, so left for a follow-up.

## Test plan
- [x] `cargo test -p nexus-pool --lib proptests` — both proptests pass (1000 cases each, up to 300 ops per case)
- [x] `cargo clippy -p nexus-pool --tests` — clean


Abso1ut3Zer0 (approved):
Strong invariant test, and I confirmed it bites: the accounting check is a real cross-check (factory `created` vs `available()` vs the test's held-count), and a leak injected into `put()` makes `fuzz_take_put_acquire_drop` fail on "accounting broken" — so it catches the class it claims. Both proptests green (1000 cases each), clippy clean under `--all-features --all-targets`, the dep matches the sibling crates, and deferring `sync.rs` to loom is the right call — sequential replay can't exercise the races.

Two Light nits, non-blocking:
- The growable test's doc says "phantom creation," but `available + outstanding == created` holds under create-instead-of-reuse for a growable pool — that invariant catches leaks and double-returns; over-acquisition is the BoundedPool test's job. Worth tightening the comment.
- `BoundedOp` / `bounded_op_strategy` are defined after the `proptest!` block that uses them, while `Op` is before it — move them up to match.

On0n0k1:
Included the two minor fixes in the next PR.

PR #695, opened 2026-08-26
test(nexus-pool): address review nits from #690

On0n0k1:
Follow-up to #690, addressing the two non-blocking nits from review:

1. The doc comment on fuzz_take_put_acquire_drop claimed the invariant
   catches "leaks, double-returns, and phantom creation." That last
   part isn't true: for the growable Pool, creating a brand new object
   instead of reusing an old one is normal and allowed, so the
   available() and outstanding equation still balances either way. The
   test genuinely can't catch a "created when it should have reused"
   bug, since the numbers add up regardless. Tightened the comment to
   only claim what the test actually checks (leaks and
   double-returns), and pointed to fuzz_bounded_try_acquire_drop as
   the place that does check over-acquisition, since BoundedPool has a
   fixed capacity where that concept applies.

2. Op (the list of possible test actions for Pool) was defined above
   the proptest! block that uses it, but BoundedOp (the same idea for
   BoundedPool) was defined below the block that uses it. Same kind of
   helper, two different spots for no real reason. Both worked fine
   either way, this is a "pick one style and be consistent" cleanup:
   moved BoundedOp/bounded_op_strategy up to match Op/op_strategy's
   placement.

No behavioral change, test logic and coverage are unchanged.

### DIFF
--- a/nexus-pool/src/local.rs
+++ b/nexus-pool/src/local.rs
@@ -699,6 +699,22 @@ mod proptests {
         ]
     }
 
+    /// Operation against a `BoundedPool<u64>`.
+    #[derive(Debug, Clone)]
+    enum BoundedOp {
+        /// `try_acquire()` — succeeds unless the pool is exhausted.
+        TryAcquire,
+        /// Drop an RAII guard (from `TryAcquire`) at `idx`.
+        DropGuard { idx: usize },
+    }
+
+    fn bounded_op_strategy() -> impl Strategy<Value = BoundedOp> {
+        prop_oneof![
+            Just(BoundedOp::TryAcquire),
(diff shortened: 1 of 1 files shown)


On0n0k1:
The build failure is unrelated to the changes from this commit

Abso1ut3Zer0 (approved):
Both nits addressed cleanly. The `BoundedOp`/`bounded_op_strategy` move is byte-identical content, now matching `Op`'s placement, and the tightened doc comment is accurate — leaks + double-returns is what the growable invariant catches, with the over-acquisition case correctly pointed at `BoundedPool`. Proptests green (1000 cases each), clippy clean under `--all-features --all-targets`. No logic change. Approving.

PR #696, opened 2026-08-30
docs(nexus-async-rt): add SAFETY comments to undocumented unsafe blocks

ChronoCoders:
Two unsafe blocks lacked a preceding `// SAFETY:` comment under the workspace `undocumented_unsafe_blocks` lint enabled in #661.

- `src/cancel.rs`: the race-fix comment block (lines 318-332) explains the invariant but does not carry the `// SAFETY:` prefix; added it on the line immediately before the block.
- `tests/tokio_compat.rs`: `Slab::with_chunk_capacity` called inside `catch_unwind` with no SAFETY comment; added one noting the single-threaded context and closure lifetime.

### DIFF
--- a/nexus-async-rt/src/cancel.rs
+++ b/nexus-async-rt/src/cancel.rs
@@ -330,6 +330,8 @@ impl Inner {
                 // this race window to make the regression test deterministic
                 // — see `cancel_race_regression`. In production builds the
                 // hook is compiled out.
+                // SAFETY: list_lock held; cur is non-null and points to a live WaiterNode
+                // (in_list not yet cleared); next field read before the Release store below.
                 let next = unsafe { *(*cur).next.get() };
                 // SAFETY: list_lock held; waker field accessed under lock while cur node is still alive.
                 let waker = unsafe { (*(*cur).waker.get()).take() };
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0 (commented):
Accurate as far as it goes — but this documents `cancel.rs` + the tokio_compat *test*, while `tokio_compat.rs` (source) still has 7 undocumented unsafe blocks under `--all-features` (lines 229, 513, 527, 540, 551, 591, 607). CI stays green either way (it doesn't build `--all-features`), but worth finishing them here — same pattern — so the crate is actually clean rather than half-covered.

Abso1ut3Zer0 (commented):
Closer — the 7 `tokio_compat.rs` blocks are covered. One left, and it's the one that matters most: `unsafe impl Sync for CrossTaskWakerInner` at tokio_compat.rs:209 is still undocumented (the lint flags `unsafe impl` separately from blocks, so it doesn't show in a block-only pass). A `Sync` impl wants the why-it's-safe-to-share justification — what `CrossTaskWakerInner` shares across tasks and why concurrent `&` access is sound — not just a one-liner. Once that's in, the crate is clean under `--all-features`.

Abso1ut3Zer0 (approved):

PR #697, opened 2026-08-30
docs(nexus-ascii): add SAFETY comments to undocumented unsafe blocks in SIMD hash

ChronoCoders:
Three SIMD hash files had undocumented `unsafe` blocks under the workspace `undocumented_unsafe_blocks` lint enabled in #661.

**`src/hash/xxh3_avx2.rs`, `xxh3_avx512.rs`, `xxh3_sse2.rs`** (dispatch sites): comments read `// Safety:` (lowercase) rather than `// SAFETY:`; corrected the casing so the lint recognises them.

**`xxh3_avx2.rs`, `xxh3_avx512.rs`** (function bodies): each `unsafe fn` opened an inner `unsafe {}` block for the SIMD intrinsics without a preceding `// SAFETY:`; added one noting that the `#[target_feature]` attribute on the enclosing function guarantees the CPU feature is available.

### DIFF
--- a/nexus-ascii/src/hash/xxh3_avx2.rs
+++ b/nexus-ascii/src/hash/xxh3_avx2.rs
@@ -28,7 +28,7 @@ pub fn hash_bounded_with_seed<const CAP: usize>(data: &[u8], seed: u64) -> u64 {
     // Large input - use AVX2
     #[cfg(target_arch = "x86_64")]
     {
-        // Safety: This module only compiles when target_feature = "avx2"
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0 (approved):
SIMD SAFETY comments are the right pattern — `#[target_feature(enable=…)]` / module `cfg` guarantees the feature is present, and `SSE2 is always available on x86_64` is correct (it's the x86_64 baseline). Clean under `--all-features --all-targets`. Approving.

Issue #694, opened 2026-08-25
clippy: enforce undocumented_unsafe_blocks under --all-features (nexus-rt reactor.rs undocumented + CI blind spot)


Abso1ut3Zer0:
## Summary

#661 enabled `clippy::undocumented_unsafe_blocks` workspace-wide, but the lint has a
feature-gated blind spot: CI runs `cargo clippy --workspace --all-targets -- -D warnings`
**without `--all-features`**, so any code behind a non-default feature is never compiled and
never linted. The lint passes CI while feature-gated unsafe stays undocumented.

Verified by running the repo's stated stricter standard
(`cargo clippy --workspace --all-targets --all-features -- -D warnings`) at #661's head.

## Gap 1 — `nexus-rt` reactor.rs is undocumented under `--all-features`

`nexus-rt/src/reactor.rs` is behind `#[cfg(feature = "reactors")]` (not a default feature;
only in `full`), so CI never compiles it. Under `--all-features` it has **~11–12
undocumented unsafe blocks** — the `unsafe { world.get_mut::<…>(id) }` dispatch pattern
(e.g. reactor.rs:908, 920, 945, 1092, …) and the `&mut *notify_ptr` derefs around it.

**Fix:** add SAFETY comments using the same World-provenance justification #659 applied to
the non-gated dispatch code — id obtained from `register()` on the same builder, `T` matches
the registered type, single-threaded sequential dispatch ⇒ no mutable aliasing; the scoped
`&mut` is dropped before `reactor.run()`.

## Gap 2 — CI clippy should run `--all-features` (the real fix)

reactor.rs is only the instance that surfaced. The standing problem is that the new lint
(and every other clippy lint) is unenforced on all feature-gated code. Add `--all-features`
to the clippy CI job so the lint enforces what #661 claims. This matches the repo's stated
clippy standard, which is `--all-features --all-targets` (earlier SAFETY-audit PRs were held
to it).

Once `--all-features` is on, do a full sweep — other feature-gated crates may have their own
undocumented blocks that CI has never linted.

## Acceptance

- `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes.
- The CI clippy job includes `--all-features`.

Follow-up to #661.

PR #659, opened 2026-08-08
docs(nexus-rt): add SAFETY comments to all unsafe blocks

ChronoCoders:
Adds missing `// SAFETY:` comments to all `unsafe` blocks in nexus-rt, satisfying `clippy::undocumented_unsafe_blocks` with zero warnings.

PR #698, opened 2026-08-30
docs(nexus-rt): add SAFETY comments to undocumented unsafe blocks

ChronoCoders:
6 undocumented `unsafe {}` blocks across 4 files.

- `examples/mio_timer.rs`, `examples/perf_reactors.rs`: `rdtsc_start` and `rdtsc_end` use x86_64 intrinsics; both functions are guarded by `#[cfg(target_arch = "x86_64")]`.
- `src/reactor.rs`, `tests/compile_tests.rs`: `world.get_mut::<ReactorNotify>(id)` -- id is a registered ResourceId with no concurrent alias at the call site.

### DIFF
--- a/nexus-rt/examples/mio_timer.rs
+++ b/nexus-rt/examples/mio_timer.rs
@@ -35,6 +35,7 @@ use nexus_rt::{
 #[inline(always)]
 #[cfg(target_arch = "x86_64")]
 fn rdtsc_start() -> u64 {
+    // SAFETY: x86_64 intrinsics; guarded by #[cfg(target_arch = "x86_64")].
     unsafe {
         core::arch::x86_64::_mm_lfence();
         core::arch::x86_64::_rdtsc()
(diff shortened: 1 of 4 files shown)


Abso1ut3Zer0 (commented):
This documents one `reactor.rs` block, but ~5 remain undocumented under `--all-features` (673, 908, 920, 945 — the `world.get_mut` dispatch pattern). So it doesn't close the reactor.rs gap tracked in #694 — worth completing them in the same PR since it's the same file and pattern. Minor: the new comment ("no concurrent alias at this call site") is terser than its siblings, which name *why* — "single-threaded sequential dispatch ensures no mutable aliasing"; worth matching for consistency.

Abso1ut3Zer0 (approved):
Now clean under `--all-features --all-targets` — the remaining reactor.rs blocks are documented, the comments match the #659 World-provenance pattern (id from the same WorldBuilder, scoped `&mut` dropped before `reactor.run()`), and the terser one now names *why* — "single-threaded sequential dispatch ensures no mutable aliasing" — so the consistency note is addressed as well. This closes the reactor.rs half of #694. Approving.

PR #701, opened 2026-08-30
docs(nexus-async-web): add SAFETY comments to undocumented unsafe blocks

ChronoCoders:
6 undocumented `unsafe {}` blocks across 3 files.

- `benches/perf_ws_cycles_nexus.rs`: 2 rdtsc functions (`rdtsc_start`, `rdtsc_end`) use x86_64 intrinsics in an x86_64-only benchmark; 1 noop waker with null data and all-noop vtable.
- `benches/perf_async_ws_nexus.rs`: 1 noop waker, same pattern.
- `tests/tls_handshake_piggyback.rs`: `alloc` and `dealloc` in a `GlobalAlloc` impl delegate directly to `System`; both cite the GlobalAlloc caller safety contract.

### DIFF
--- a/nexus-async-web/benches/perf_async_ws_nexus.rs
+++ b/nexus-async-web/benches/perf_async_ws_nexus.rs
@@ -120,6 +120,7 @@ fn noop_waker() -> Waker {
         RawWaker::new(p, &VTABLE)
     }
     const VTABLE: RawWakerVTable = RawWakerVTable::new(clone, noop, noop, noop);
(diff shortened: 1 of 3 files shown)


Abso1ut3Zer0 (approved):
noop-waker (null data is never dereferenced by the vtable), rdtsc arch-gate, and GlobalAlloc-delegates-to-System comments all accurate. Passes CI. (`--all-features` is N/A for this crate — `tokio-rt` and `nexus` are mutually exclusive by `compile_error!`.) Approving.

PR #709, opened 2026-09-08
fix(nexus-decimal): round_to_tick misrounds for odd tick sizes

### LINKED ISSUE #704: nexus-decimal: round_to_tick returns wrong result for odd tick sizes
`from_raw(4).round_to_tick(from_raw(3))` returns `0.00000006`. The nearest tick multiple of 3 from 4 is 3 (distance 1), not 6 (distance 2).

```
test financial::audit_f08::round_to_tick_odd_tick_misrounds ... FAILED
assertion `left == right` failed
  left: 0.00000006
 right: 0.00000003
```

Toolchain: rustc 1.94.0.

**Cause.**

`nexus-decimal/src/financial.rs:83`:

```rust
let half_tick = tick.value / 2;
```

Integer division truncates. For odd ticks, `tick.value / 2` is not the true midpoint: `3 / 2 = 1`, but the midpoint between 3 and 6 is 4.5. With `value=4` and `tick=3`:

- `remainder = 4 % 3 = 1`
- `half_tick = 3 / 2 = 1`
- `remainder == half_tick` is true (line 96), so the banker's rounding branch fires
- `quotient = 4 / 3 = 1`, which is odd, so the code rounds up: `3 + 3 = 6`

The value 4 is not at the midpoint. It is closer to 3. The banker's rounding branch should only fire at the true midpoint.

The same false-midpoint condition applies to negative values: for `value=-4, tick=3`, `remainder == -half_tick` (-1 == -1) fires the same branch; whether it produces the wrong result there is untested.

**Existing property tests do not catch this.** The tests in `nexus-decimal/tests/financial_props.rs:88-157` check that the result is aligned to the tick and within the floor/ceil bounds. The value 6 passes both checks for input 4 with tick 3. Nearest-distance correctness is not asserted.

ChronoCoders:
Closes #704

The implementation divided tick by 2 to find the midpoint, which truncates
for odd ticks. For tick=3, half_tick=1, so a remainder of 1 was placed at
the midpoint branch incorrectly (true midpoint is 1.5), causing values like
4 to round up to 6 instead of down to 3.

Replace the division with a complement: complement = tick - |remainder|.
Compare |remainder| against complement directly. No division, no truncation.
For odd ticks there is no exact midpoint, so the banker's rounding branch is
never reached incorrectly. For even ticks the midpoint branch fires only when
|remainder| == complement, which is the true half-tick.

Two new tests: round_to_tick_odd_tick covers tick=3 for positive and
negative values; round_to_tick_even_tick_bankers covers tick=4 banker's
cases. The audit test (f08) was moved into nexus-decimal/tests/financial.rs
under those names.

### DIFF
--- a/nexus-decimal/CHANGELOG.md
+++ b/nexus-decimal/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `round_to_tick` misrounded values for odd tick sizes. The old code computed
+  `half_tick = tick / 2` (integer division), which truncates for odd ticks:
+  with `tick=3`, `half_tick=1`, so a remainder of `1` landed in the
+  banker's-rounding branch even though the true midpoint is `1.5`. The fix
+  replaces the division with a complement: `complement = tick - |remainder|`.
+  Comparing `|remainder|` against `complement` is exact and needs no
+  division — for odd ticks `2 * |remainder|` can never equal an odd tick, so
+  the midpoint branch is unreachable. This subsumes the `1.2.2` `tick=1` case
+  (where `half_tick=0` made every non-zero remainder hit the branch) and the
+  general odd-tick class.
+
 ## [1.2.2] — 2026-05-23
 
 ### Fixed
--- a/nexus-decimal/src/financial.rs
+++ b/nexus-decimal/src/financial.rs
@@ -80,38 +80,59 @@ macro_rules! impl_decimal_financial {
                 if remainder == 0 {
                     return Some(self);
                 }
-                let half_tick = tick.value / 2;
                 let base = self.value - remainder;
 
-                if remainder > half_tick {
-                    match base.checked_add(tick.value) {
-                        Some(v) => Some(Self { value: v }),
-                        None => None,
-                    }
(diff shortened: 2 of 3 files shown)


Abso1ut3Zer0:
Correctness verified — traced both signs for odd-tick (`3`) and banker's (`4`); the fix is surgical (even-tick behavior unchanged, only odd ticks differ), and `tick=1` still hits the `remainder == 0` early return so the `[1.2.2]` fix is preserved. One ask + one note:

**CHANGELOG (please add before merge).** User-facing correctness fix, and `round_to_tick` already has a `[1.2.2]` entry for the sibling `tick=1` half-tick bug — add an `## [Unreleased]` entry (no version bump). Worth noting the complement approach subsumes the whole `half_tick` integer-division bug class (the `1.2.2` `tick=1` case and odd ticks here).

**Perf (not blocking — tracked in #712).** Review found a redundant division on the banker's path: `self.value % tick.value` and `self.value / tick.value` are the same division but LLVM doesn't fuse them (asm shows two `idivq`). A quotient-first form removes it and merges the mirror branches, but it's a p50-vs-tail tradeoff, so we're measuring it separately rather than gating this. Correctness fix stands on its own.

Abso1ut3Zer0 (approved):

PR #710, opened 2026-09-08
fix(nexus-id): uuid v7 sequence wraps through u16 after exhaustion

### LINKED ISSUE #705: nexus-id: UuidV7Generator sequence counter keeps advancing after SequenceExhausted, re-using sequence 0
After 4096 UUIDs in the same millisecond, `next_raw` correctly returns `SequenceExhausted`. The sequence counter continues advancing on each failed call. After 61440 further errors, the `u16` counter wraps back to 0 via `wrapping_add`. The next call then succeeds, returning a UUID whose sequence field is 0: the same sequence as the first UUID generated for that millisecond.

```
test uuid::v7::tests::f13_sequence_does_not_wrap_through_u16 ... FAILED
assertion `left != right` failed
  left: 111411200000028672
 right: 111411200000028672
```

Toolchain: rustc 1.94.0.

**Cause.**

`nexus-id/src/uuid/v7.rs:134`:

```rust
self.sequence = self.sequence.wrapping_add(1);
if self.sequence > SEQUENCE_MAX {
    return Err(SequenceExhausted { ... });
}
```

The counter is incremented before the limit check. On error, the incremented value is kept. After 65536 total increments from 0, `wrapping_add` returns the counter to 0. `0 > SEQUENCE_MAX` is false, so the call succeeds and packs sequence 0 into the UUID's `hi` word, producing the same `hi` value as the first UUID in the millisecond.

**Correct behavior.** Once the sequence is exhausted for a given timestamp, subsequent calls for the same timestamp should all return `SequenceExhausted`. The counter should not advance past the exhausted state.

ChronoCoders:
Closes #705

wrapping_add incremented the counter past SEQUENCE_MAX, cycling it back
through 0 after 65536 total calls (4096 successes + 61440 errors). The
counter returned to 0, allowing sequence reuse within the same millisecond.

Fix: check the limit before incrementing. The counter now stays at
SEQUENCE_MAX permanently once exhausted. Any call after that returns
SequenceExhausted for the lifetime of the generator.

New test sequence_exhausted_state_is_permanent drives the counter through
its remaining u16 range after exhaustion and asserts the final call still
fails. The audit test (f13) was replaced by this test.

### DIFF
--- a/nexus-id/CHANGELOG.md
+++ b/nexus-id/CHANGELOG.md
@@ -10,6 +10,18 @@ contained.
 
 ## [Unreleased]
 
+### Fixed
+
+- `UuidV7` sequence counter wrapped through `u16::MAX` after exhaustion.
+  `wrapping_add` incremented the counter past `SEQUENCE_MAX` (4095) on
+  every error-returning call. After 61440 further calls the counter
+  cycled back to `0`, and the next call succeeded with sequence `0`,
+  reusing a value within the same millisecond. This produced duplicate
+  and non-monotonic UUIDs. Fixed by checking `sequence >= SEQUENCE_MAX`
+  before incrementing: once exhausted the counter stays at `4095`
+  permanently and every subsequent call returns `SequenceExhausted`.
+
+
 ## [1.1.5] — 2026-05-10
 
 Doc + bench infra release. No public API change.
--- a/nexus-id/src/uuid/v7.rs
+++ b/nexus-id/src/uuid/v7.rs
@@ -131,13 +131,13 @@ impl UuidV7 {
 
         // Handle sequence
         if ts_ms == self.last_ts_ms {
-            self.sequence = self.sequence.wrapping_add(1);
-            if self.sequence > SEQUENCE_MAX {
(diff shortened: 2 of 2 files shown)


Abso1ut3Zer0:
Correctness verified — check-before-increment pins the counter at `SEQUENCE_MAX` (4095) permanently, the success path (sequences 0..4095, 4096 values) is unchanged, and `+= 1` can't overflow since it's guarded by `< SEQUENCE_MAX`. The test is a real regression guard: it's deterministic (passing `epoch` as `now` fixes `ts_ms`, so it doesn't depend on wall-clock), tuned to hit the u16 wrap exactly on the final assert, and it fails on the old code (the wrapped-to-0 call returns `Ok`). `f13` fully absorbed — nothing left behind.

One ask before merge:

**CHANGELOG.** User-facing correctness fix (sequence reuse → duplicate, non-monotonic UUIDs within a millisecond) — please add a `## [Unreleased]` entry to `nexus-id/CHANGELOG.md` (no version bump).

Abso1ut3Zer0 (approved):

PR #711, opened 2026-09-08
fix(nexus-id): snowflake next() does not reject tick exceeding TIMESTAMP_MAX

### LINKED ISSUE #706: nexus-id: Snowflake64::next silently truncates the timestamp field when tick exceeds TIMESTAMP_MAX
`Snowflake64<2, 0, 62>::next(0)` and `Snowflake64<2, 0, 62>::next(4)` both return `0`. For this layout, `TIMESTAMP_MAX` is 3 (2-bit timestamp field). Tick 4 (`0b100`) shifted left 62 bits exceeds 64 bits; the high bits are silently dropped, producing the same raw ID as tick 0.

The 2-bit timestamp is a test convenience to reach the boundary in four ticks; the missing check is layout-independent: a typical 42-bit layout hits the same silent truncation after 2^42 ticks instead of 4.

```
test snowflake::tests::f14_timestamp_overflow_produces_duplicate_id ... FAILED
assertion `left != right` failed
  left: 0
 right: 0
```

Toolchain: rustc 1.94.0.

**Cause.**

`nexus-id/src/snowflake.rs:307`:

```rust
Ok(T::from_raw(
    (tick << Self::TS_SHIFT) | self.worker_shifted | self.sequence,
))
```

`tick` is shifted left by `TS_SHIFT` (= `WK + SQ`) without a prior check against `TIMESTAMP_MAX`. When `tick > TIMESTAMP_MAX`, the upper bits of the field are lost in the shift. `TIMESTAMP_MAX` is defined at line 221:

```rust
pub const TIMESTAMP_MAX: u64 = (1u64 << TS) - 1;
```

It is not checked anywhere in `next()`.

**Correct behavior.** `next()` should return an error when `tick > Self::TIMESTAMP_MAX`. Silently truncating the timestamp field produces duplicate IDs without any signal to the caller. The monotonically non-decreasing contract stated in the docs cannot hold once the field wraps.

ChronoCoders:
Closes #706

`tick << TS_SHIFT` silently truncated the high bits when tick > TIMESTAMP_MAX,
producing IDs with the same encoded timestamp as a smaller tick and breaking
the uniqueness guarantee.

Add a guard at the top of next(): return Err(SnowflakeError::TimestampOverflow)
when tick > TIMESTAMP_MAX.

Option A (change all six signatures to return SnowflakeError outright, dropping
SequenceExhausted entirely) was rejected because it breaks the error type for
every existing caller. Option B (add next_checked and leave next() unguarded)
was rejected because next() without the guard silently produces duplicate IDs
for any caller who does not know to switch. The default must be safe.

SnowflakeError is a new enum with two variants: Exhausted (carries tick and
max_sequence) and TimestampOverflow (carries tick and max). SequenceExhausted
is now a type alias for SnowflakeError so the six generation signatures compile
unchanged.

What breaks: struct-literal construction (SequenceExhausted { tick, max_sequence })
and struct-pattern destructuring against those field names. All construction is
internal to nexus-id. Checked: nexus-id/src/snowflake.rs, nexus-id/src/uuid/v7.rs,
nexus-id/src/ulid.rs. All updated to use SnowflakeError::Exhausted { ... }.
The uuid/v7 sequence_exhaustion test accessed err.max_sequence as a struct
field; updated to destructure SnowflakeError::Exhausted.

Mutation check: guard removed from next(), test timestamp_overflow_is_rejected:

    test snowflake::tests::timestamp_overflow_is_rejected ... FAILED
    thread panicked at nexus-id\src\snowflake.rs:639:9:
    assertion failed: matches!(g.next(4),
        Err(SnowflakeError::TimestampOverflow { tick: 4, max: 3 }))

Guard restored:

    test snowflake::tests::timestamp_overflow_is_rejected ... ok

### DIFF
--- a/nexus-id/CHANGELOG.md
+++ b/nexus-id/CHANGELOG.md
@@ -10,6 +10,30 @@ contained.
 
 ## [Unreleased]
 
+### Changed (breaking)
+
+- `SequenceExhausted` is restored as a standalone struct with fields `tick: u64`
+  and `max_sequence: u64`. It is no longer a type alias for `SnowflakeError`.
+  ULID `try_next` and UUID v7 `next*` continue to return `SequenceExhausted`.
+  Migration: replace any `SequenceExhausted::Exhausted { tick, max_sequence }`
+  construction or pattern with `SequenceExhausted { tick, max_sequence }`.
+
+- `SnowflakeError` is now a separate `#[non_exhaustive]` enum returned only by
+  the six Snowflake methods (`next`, `mixed`, `next_id`, `mixed_id`,
+  `next_signed`, `mixed_signed`). Variants: `Exhausted(SequenceExhausted)` and
+  `TimestampOverflow { tick: u64, max: u64 }`. A `From<SequenceExhausted> for
+  SnowflakeError` impl is provided for composition.
+  Migration: change `SnowflakeError::Exhausted { max_sequence, .. }` patterns
+  to `SnowflakeError::Exhausted(err)` and read `err.max_sequence`.
+
+### Added
+
+- `Snowflake::next()` now returns `Err(SnowflakeError::TimestampOverflow)`
+  when the caller supplies a tick value that exceeds `TIMESTAMP_MAX` for the
+  generator's bit layout. Previously the tick was silently truncated, producing
+  IDs with the wrong timestamp and potentially colliding with earlier IDs at
+  tick 0.
+
 ## [1.1.5] — 2026-05-10
 
 Doc + bench infra release. No public API change.
--- a/nexus-id/src/lib.rs
+++ b/nexus-id/src/lib.rs
@@ -101,8 +101,8 @@ pub mod ulid;
 pub mod uuid;
 
 pub use snowflake::{
-    IdInt, SequenceExhausted, Snowflake, Snowflake32, Snowflake64, SnowflakeSigned32,
-    SnowflakeSigned64,
+    IdInt, SequenceExhausted, Snowflake, Snowflake32, Snowflake64, SnowflakeError,
+    SnowflakeSigned32, SnowflakeSigned64,
 };
 
 pub use parse::{DecodeError, ParseError, TypeIdParseError, UuidParseError};
--- a/nexus-id/src/snowflake.rs
+++ b/nexus-id/src/snowflake.rs
@@ -112,38 +112,49 @@ impl_id_int!(
     i64 => 64,
 );
 
-/// Sequence exhausted within the current tick.
-///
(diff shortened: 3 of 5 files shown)


Abso1ut3Zer0:
This is going into a batched **`nexus-id` major (2.0)** — tracked in #713 — so it ships alongside the other breaking cleanups rather than as its own major. Two things for this PR:

**1. Separate the error rather than share one enum**, so ULID and UUID v7 don't carry a `TimestampOverflow` variant they can never produce, and don't return a type named `SnowflakeError`:

- **Snowflake** (the six `next*`/`mixed` signatures) → new `SnowflakeError` enum: `Exhausted { tick, max_sequence }` + `TimestampOverflow { tick, max }`, and mark it `#[non_exhaustive]` from birth (so future failure modes don't force another major).
- **ULID** (`try_next`) and **UUID v7** (`next*`) → keep returning `SequenceExhausted` unchanged. They only ever exhaust the sequence — no ghost variant, no rename, zero break for their callers.
- Drop the `pub type SequenceExhausted = SnowflakeError` alias; the two are separate now. (Optional: `SnowflakeError::Exhausted(SequenceExhausted)` + a `From` impl if you prefer composition over duplicated fields.)

The overflow guard itself is exactly right — keep it as-is.

**2. CHANGELOG + semver.** This is a major break (external callers reading `err.max_sequence` or matching the struct break — the alias only preserved the type name in signatures). Add an `## [Unreleased]` entry documenting the break + migration (`err.max_sequence` → `match SnowflakeError::Exhausted { max_sequence, .. }`). No version bump in the PR — the 2.0 bump happens at release time when the #713 items are assembled.

Abso1ut3Zer0:
Code approved — option B is exactly right: `SnowflakeError` is `#[non_exhaustive]` from birth, uses `Exhausted(SequenceExhausted)` + a `From` impl, `SequenceExhausted` stays its own struct so ULID/UUID v7 are untouched, guard + tests intact, CI green.

The two CHANGELOG items are **nits to fix before release, not merge blockers** — the entry sits in `[Unreleased]` and gets assembled/reviewed when 2.0 is cut, so they're tracked as a CHANGELOG accuracy pass on #713:

1. Method-name list should be `next` / `mixed` / `next_id` / `next_mixed` (no `mixed_id` / `next_signed` / `mixed_signed` — signed Snowflakes only have raw `next()`).
2. Reframe the `SequenceExhausted` bullet — from the last release it was always a struct, so for users it's *unchanged* for ULID/UUID v7 (no migration there); only Snowflake's error type changes (`SequenceExhausted` → `SnowflakeError`).

Good to merge into the major.

Abso1ut3Zer0 (approved):

PR #714, opened 2026-09-12
test: add proptests for nexus-logbuf, nexus-collections, nexus-stats-core (#223)

On0n0k1:
## Summary

Continues the proptest coverage push from #223. Three crates get new property tests in this PR; two others turned out to already be covered and needed no changes (noted below for the record).

**nexus-logbuf** (`queue/spsc.rs`)
- `fuzz_write_abort_read_roundtrip`: random write/abort/read sequences against small power-of-two capacities, checked against an oracle queue of committed payloads for exact content/order. `Read` returning nothing is only valid when the oracle is also empty, cross-checking the head/tail occupancy bookkeeping. Small capacities force records across the ring boundary constantly, exercising the wraparound/skip-marker path.
- `fuzz_capacity_recovery`: fills to `BufferFull`, frees one record, asserts a same-size write immediately recovers — no bookkeeping leak.

**nexus-collections** (`btree_test.rs`, `rbtree_test.rs`)
- Fuzzes insert/remove for both `BTree` and `RbTree` against a `std::collections::BTreeMap` oracle, with a small key range relative to op count so keys collide often (replace-existing, remove-then-reinsert, remove-of-absent-key).
- Cross-checks each tree's own `verify_invariants()` after every op (already existed in the source, just wasn't exercised by randomized input) — min-occupancy/sorted-keys for BTree, red-black color/black-height for RbTree.

**nexus-stats-core** (`statistics/welford.rs`, `monitoring/drawdown.rs`)
- The `nexus-stats` crate named in #223 is a pure re-export facade; the actual implementations live in `nexus-stats-core`, so that's where these landed.
- `WelfordF64`: variance never negative, matches an independent two-pass reference within tolerance, and merge (Chan's algorithm) matches a single-pass accumulation of the same data.
- `DrawdownF64`: `peak()`/`max_drawdown()` monotonically non-decreasing across updates, `drawdown()` never negative.

**Already covered, no changes needed:**
- `nexus-decimal` — had proptest coverage (`tests/*_props.rs`) predating #223.
- `nexus-queue` — has loom-based tests (`tests/loom_{spsc,mpsc,spmc}.rs`) covering FIFO order, no-lost-items, no-duplicates, and wraparound under real concurrency, which is the correct tool for that crate's invariant (plain proptest replays on one thread and can't exercise contention).

This closes out every crate on #223's list.

### DIFF
--- a/nexus-collections/Cargo.toml
+++ b/nexus-collections/Cargo.toml
@@ -16,6 +16,7 @@ slab = ["dep:nexus-slab"]
 
 [dev-dependencies]
 hdrhistogram = "7.5"
+proptest = "1.4"
 rand = "0.9.2"
 seq-macro.workspace = true
 
--- a/nexus-collections/tests/btree_test.rs
+++ b/nexus-collections/tests/btree_test.rs
@@ -328,3 +328,82 @@ fn non_empty_drop_panics_in_debug() {
         "dropping non-empty btree should panic in debug"
     );
 }
+
+// =============================================================================
+// Proptests
+// =============================================================================
+
+mod proptests {
+    use super::*;
+    use proptest::prelude::*;
+    use std::collections::BTreeMap as StdMap;
+
+    /// Operation against a `BTree<u16, u64>` under test.
+    #[derive(Debug, Clone)]
+    enum Op {
+        Insert { key: u16, value: u64 },
+        Remove { key: u16 },
+    }
+
+    fn op_strategy() -> impl Strategy<Value = Op> {
+        prop_oneof![
+            (0u16..64, any::<u64>()).prop_map(|(key, value)| Op::Insert { key, value }),
+            (0u16..64).prop_map(|key| Op::Remove { key }),
+        ]
+    }
+
+    proptest! {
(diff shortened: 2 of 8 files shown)


Abso1ut3Zer0:
Reviewed — clean, approving. Test-only (+453/-0), all CI green. Spot-checked the usual proptest failure modes:

- Oracle + `verify_invariants()` per op + final content compare (collections) — real efficacy.
- Welford uses **relative** tolerance (`1e-6 * val.abs().max(1.0)`) over bounded input, and the merge test clamps `split.min(len-1).max(1)` so both halves stay in-bounds — no float-flakiness or index-panic traps.
- `proptest = "1.4"` matches the workspace convention.

Trivial optional: `split in 1usize..199` hard-codes 199 against the samples max of 200 — the clamp keeps it correct, but deriving it would avoid a footgun if that max changes. Non-blocking.

Good call using loom for `nexus-queue` rather than forcing proptest, and catching that `nexus-stats` is a re-export facade.

Abso1ut3Zer0 (approved):

On0n0k1:
Included the optional change mentioned.

PR #729, opened 2026-09-21
fix: reject cyclic and shared-child GBDT trees before compaction

### LINKED ISSUE #707: nexus-inference: Gbdt::from_lightgbm hangs on model with self-referencing internal node
Feeding the following model text to `Gbdt::from_lightgbm` causes the process to hang indefinitely:

```
max_feature_idx=0
num_class=1
Tree=0
num_leaves=2
split_feature=0
threshold=0.5
decision_type=0
left_child=0
right_child=-1
leaf_value=1 2
end of trees
```

The test binary started (`running 1 test`), produced no further output, and was killed after a 10-second wall-clock timeout (PID 7692). No panic, no error return.

**Cause.**

`nexus-inference/src/loader/lightgbm.rs:200-213`:

```rust
fn remap_child(child: i32, num_internal: usize, num_leaves: usize) -> Result<usize, LoadError> {
    if child >= 0 {
        let idx = child as usize;
        if idx >= num_internal {
            return Err(LoadError::Validation("internal child index out of range"));
        }
        Ok(idx)
    } else {
        let leaf_idx = (-(child + 1)) as usize;
        if leaf_idx >= num_leaves {
            return Err(LoadError::Validation("leaf child index out of range"));
        }
        Ok(num_internal + leaf_idx)
    }
}
```

`remap_child(0, 1, 2)` returns `Ok(0)`: index 0 is in range (`0 < 1`), so the root's left child is accepted as node 0 itself. No check that the child index differs from the current node.

`nexus-inference/src/gbdt.rs:70-93`:

```rust
while let Some(old_idx) = stack.pop() {
    // ...
    stack.push(r.left as usize);
    stack.push(r.right as usize);
}
```

No visited set. With `left=0`, popping node 0 pushes node 0 back. The stack never empties and the `nodes` Vec grows without bound.

The model input above is not a programming error. It passes all range checks in `remap_child`. Untrusted model files loaded from disk or a remote source are the realistic trigger: a malformed or adversarial LightGBM model file causes unbounded CPU and memory consumption with no error returned to the caller.

ChronoCoders:
`validate_tree` runs a single-pass iterative DFS from node 0 with a visited bitset before `reorder_and_compact` sees the data. Two checks:

1. A node reached twice is an error (cycle or shared child). This deviates from the three-colour design in the issue. Three-colour rejects back-edges (cycles) but accepts shared children (node i where both left and right point to i+1) because they produce no back-edge and the DFS completes. Shared children are equally invalid in a binary tree and cost 2^k: `reorder_and_compact` visits node i with 2^i frequency, silently returning Ok for a corrupt model past 65536 compacted nodes. Measured on main (rustc 1.94.0, debug): k=10 180us, k=14 1.5ms, k=18 23.7ms. A visited bitset catches both defects with one check.

2. After the walk, any unvisited node is an error (orphan not reachable from root).

The DFS is iterative (not recursive) so a hostile depth cannot overflow the stack. O(V+E), no allocation beyond the bitset and the stack.

Tests: root self-loop, two-cycle, shared-child chain (k=20), unreachable node, valid multi-level tree that still loads and predicts correctly. Tests 1-4 assert Err within 200ms; tests 1 and 2 use a channel timeout to turn a hang into a test failure rather than a silent CI block.

Mutation check (a): removing the second-visit check causes tests 1 and 2 to timeout at 200ms; test 3 still returns Err via the count check (shared children always imply unreachable leaves). Mutation check (b): removing the count check causes test 4 to return Ok.

Closes #707.


Abso1ut3Zer0 (commented):
Correct fix, and the shared-child chain test is a good catch beyond the issue as filed (three-colour would have passed it). Visited-once plus the count check is the right test here: n nodes reached from the root over n - 1 edges with no repeat is exactly "this is a tree", and an in-degree count alone would miss a cycle disconnected from the root.

Three small things before merge:

1. **Fold the check into `reorder_and_compact`'s existing DFS** (`gbdt.rs:58`, called at `:308`) instead of a separate pre-pass in the LightGBM loader: mark visited on pop, error on a repeat, check the count after. One traversal instead of two, and the guard sits in the loop that would otherwise hang, so a second loader can't bypass it. `reorder_and_compact` returns `Result` then; one call site.
2. **Drop the `elapsed < 100ms` assert in `rejects_unreachable_node`.** Wall-clock asserts flake under CI load in debug builds, and the `unwrap_err()` on the line before already proves it returned. The hang case is covered by the 200 ms harness on the other tests.
3. **CHANGELOG:** a `Fixed` line under `[Unreleased]` in `nexus-inference/CHANGELOG.md` for #707.

ChronoCoders:
All three done in 00d41f7, squashed to one commit. The check now lives in reorder_and_compact's own DFS: visited on pop, error on a repeat, count after the loop. validate_tree and its call are gone, reorder_and_compact and from_parts return Result, and the wall-clock assert is dropped. Mutation redone against the new shape: removing the repeat-pop check times out the two cycle tests at the 200 ms harness and fails the shared-child one; removing the count check fails the unreachable test.

Abso1ut3Zer0 (approved):

PR #730, opened 2026-09-21
fix: bound max_head_size check on head_len, not buffer length

### LINKED ISSUE #708: nexus-web: ResponseReader::try_parse rejects a valid response when header and body arrive in the same read
The same 9041 bytes (41-byte header, 9000-byte body) parsed three ways:

| Variant | Input | Result |
|---------|-------|--------|
| One read (header + body together) | 9041 bytes | `Err(HeadTooLarge { max: 8192 })` |
| Two reads (header first, then body) | 41 bytes, then 9000 bytes | `Ok(Some(...))` |
| Byte by byte | 9041 reads of 1 byte | `Ok(Some(...))` |

Wire: `HTTP/1.1 200 OK\r\nContent-Length: 9000\r\n\r\n` (41 bytes) followed by 9000 `x` bytes. The header alone is 41 bytes. The limit is 8192. The header is well under the limit in all three variants; only the split determines the result.

Test ran with `cargo +1.94.0 test --package nexus-web --test audit_f10`.

**Cause.**

`nexus-web/src/http/response.rs:331`:

```rust
if data.len() > self.max_head_size {
    return Err(HttpError::HeadTooLarge { max: self.max_head_size });
}
```

This check runs at the top of `try_parse`, before searching for `\r\n\r\n`. When header and body arrive in the same `read` call, `data.len()` includes body bytes. The header terminator has not yet been found, so the body bytes count against the head size limit.

`ResponseReader::new(capacity)` sets the read buffer capacity, not `max_head_size`. The head size limit defaults to 8192 and is set separately via `.max_head_size(n)`. The test used `ResponseReader::new(8192)` with the default `max_head_size` of 8192. The 41-byte header is under the limit; the rejection happens because body bytes inflate the buffer before the terminator is found.

The accepted behavior depends on how the network happened to split the response. A response that parses correctly when headers and body arrive in separate TCP segments fails when they are coalesced, which is a normal and common occurrence.

ChronoCoders:
| Variant | `data.len()` | `head_len` | Old result | New result |
|---|---|---|---|---|
| head only, within limit | 27 | 27 | ok | ok |
| head + body, head within limit, total over | 127 | 27 | `HeadTooLarge` (wrong) | ok |
| head over limit | 60 | 60 | `HeadTooLarge` | `HeadTooLarge` |

`try_parse` in both readers checked `data.len() > max_head_size` before calling httparse. When a response or request arrives with body bytes coalesced into the same read, `data.len()` includes those bytes, so a header of 27 bytes with a 100-byte body triggers the limit at 127, not at 27. The correct measure is `head_len` from `Status::Complete(head_len)`.

Fix: remove the upfront check. In the `Complete` arm, reject if `head_len > max_head_size` (real head size). In the `Partial` arm, retain the check on `data.len()` (all buffered bytes are candidate header while the terminator is not yet seen, so this remains the DoS guard).

Six tests per reader: `split_read_one_chunk`, `split_read_two_chunks`, `split_read_byte_by_byte` (regression for the split-read case), `complete_head_over_limit`, `partial_head_over_limit`, `exact_head_size_boundary` (at-limit accepted, over-limit rejected).

Mutation check on `RequestReader`:
- Remove `Complete` arm check: `complete_head_over_limit`, `head_too_large`, `exact_head_size_boundary` fail.
- Remove `Partial` arm check: `partial_head_over_limit` fails exclusively.
- Remove both: all four fail.

Closes #708.


Abso1ut3Zer0 (commented):
Right fix, and doing both readers in one PR with matching tests is the way to do it.

Two things before merge:

1. **Doc the two regimes on `max_head_size`** (the builder method): the head must fit in N bytes once the terminator is seen; until then every buffered byte counts. The behavior is now correct, but the semantics live only in the two match arms.
2. **CHANGELOG:** a `Fixed` line under `[Unreleased]` in `nexus-web/CHANGELOG.md` for #708. Responses that were rejected are now accepted, so it's user-visible.

ChronoCoders:
Both done in 8c35f20. The max_head_size doc now states both regimes, once the terminator is seen the head must fit in n bytes, and until then every buffered byte counts. There is a Fixed entry under [Unreleased] for #708.

Abso1ut3Zer0 (approved):

